PR #2, opened 2022-10-07
Add minimal instructions, close #1

### LINKED ISSUE #1: Not clear how to draw boundaries
First time I tried it, took some time to find the little edit button in top right. Minimal text could solve that.

Robinlovelace:


dabreegster (approved):
I agree the polygon tool is not fluid or obvious to use. I suspect we'll switch plugins or roll our own to solve the underlying design problem here. Instructions in the meantime are good, thanks!

PR #6, opened 2022-10-22
Proof-of-concept: snap routes to roads

dabreegster:
Route-based schemes will usually follow existing roads. To make it as smooth as possible for people to sketch, this PR introduces a tool to snap to roads and intersections.

https://user-images.githubusercontent.com/1664407/197337513-a3b6c631-cb0c-4658-9aec-2f466fe4fde7.mp4

## How it works

This is a "port" of https://github.com/a-b-street/abstreet/blob/master/apps/game/src/common/route_sketcher.rs from bike.abstreet.org. It reads in a serialized A/B Street map, which does the pathfinding between waypoints and generates the road center-line geometry. The business logic of snapping and responding to input events happens in Rust. A small JS shim makes it work alongside https://github.com/mapbox/mapbox-gl-draw.

## Limits/bugs

This is hardcoded to work for a small area in Bristol right now, by shoving 30MB of a map file into the WASM binary! I wanted to at least HTTP fetch a map for this first step, but I'm hitting bizarre serde issues going from JS to WASM right now. Longer term, we need to figure out how to make this work anywhere. A simple, but slow, short-term option is to dynamically fetch OSM data from Overpass, feed into osm2streets, and port the simple pathfinding logic to work off of that instead. https://github.com/a-b-street/abstreet/issues/1009 has more relevant ideas for scaling nation-wide.

You can only draw one snapped route right now. You can click on the result and edit the form as usual. If you then refresh the page, the snapped route acts like a normally drawn one (managed by mapbox-gl-draw) and you can't drag the points around anymore. And if you start a new snapped route, it overwrites the previous route, because they use a fixed ID. I need to figure out how to make the tool work seamlessly with multiple routes at a time. It's tricky, because there's extra state about the waypoints. Do we save that in the geojson as extra info so it survives page reloads, or do we snap just the linestring back to intersections later? (Probably the second!)

Once we solve the above problems and make it scale everywhere easily, this would be a great standalone maplibre plugin to show off osm2streets. CC @BudgieInWA

## Next steps

I'll merge this PR, check the github actions works correctly, and post the demo link here. (And then keep iterating on the above problems)


route-snapper/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+abstutil = { git = "https://github.com/a-b-street/abstreet" }
+console_error_panic_hook = "0.1.6"
+geojson = "0.24.0"
+geom = { git = "https://github.com/a-b-street/abstreet" }
+getrandom = { version = "0.2.3", features = ["js"] }
```
dabreegster, line 14: There's a bunch of junk dependencies added in for not great reasons. `map_model` pulls in too much and needs extra workarounds on web. I might trim this down, but probably just making this work off osm2streets is the better next step

route-snapper/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+#[wasm_bindgen]
+impl JsRouteSnapper {
+    #[wasm_bindgen(constructor)]
+    pub fn new() -> Result<JsRouteSnapper, JsValue> {
+        // Panics shouldn't happen, but if they do, console.log them.
+        console_error_panic_hook::set_once();
+
+        let map: Map = abstutil::from_binary(include_bytes!("../bristol.bin"))
```
dabreegster, line 35: Here's the hack! https://github.com/acteng/atip/tree/route-snapper-fetch-gunzip uses a dynamic fetch, but serde is blowing up somehow

route-snapper/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+                .map
+                .get_r(self.map.find_road_between(pair[0], pair[1]).unwrap());
+            pairs.push((
+                r.center_pts.to_geojson(gps_bounds),
+                make_props("type", "confirmed route"),
+            ));
+        }
+        for i in &self.route.full_path {
+            pairs.push((
+                Circle::new(self.map.get_i(*i).polygon.center(), INTERSECTON_RADIUS)
```
dabreegster, line 70: Passing back tessellated circles is probably silly, we could just send back a single point and use maplibre styling to draw it nicely.

route-snapper/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+            id: Some(geojson::feature::Id::String("snappy-route".to_string())),
+            properties: None,
+            foreign_members: None,
+        };
+        Some(serde_json::to_string_pretty(&feature).unwrap())
+    }
+
+    // True if something has changed
+    #[wasm_bindgen(js_name = onMouseMove)]
+    pub fn on_mouse_move(&mut self, lon: f64, lat: f64) -> bool {
```
dabreegster, line 143: There's probably too much UI logic in here. mapbox draw plugins handle this kind of thing in more natural ways. The responsibility of the Rust lib should probably shrink dramatically

dabreegster:
Live at https://acteng.github.io/atip/routes.html?authority=West%20of%20England%20Combined%20Authority#14.29/51.44965/-2.57773

PR #9, opened 2022-10-24
Layout the sidebar/form more ergenomically and improve the style a bit

BudgieInWA:


dabreegster:
Such awesome improvements, thanks! Looking through the code now to make sure I follow everything...

https://user-images.githubusercontent.com/1664407/197536273-f9d64323-6187-42e8-81d9-8bd7b781f3fc.mp4

BudgieInWA:
Oh, I forgot a call to `app.map.resize()` after changing the content of `#panel` (setting and clearing).

map_with_sidebar.css
```diff
@@ -1,22 +1,77 @@
...
+  max-width: 25rem;
+}
+.sidebar-content {
+  width: 25em; /* Keep the size constant, so the map overlaps it when it collapses. */
+
+  max-height: 100%;
+  overflow-y: auto;
+}
+body > aside.collapsed {
+  min-width: 1.5rem;
```
dabreegster, line 42: Screenshot from 2022-10-24 14-33-28
This looks a bit weird to me; I might adjust this later to totally hide it

map_with_sidebar.css
```diff
@@ -1,22 +1,77 @@
...
+  display: none;
 }
 
+/* The map */
 #basemaps-wrapper {
+  z-index: 1;
   position: absolute;
-  bottom: 100px;
-  right: 20px;
+  bottom: 80px;
```
dabreegster, line 73: In A/B Street-land, we've been careful about things based on pixels, because they scale with a monitor's DPI (Mac Retina displays have been a source of endless bugs). Playing around with my OS's display scale factor, things look reasonable, so it's cool to see that browsers handle this pretty well

areas.html
```diff
@@ -139,47 +145,48 @@
...
-          <h2>Area Interventions</h2>
-          <p>
-            Draw the boundaries by clicking on draw button (top right of the
-            map). Click on the first point of when the boundary is complete and
-            add data fields.
-          </p>
-          <h1 id="authority"></h1>
-          <p>Population: <span id="population"></span></p>
-          <p>Number of areas drawn: <span id="num_areas">0</span></p>
+          <a href="areas.html" class="authority-link">Areas</a>
```
dabreegster, line 155: It's much nicer to modify these in place, instead of inserting everything from JS

dabreegster (approved):

dabreegster:
> Oh, I forgot a call to app.map.resize() after changing the content of #panel (setting and clearing).

Added! I'm a bit confused why we don't need to call this when creating the form, but it already works as-is -- the position of the attribution legend is fixed. Maybe it handles the container shrinking, but not growing, without a `resize()`

dabreegster:
I'll merge this in now. You're a life-saver, thanks!

map_with_sidebar.css
BudgieInWA: "CSS pixels" are a whole thing. They started out defined as representing device pixels, but the definition changed to ... (looking up definition)... 1/96th of 1 inch, sure.

But the long and short of it is, they are just another unit of "absolute length" in CSS land, that will be scaled to look similar in different environments, so you can use them naively.

map_with_sidebar.css
BudgieInWA: Yes, it does look a bit janky. I was going for a nice hint that there is content there to be revealed, but it needs a bit more finesse to look intentional. Maybe a big drop shadow would make it look deliberate. Maybe `.sidebar-content { position: absolute; right: 0; }`, so that the content is hanging off the edge of the screen and just poking in would do the trick.

Or `min-width: 0`, shift the toggle a bit to the right and call it a day :)

BudgieInWA:
> > Oh, I forgot a call to app.map.resize() after changing the content of #panel (setting and clearing).
> 
> Added! I'm a bit confused why we don't need to call this when creating the form, but it already works as-is -- the position of the attribution legend is fixed. Maybe it handles the container shrinking, but not growing, without a `resize()`

When the form is created without the `resize()` the map canvas is not changing size, the form is actually covering the bottom portion of it. The map canvas is actually sticking out of its container, but the basemaps selector and attribution are positioned relative to the container, not the canvas, so it was hard to notice! 

When the map `resize()`s you can notice the map is panned to keep the center in the center, which is a bit nicer.

dabreegster:
> When the form is created without the resize() the map canvas is not changing size, the form is actually covering the bottom portion of it.

Ahh, subtle! Fixed in fddd2f8b4d963aff6408b586fc0e20f4319c63c8

PR #14, opened 2022-10-27
Use a custom file for route snapper, per authority

dabreegster:
Instead of loading a fixed A/B Street map file for the route snapper, now we dynamically fetch a file. Its contents are a new binary-encoded format, created in https://github.com/acteng/abstreet-to-atip/blob/main/src/route_snapper.rs. This just has the geometry and graph structure needed to do route snapping, so it's way smaller than the abst map.

This gets things working for a few more areas. The files are hosted in my abst S3 bucket right now, and manually created/uploaded with scripts in the other repo. I need to do more work to clean up that process and run it for all authorities.

Still loads of cleanup needed. I want to make a separate git repo, just for this new route snapping plugin. It has nothing to do with ATIP, other than being needed by it. :)

### DIFF
--- a/routes.html
+++ b/routes.html
@@ -45,7 +45,18 @@
         saveForm
       );
 
-      window.routeSnapper = new RouteSnapper(window.app.map, drawControls);
+      async function setupRouteSnapper() {
+        const url = `http://abstreet.s3-website.us-east-2.amazonaws.com/route-snappers/${window.app.authority}.bin`;
+        console.log(`Grabbing ${url}`);
+        const resp = await fetch(url);
+        const mapBytes = await resp.arrayBuffer();
+        window.routeSnapper = new RouteSnapper(
+          window.app.map,
+          drawControls,
+          new Uint8Array(mapBytes)
+        );
+      }
+      await setupRouteSnapper();
 
       document.getElementById("export").onclick = () => {
         window.app.downloadGeojsonFile();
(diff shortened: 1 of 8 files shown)


routes.html
```diff
@@ -45,7 +45,18 @@
...
-      window.routeSnapper = new RouteSnapper(window.app.map, drawControls);
+      async function setupRouteSnapper() {
+        const url = `http://abstreet.s3-website.us-east-2.amazonaws.com/route-snappers/${window.app.authority}.bin`;
+        console.log(`Grabbing ${url}`);
+        const resp = await fetch(url);
+        const mapBytes = await resp.arrayBuffer();
+        window.routeSnapper = new RouteSnapper(
+          window.app.map,
+          drawControls,
+          new Uint8Array(mapBytes)
```
dabreegster, line 56: The WASM bridge stuff is confusing. If I just pass the ArrayBuffer here, everything "works", but over in Rust, we see 0 bytes

dabreegster:
Try it out at https://acteng.github.io/atip/routes.html?authority=Bedford#1/0/0. Only a handful of areas work, I'm on slow internet and need to work more on importing all areas

dabreegster:
Ahh actually, https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content doesn't let this work with github pages. S3 is http, pages is https. We'll host consistently on one or the other.

PR #18, opened 2022-10-29
16 minor updates to fields based

Robinlovelace:


dabreegster:
Thanks for getting this started! `input type="textentry"` doesn't quite work. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea has examples. Fixed things up

PR #24, opened 2022-11-07
23 interventions

### LINKED ISSUE #17: Consistent language in copy: intervention not scheme
In https://acteng.github.io/schema/v1 changes are referred to as interventions. Scheme is also a commonly used term but I think Intervention may be better as it can cover a wider range of things and has fewer meanings.

### LINKED ISSUE #23: Load the route snapper in a web worker
Many perf improvements to the snap tool needed, but one short-term might be to do the initial load in a web worker. Otherwise the map loading will hang

Robinlovelace:
- Rename lowercase scheme text
- Rename Titlecase words

Close #17

### DIFF
--- a/app.js
+++ b/app.js
@@ -328,14 +328,14 @@ function sidebarEntry(props) {
 export function makeCommonFormFields(props) {
   return `
           <div class="form-row">
-            <label for="intervention_name">Scheme name:</label>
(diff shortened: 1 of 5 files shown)


dabreegster (approved):
Thanks!

Issue #21, opened 2022-11-07
Unified UI with Intervention type not in tabs


Robinlovelace:
For each intervention we definitely want:

MVP:

- [ ] Name (freeform)
- [ ] Intervention type
- [ ] Description (optional)

Details (outdated)

<details>

- [ ] [areas](https://github.com/acteng/schema/blob/main/schemas/areas.json)
- [ ] [crossings](https://github.com/acteng/schema/blob/main/schemas/crossings.json)
- [ ] [furniture](https://github.com/acteng/schema/blob/main/schemas/furniture.json)
- [ ] [other](https://github.com/acteng/schema/blob/main/schemas/other.json)
- [ ] [routes](https://github.com/acteng/schema/blob/main/schemas/routes.json)

Currently users select which with a tab as shown below, unifying the UI will have various benefits.

![image](https://user-images.githubusercontent.com/1825120/200360683-3d5ef881-7f98-41b7-9c03-84528bf1699d.png)

Tricky part: making the UI reactive to the different types of intervention.

dabreegster:
Ideally the right bar would have a different tool to start creating each type of thing. It'll be a little tricky to work around mapbox-gl-draw's weird API.

On the left, we could have a little icon to represent each type of intervention in the list. And probably simple checkboxes to filter that list.

dabreegster:
Budget and completion year are now scheme-wide, not per intervention

PR #28, opened 2022-11-16
One intervention page

dabreegster:
This is most of the way towards #21. The per-type detailed form is totally gone for now. Lots more tweaks needed, but going to deploy this now and iterate


Robinlovelace:
Looking good 
image

PR #33, opened 2022-11-18
Instructions popup

Robinlovelace:
Largely duplicate of #32 but with instructions more neatly in popup, not popping-up tho for some reason. I votethis way if we get it working so the app is neater, although a bit ugly buttons atm:

image

### DIFF
--- a/app.js
+++ b/app.js
@@ -51,6 +51,7 @@ export class App {
     this.#setupMap(setCamera);
 
     setupCredits();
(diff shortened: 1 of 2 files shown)


dabreegster:
You were just missing one bit, making a `div` with the right ID. If you open your browser console and click the button, the error message would be a good clue there.

This is a great step forward, thanks!

PR #39, opened 2022-11-20
UI tweaks

Robinlovelace:


scheme.html
```diff
@@ -48,10 +48,7 @@
         <h1 id="authority">Loading...</h1>
         <label for="scheme_name">Scheme name:</label>
         <input type="text" id="scheme_name" />
-        <button type="button" id="export" style="margin-top: 6px">
-          Export to GeoJSON
-        </button>
-        <p>Load from GeoJSON: <input type="file" id="load_geojson" /></p>
+        <p>Add interventions that make up the scheme:</p>>
```
dabreegster, line 51: Extra trailing > here btw

scheme.html
```diff
@@ -84,14 +81,19 @@ <h1 id="authority">Loading...</h1>
...
       </div>
-
       <div
         class="sidebar-toggle rounded-rect"
         onclick="document.getElementById('sidebar').classList.toggle('collapsed'); app.map.resize();"
       >
         &rarr;
       </div>
-    </aside>
+      <div class="sidebar-content content-container">
```
dabreegster, line 90: I don't think we want this <div> again. The two items should be at the same "level" they were both, just moved down

dabreegster:
Done by 8cac3a7dacd571c1c36b8c9094f985e9b4ddc749

PR #54, opened 2022-12-23
Switch to using Vite and move the route-snapper out

dabreegster:
Two related changes here:

1) Move the route snapper to https://github.com/dabreegster/route_snapper/. It's not specific to ATIP at all, and a separate repository will make it much faster to iterate on it, get wider adoption, and encourage contributors.
2) Use Vite to manage dependencies

The second is a somewhat big shift. Previously I've been avoiding all JS frameworks and build tools entirely, because the ecosystem is kind of a mess and the learning curve is higher. After some experimentation elsewhere, I'm very happy with Vite (after fighting some problems getting WASM modules working...). Now we can manage external dependencies in a normal way for modern web dev.


app.js
```diff
@@ -1,7 +1,16 @@
...
 
+import maplibregl from "maplibre-gl";
+import "maplibre-gl/dist/maplibre-gl.css";
+import geojsonExtent from "@mapbox/geojson-extent";
+import MapboxDraw from "@mapbox/mapbox-gl-draw";
+import "@mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.css";
+import { RouteSnapper, fetchWithProgress } from "route-snapper/lib.js";
+
+// Do this after mapbox-gl-draw.css to override some stuff
+import "./css/map_controls.css";
```
dabreegster, line 11: This is a little magic to me still. If you want some CSS included in a web page, you can just import it in a `.js` and vite does some kind of bundling magic. It just works.

package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "atip",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite preview",
+    "fmt": "prettier --write *.html *.js css/*"
```
dabreegster, line 10: This is using globally installed prettier right now, it should really be a dev dependency. Will clean later

app.js
Robinlovelace: Nice!

package.json
```diff
@@ -0,0 +1,23 @@
...
+  "devDependencies": {
+    "vite": "^4.0.0",
+    "vite-plugin-top-level-await": "^1.2.2",
+    "vite-plugin-wasm-pack": "0.1.11"
+  },
+  "dependencies": {
+    "@mapbox/geojson-extent": "^1.0.1",
+    "@mapbox/mapbox-gl-draw": "^1.3.0",
+    "maplibre-gl": "^2.4.0",
+    "route-snapper": "^0.1.0"
```
Robinlovelace, line 21: :tada:

Robinlovelace (commented):
Amazing refactor by the looks, good lesson in modularity and refactoring for maintainaiblity and more.

Issue #65, opened 2023-01-12
Detect junk input


dabreegster:
https://github.com/acteng/atip/blob/5ff0942dc3790cda6c5cc302ae0314447be8e221/components/EntireScheme.svelte#L66

You can load an arbitrary file. If it's not JSON, it'll fail reasonably. But otherwise, we don't even constrain the input to be GeoJSON. There's many more things we should require (no CRS on the geojson, no unexpected nested feature collections). We should probably just ignore unknown or invalid properties, but unsure. (Longer term we could come up with a proper schema for the data we collect, with a strategy for backwards compatibility)

Robinlovelace:
Trying this in session with @sparrow0hawk. Test input:

```
echo 'junk-input' > test-junk.txt
```

Robinlovelace:
Current behaviour is really not bad: detects that it's not valid json.

![image](https://user-images.githubusercontent.com/1825120/212329219-cb3d47d1-8af7-436c-99be-2273fe691e3c.png)

Robinlovelace:
Ha, OK wasn't expecting this!

![image](https://user-images.githubusercontent.com/1825120/212331691-d6a8954b-048f-46c3-861d-71559caa8325.png)

Sparrow0hawk:
Tested with non ATIP geojson ([source](https://raw.githubusercontent.com/Sparrow0hawk/mynearestleedsmarginal/main/assets/data/leeds_wards_2018.geojson))

Robinlovelace:
I see this as a low priority. Key question: how to detect non ATIP input?

Sparrow0hawk:
Could adjust the schema for ATIP-generated geojson files to include a high level field specifying ATIP-generated. Could then validate this within `loadFile`.

dabreegster:
While testing this, I thought of another mistake people could make: loading a file from a different authority. We could check for this and ban it, but maybe there are valid reasons why someone would want to load an adjacent scheme? Or we could prompt them to go load the scheme in the other page?

Robinlovelace:
I don't think we should put up too many guardrails. There may well be valid reasons: looking at what the neighbours have planned is a good way to up you game! No harm in a prompt but the fact that the map zooms to a new place will surely be a prompt in itself?

dabreegster:
Flagging https://zod.dev/ as possibly interesting for this issue and maybe our schemas/forms

PR #66, opened 2023-01-13
Defining origin of generated geojson files

Sparrow0hawk:
This PR attempts to start to address #65 

In particular including some validation to check an uploaded geojson is one that was generated from the ATIP platform.

Summary of changes:
- Adding an origin field to generated geojson that specifies `atip-v1`
- Adding a validation step in `loadFile` that checks for this origin field

Outstanding tasks:
- [ ] Make backwards compatible

### DIFF
--- a/components/EntireScheme.svelte
+++ b/components/EntireScheme.svelte
@@ -77,15 +77,11 @@
         if (result["origin"] || result.hasOwnProperty("authority")) {
           gjScheme.set(result);
         } else {
-          console.log(
-            `Non-ATIP based geojson uploaded. Please try again with an ATIP generated file.`
-          );
           window.alert(
             `Non-ATIP based geojson uploaded. Please try again with an ATIP generated file.`
           );
         }
       } catch (err) {
-        console.log(`Couldn't load from a file: ${err}`);
         window.alert(`Couldn't load scheme from a file: ${err}`);
       }
     };
--- a/components/EntireScheme.svelte
+++ b/components/EntireScheme.svelte
@@ -69,13 +69,9 @@
     if (geojson.hasOwnProperty("origin")) {
       let origin = geojson.origin;
       return origin.startsWith("atip-");
-    } else if (
-      geojson.hasOwnProperty("authority") &&
-      !geojson.hasOwnProperty("origin")
-    ) {
-      return true;
     } else {
-      return false;
+      return geojson.hasOwnProperty("authority");
+    }
     }
   }
 
--- a/components/EntireScheme.svelte
+++ b/components/EntireScheme.svelte
@@ -73,9 +73,8 @@
       try {
         let result = JSON.parse(e.target.result);
 
-        // test if origin field is set
-        // if not then file is not from atip
-        if (result["origin"]) {
+        // check for origin or authority foreign member to confirm atip generated
+        if (result["origin"] || result.hasOwnProperty("authority")) {
           gjScheme.set(result);
         } else {
           console.log(
--- a/components/EntireScheme.svelte
+++ b/components/EntireScheme.svelte
@@ -71,7 +71,17 @@
     // TODO Should we prompt before deleting the current scheme?
     reader.onload = (e) => {
       try {
-        gjScheme.set(JSON.parse(e.target.result));
+
+        let result = JSON.parse(e.target.result);
+
+        // test if origin field is set
+        // if not then file is not from atip
+        if(result['origin']) {
+          gjScheme.set(result);
+        } else {
+          console.log(`Non-ATIP based geojson uploaded. Please try again with an ATIP generated file.`);
+          window.alert(`Non-ATIP based geojson uploaded. Please try again with an ATIP generated file.`);
+        }
       } catch (err) {
         console.log(`Couldn't load from a file: ${err}`);
         window.alert(`Couldn't load scheme from a file: ${err}`);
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Is it backwards compatible? Should work for scheme files created before this change. That's my only thought, apologies probably could test it but limited time!

Sparrow0hawk:
True, this isn't backwards compatible in it's current implementation. I shall have a think about that!

Sparrow0hawk:
My hunch to make it backwards compatible would be to check for authority at the top level as that should be present in all files. But might also be better to check for some of the `properties` keys like `intervention_type`.

components/EntireScheme.svelte
```diff
@@ -18,6 +18,8 @@
 
   gjScheme.update((gj) => {
     gj.authority = authorityName;
+    // we could probably be more sophisticated here and set version more centrally
```
dabreegster, line 21: You chose well, this is currently the central place to do things like this. It's maybe a bit misleading how this `update` always clobbers things, though.

components/EntireScheme.svelte
```diff
@@ -69,7 +71,17 @@
     // TODO Should we prompt before deleting the current scheme?
     reader.onload = (e) => {
       try {
-        gjScheme.set(JSON.parse(e.target.result));
+
+        let result = JSON.parse(e.target.result);
+
+        // test if origin field is set
```
dabreegster, line 77: Style nit-picks:

1) In the final code we check in, I would vote against these comments, since they just repeat what the code says. A helpful thing to comment would be why checking for this field is sufficient to decide this comes from ATIP
2) If you run `npm run fmt`, it'll add some standardized spacing
3) Don't we want to check `result.origin` starts with `atip-`? It's unlikely someone has a geojson file with this [foreign member](https://www.rfc-editor.org/rfc/rfc7946#section-6) set, but still seems useful

components/EntireScheme.svelte
```diff
@@ -69,7 +71,17 @@
...
+        // if not then file is not from atip
+        if(result['origin']) {
+          gjScheme.set(result);
+        } else {
+          console.log(`Non-ATIP based geojson uploaded. Please try again with an ATIP generated file.`);
```
dabreegster, line 82: We could probably remove this and the below `console.log` now, I mistakenly left them in from a previous debugging session

dabreegster (commented):
Welcome to the project, and thanks for the PR!

The backwards compatibility check is a good question. I think just checking for `authority` is fine for now; we don't need to go more specific than that probably. We could go all-out and insist on **only** the properties we expect, with things like the 4 possibilities for `intervention_type`. But it's overkill at this stage; we need to have a bigger conversation about the long-term schema, what format we use for it (something like protobuf?), and how to evolve it over time

components/EntireScheme.svelte
dabreegster: Oops, I just realized `npm run fmt` is assuming a globally installed copy of `prettier` exists. I'll fix that...

dabreegster: If you sync past 50bc1d0126b1dd28643fc1bd7c424ad34be56f8a, `npm run fmt` now does the right thing :)

components/EntireScheme.svelte
Sparrow0hawk: It felt like the right place at this stage but happy to take suggestions on whether we do this elsewhere.

components/EntireScheme.svelte
Sparrow0hawk: Removed these in new commits

components/EntireScheme.svelte
Sparrow0hawk: Merged my branch up past 50bc1d0 and run `npm run fmt`

I've also added an or condition to check for `authority` to allow for backwards compatibility.

On point 3, yes we probably should do that. I couldn't implement it nicely just now (can't check the string if `origin` doesn't exist to be backwards compatible) but will have a think on it.

Sparrow0hawk:
Thought some more about this, particularly handling checking the `origin` foreign member and have decided to best handle this with a function. Look forward to your thoughts!

components/EntireScheme.svelte
```diff
@@ -63,15 +65,37 @@
...
+    if (geojson.hasOwnProperty("origin")) {
+      let origin = geojson.origin;
+      return origin.startsWith("atip-");
+    } else if (
+      geojson.hasOwnProperty("authority") &&
+      !geojson.hasOwnProperty("origin")
+    ) {
+      return true;
+    } else {
+      return false;
```
dabreegster, line 78: We can simplify this in the else block; we know origin isn't set

```suggestion
    } else {
      return geojson.hasOwnProperty("authority");
    }
```

components/EntireScheme.svelte
```diff
@@ -63,15 +65,33 @@
     document.body.removeChild(element);
   }
 
+  function checkOrigin(geojson) {
+    if (geojson.hasOwnProperty("origin")) {
+      let origin = geojson.origin;
+      return origin.startsWith("atip-");
+    } else {
+      return geojson.hasOwnProperty("authority");
+    }
```
dabreegster, line 74:
```suggestion
```

Oops, fixing the suggestion in github

dabreegster (approved):
Thanks for the PR! I'll merge this now

PR #74, opened 2023-02-09
fix #71: validation logic for user authority selection

### LINKED ISSUE #71: Submitting a random string on the home page select LAD box returns a blank map
Whilst looking through the log data I saw a couple of entries that didn't match available LADs and Transport authorities on the app.

It looks like some users might type in their authority and hit `Start` without selecting from the presented drop-down.

E.g. https://atip.uk/scheme.html?authority=narnia#1/0/0 

Screenshot 2023-02-08 at 11 32 21

## Steps to reproduce

1. Visit https://atip.uk/index.html
2. In the `Select Transport Authority or Local Authority District:` box enter your own random string
3. Click Start

## Suggested fix

I think we'd want to see if we could write in some user validation at the select step, just to make sure they are constrained to available LADs/authorities.

Sparrow0hawk:
This adds some logic for transforming the `dataList` into an array which we can check if the `inputList.value` exists within.

Need to decide on how to notify the user of their mistake.

### DIFF
--- a/ChooseArea.svelte
+++ b/ChooseArea.svelte
@@ -89,7 +89,11 @@
   loadAuthorities();
 
   function start() {
-    if (inputList.value) {
+    let auth_names = Array.from(dataList.childNodes.values()).map(function (x) {
+      return x.value;
+    });
+    console.log();
+    if (auth_names.includes(inputList.value)) {
       window.location.href = `scheme.html?authority=${inputList.value}`;
     }
   }
--- a/ChooseArea.svelte
+++ b/ChooseArea.svelte
@@ -89,9 +89,9 @@
   loadAuthorities();
 
   function start() {
-    let auth_names = Array.from(dataList.childNodes.values()).map(function (x) {
-      return x.value;
-    });
+    let auth_names = Array.from(dataList.childNodes.values()).map(
+      (x) => x.value
+    );
     console.log();
     if (auth_names.includes(inputList.value)) {
       window.location.href = `scheme.html?authority=${inputList.value}`;
--- a/ChooseArea.svelte
+++ b/ChooseArea.svelte
@@ -11,6 +11,7 @@
 
   let inputList;
   let dataList;
+  let warning = false;
 
   async function loadAuthorities() {
     let source = "boundary";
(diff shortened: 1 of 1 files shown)


Sparrow0hawk:
@dabreegster would appreciate your input and how we'd like to notify the user of a mistake? A modal popup or inject some warning text underneath? As at the moment it just does nothing which I don't think is adequate.

ChooseArea.svelte
```diff
@@ -89,7 +89,11 @@
   loadAuthorities();
 
   function start() {
-    if (inputList.value) {
+    let auth_names = Array.from(dataList.childNodes.values()).map(function (x) {
```
dabreegster, line 92: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions may be more readable here: `(x) => x.value`

dabreegster (commented):
Hmm, to tell people about the invalid input, we could

1) Use `window.alert` or the `Modal` component from carbon (see About/Instructions)
2) Disable the start button and have some warning text

I don't know UX best practices, but the 2nd seems nicer to me. It could update whenever the input field changes to show or hide the warning and enable or disable the button. Svelte should make this very declarative to write -- we could use a reactive var that means "inputList.value is currently valid", and change the button state and visibility of some warning text accordingly

Sparrow0hawk:
There's probably a much nicer, trimer way of including the warning message but have a look and let me know what you think.

ChooseArea.svelte
Sparrow0hawk: Resolved in 06ff173

dabreegster:
This gets the job done, but I find the UX a bit confusing, because the warning only resets when you click Start:
Screenshot from 2023-02-13 09-42-14
Svelte should make it pretty easy for us to change the button state and show/hide the warning when the input field changes. I'll have a go...

dabreegster:
It felt cleaner to open #75 instead of pushing more commits to this branch; let me know what you think

Sparrow0hawk:
Closing in favour of #75

PR #75, opened 2023-02-13
Fix #71 by toggling the start button and showing a warning for invali…

### LINKED ISSUE #71: Submitting a random string on the home page select LAD box returns a blank map

dabreegster:
…d input

https://user-images.githubusercontent.com/1664407/218427282-29e56454-0e7e-406d-9541-189eb49dd678.mp4

### DIFF
--- a/ChooseArea.svelte
+++ b/ChooseArea.svelte
@@ -9,8 +9,11 @@
 
   let showAbout = false;
 
-  let inputList;
+  let inputValue;
   let dataList;
+  let authoritySet = new Set();
+
+  $: validEntry = authoritySet.has(inputValue);
 
   async function loadAuthorities() {
     let source = "boundary";
@@ -23,6 +26,7 @@
       let option = document.createElement("option");
       option.value = feature.properties.name;
       dataList.appendChild(option);
+      authoritySet.add(feature.properties.name);
     }
 
     // Only show TAs, not LADs, in the map
@@ -89,9 +93,7 @@
   loadAuthorities();
 
   function start() {
-    if (inputList.value) {
-      window.location.href = `scheme.html?authority=${inputList.value}`;
-    }
+    window.location.href = `scheme.html?authority=${inputValue}`;
   }
 </script>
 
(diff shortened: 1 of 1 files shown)


dabreegster:
This is an alternative to #74 that constantly updates the button state and warning text upon every keypress, not just when you click the start button. Any thoughts?

Putting the warning next to the button felt a bit more clear to me than a new paragraph, but not really sure where it belongs

ChooseArea.svelte
```diff
@@ -9,8 +9,11 @@
 
   let showAbout = false;
 
-  let inputList;
+  let inputValue;
   let dataList;
+  let authoritySet = new Set();
+
+  $: validEntry = authoritySet.has(inputValue);
```
dabreegster, line 16: https://svelte.dev/tutorial/reactive-declarations makes it trivial for us to recalculate this when the input changes

ChooseArea.svelte
```diff
@@ -103,9 +105,16 @@
   <br />
 
   <p>Select Transport Authority or Local Authority District:</p>
-  <input list="authorities-list" bind:this={inputList} />
-  <datalist id="authorities-list" bind:this={dataList} />
-  <button type="button" on:click={start}>Start</button>
+  <div>
+    <input list="authorities-list" bind:value={inputValue} />
```
dabreegster, line 109: I realized it was easier to just `bind:value` instead of binding a DOM element and then doing `x.value`

Robinlovelace:
Taking a look now...

Robinlovelace:
Looking at the video you posted Dustin and trying what's currently on atip.uk, I approve of this, seems like a good solution.

Robinlovelace (approved):

Sparrow0hawk:
Much sveltier! 

My only comment would be should we just drop the warning text and have the button disabled until someone picks a valid option from the list?

But that's just a personal UX take.

Robinlovelace:
Nice adjective: svelty!

Robinlovelace:
Nice adjective: svelty!

dabreegster:
Interpreting closure of #74 as being in favor of this, so merging

dabreegster:
Oops, wifi here is weird and didn't pick up new messages, but let me submit?

> My only comment would be should we just drop the warning text and have the button disabled until someone picks a valid option from the list?

Agreed... it's clear enough that the button is disabled. Done in e7f70ba37007cd5fb14e2526b0f90eb2618036f8

PR #76, opened 2023-02-13
Update route-snapper. Users can now prevent routes from doubling back,

dabreegster:
and new routes drawn will have a length_meters property

https://user-images.githubusercontent.com/1664407/218472870-0d2a75b0-2b1a-4075-a830-ced88b656b1a.mp4

For convenience, you can play with the new option at https://dabreegster.github.io/route_snapper/ too. Note this change also starts adding a `length_meters` property to the GeoJSON. Files people submit may or may not contain this field, since people might've used an old version of ATIP. We can easily "backfill" this on the analysis side if needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2073,9 +2073,9 @@
       }
     },
     "node_modules/route-snapper": {
-      "version": "0.1.5",
-      "resolved": "https://registry.npmjs.org/route-snapper/-/route-snapper-0.1.5.tgz",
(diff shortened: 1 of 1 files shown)


dabreegster:
And for the record, `npm update route-snapper` is enough to update the lockfile (that's how I generated this PR)

Sparrow0hawk (approved):
This also works nicely for me

Robinlovelace:
Great new feature!

Issue #72, opened 2023-02-09
Issue uploading a .txt file


robinlovelace-ate:
As reported via email there is a reproducible upload issue shown below.

![image](https://user-images.githubusercontent.com/122299965/217791208-58fb94dc-aff8-49e1-894d-5533c9a52514.png)

Reproducible example: navigate to https://atip.uk/scheme.html?authority=Norfolk#8.45/52.673/0.949 and upload the attached file.
[Norfolk_scheme 2 Dereham Quebec Rd.txt](https://github.com/acteng/atip/files/10696227/Norfolk_scheme.2.Dereham.Quebec.Rd.txt)

robinlovelace-ate:
I have tried to reproduce the issue with a simple file but cannot. See attached minimal example.
[Uploading Norfolk.txt…]()

PhillReilly:
tried again with this file and still not working
[Norfolk_Scheme 2 B1146 Quebec Rd Dereham.txt](https://github.com/acteng/atip/files/10696448/Norfolk_Scheme.2.B1146.Quebec.Rd.Dereham.txt)

dabreegster:
I can reproduce the failure. That file is failing [checkOrigin](https://github.com/acteng/atip/blob/84e37a50e25f111c1a6ab2003347db3609b33397/components/EntireScheme.svelte#L68). It doesn't have an `authority` or `origin` property at the top level of the JSON object.

@PhillReilly, did you produce this file with ATIP, or another program?

PhillReilly:
i used ATIP with link from active travel england - i plotted the object named and saved, then exported to GeoTSON and saved the text file produced.  then try to reload it form the saved location (its a networked drive rather than laptop desktop) and the error appears - ju
![image](https://user-images.githubusercontent.com/124881576/217799267-295da1eb-e235-4056-9547-009f23252626.png)
st tried again to no avail

PhillReilly:
odd that this one works
[Norfolk_B1370 Church Lane 6th From College.txt](https://github.com/acteng/atip/files/10696533/Norfolk_B1370.Church.Lane.6th.From.College.txt)

PhillReilly:
browser is MS edge

robinlovelace-ate:
```
  "authority": "Norfolk",
  "origin": "atip-v1"
```

Add those new lines in the final 2 lines just before

```
}
```

And I predict: it will work 🤞

robinlovelace-ate:
[...]

Issue #78, opened 2023-02-15
Issues loading a .TXT file 


WilliamPhillips-WCC:
I was working on ATIP yesterday and saved by exporting the GeoJSON FILE from the tool.

When trying to load this back in this morning to continue my work it will not load.

I saw from the workshop chat things were "fixed" on this yesterday and also followed the solutions in thread here https://github.com/acteng/atip/issues/72 but it is still not working.

This is what I am getting with the file which was exported yesterday afternoon 

![image](https://user-images.githubusercontent.com/125447325/218983974-7ebdfeee-4293-4a55-8d1c-ef27e4e005d4.png)

And this is what I am getting with the same file that I altered the last two lines of text as suggested here https://github.com/acteng/atip/issues/72 

![image](https://user-images.githubusercontent.com/125447325/218984591-cf9317c3-6eb1-4c69-924b-6cc22bfa832d.png)

I have also attached the two .txt files

[Warwickshire_A47 Hinckley Road.txt](https://github.com/acteng/atip/files/10741344/Warwickshire_A47.Hinckley.Road.txt)

[Warwickshire_A47 Hinckley Road - with edit.txt](https://github.com/acteng/atip/files/10741335/Warwickshire_A47.Hinckley.Road.-.with.edit.txt)

Is there still an issue with the tool or is it somthing our end?

dabreegster:
Hi,

I can reproduce this problem. I think you might've gotten unlucky and used the old version (before ~11am yesterday) to start this file and used "Clear all" (triggering the bug), so it was missing some fields. Your edit to fix manually is almost correct; it's just missing a comma at the end of the `"scheme_name": "A47 Hinckley Road"` line.

You can use this corrected file:

[Warwickshire_A47.Hinckley.Road.txt](https://github.com/acteng/atip/files/10741884/Warwickshire_A47.Hinckley.Road.txt)

Sorry for the hassle!

We'll keep this issue open for now. @Sparrow0hawk, maybe we should totally disable the validation of origin fields for now, since more people might also have files missing the field.

WilliamPhillips-WCC:
I used the tool from about 2:30pm yesterday if it helps

Sparrow0hawk:
@dabreegster would support at this stage rolling back the validation step and looking to re-incorporate in a future release.

Robinlovelace:
> @dabreegster would support at this stage rolling back the validation step and looking to re-incorporate in a future release.

:+1:

[...]

PR #80, opened 2023-02-15
adjust validation criteria

Sparrow0hawk:
This PR attempts to resolve issues as seen in #78 where previously downloaded files where schemes were cleared removed foreign members used for validation (origin and authority).

It introduces:
- Validation on either "authority" or "scheme_name"
- migrates the creation of validation foreign members to exportToGeoJson (this is to ensure if you upload an old file where these foreign members ("authority","origin") don't exist they are added)

Tested instances:

✅ Current exported geojson from prod ATIP
✅ Exported geojson from #78 where validation foreign members were cleared
❌ If you exported a .txt when foreign members were cleared and didn't add a scheme name this does not work (but this also doesn't work currently either)

### DIFF
--- a/components/EntireScheme.svelte
+++ b/components/EntireScheme.svelte
@@ -75,7 +75,10 @@
       let origin = geojson.origin;
       return origin.startsWith("atip-");
     } else {
-      return geojson.hasOwnProperty("authority");
+      return (
+        geojson.hasOwnProperty("authority") ||
+        geojson.hasOwnProperty("scheme_name")
+      );
     }
   }
 
--- a/components/EntireScheme.svelte
+++ b/components/EntireScheme.svelte
@@ -16,13 +16,6 @@
     window.localStorage.setItem(authorityName, JSON.stringify(geojsonToSave()))
   );
 
-  gjScheme.update((gj) => {
-    gj.authority = authorityName;
-    // we could probably be more sophisticated here and set version more centrally
-    gj.origin = "atip-v1";
-    return gj;
-  });
-
   function clearAll() {
     if (
       confirm(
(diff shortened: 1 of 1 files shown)


Sparrow0hawk:
This is an alternate approach to completely rolling back validation

Robinlovelace:
Big thumbs up from me on this.

Regarding 

>  If you exported a .txt when foreign members were cleared and didn't add a scheme name this does not work (but this also doesn't work currently either)

that is being tracked in https://github.com/acteng/atip/issues/72 and I think is still not resolved: even after this change people cannot upload this file, right?

https://github.com/acteng/atip/files/10696227/Norfolk_scheme.2.Dereham.Quebec.Rd.txt

Sparrow0hawk:
@Robinlovelace no so that example would actually now work as it does have a scheme_name in the attached file. 

The scenario where someone has cleared the map and then created a new scheme and exported it without a scheme_name feels potentially quite unlikely but it's an edge case I wanted to test.

Robinlovelace:
Great. This PR will close #72 in that case. Even bigger thumbs up!

dabreegster:
How about we remove all the input validation entirely for the moment? It feels less important to catch invalid input than to fix the problems people are hitting now. I would vote for:
- no input validation
- keeping the change to set authority/origin upon export (good idea!)

and we can revisit input validation for a next version, and set up tests more carefully to prevent this

Sparrow0hawk:
> How about we remove all the input validation entirely for the moment? It feels less important to catch invalid input than to fix the problems people are hitting now. I would vote for:
> 
>     * no input validation
> 
>     * keeping the change to set authority/origin upon export (good idea!)
> 
> 
> and we can revisit input validation for a next version, and set up tests more carefully to prevent this

Happy to do that, let me roll an alternate PR with these suggested changes

Robinlovelace:
> How about we remove all the input validation entirely for the moment? It feels less important to catch invalid input than to fix the problems people are hitting now. I would vote for:
> 
>     * no input validation
> 
>     * keeping the change to set authority/origin upon export (good idea!)
> 
> 
> and we can revisit input validation for a next version, and set up tests more carefully to prevent this

:+1: from me too, sounds like that solution will be a simpler fix to #72, unintended fallout from adding validation.

PR #83, opened 2023-02-18
Preliminary play with using playwright for E2E testing

Sparrow0hawk:
This PR is not ready but a quick go at setting up [Playwright](https://playwright.dev/) on this repository and including a basic test showcasing end-to-end testing.


Sparrow0hawk:
Always a good start when the out-the-box github action fails but i've got a good hunch about how to sort that so will resolve.

Sparrow0hawk:
Realised I left off steps for having a go with this PR:

```bash
# assuming you're working with a clone of this repo
git fetch origin pull/83/head:pr-83

git checkout pr-83

# install additional deps
npm install

# do a clean build and run the test suite 
npm test
```

package.json
```diff
@@ -8,16 +8,20 @@
     "build": "vite build",
     "preview": "vite preview",
     "fmt": "npx prettier --write *.html *.js *.svelte components/*",
-    "lint": "npx eslint . --ignore-pattern=vite.config.js"
+    "lint": "npx eslint . --ignore-pattern=vite.config.js",
+    "test": "vite build && vitest"
```
dabreegster, line 12: `build` takes a few seconds... but I think this default makes sense. If you're working on the test code and not ATIP and don't trust the auto-rebuild, `npx vitest` works too.

dabreegster (commented):
Playwright looks perfect!

The first time I ran, I hit a failure:

```
 FAIL  tests/atip-map-interact.spec.js > basic > test if typing a local authority name and clicking start changes the url
Error: expect(received).toHaveURL(expected)

Expected pattern: /.*scheme.html\?authority=West%20Yorkshire%20Combined%20Authority/
Received string:  "chrome-error://chromewebdata/"
```

which looks like some setup flake. Every other time worked fine.

I'm also wondering if you've looked into enabling the HTML reports when running locally. I was a bit surprised that the results just look like they're summarized on the console, without some little local web server wrapping the error logs.

tests/atip-map-interact.spec.js
```diff
@@ -0,0 +1,33 @@
...
+import { expect } from '@playwright/test'
+
+// unstable in Windows, TODO: investigate
+describe.runIf(process.platform !== 'win32')('basic', async () => {
+    let server;
+    let browser;
+    let page;
+
+    beforeAll(async () => {
+        server = await preview({ preview: { port: 4173 } })
```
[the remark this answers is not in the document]
Sparrow0hawk: In [7195e52](https://github.com/acteng/atip/pull/83/commits/7195e52476f5fd5081ad0f7eca62f9c3390c919e) I've tried to add some logic for using a random port within a specified range. 
It's not perfect but would avoid this message, although as you say it doesn't seem to be a major problem but could be as test suite grows.

package.json
Sparrow0hawk: That's a good spot and saves time for me locally!

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
```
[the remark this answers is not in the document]
Sparrow0hawk: Resolved in [08aac7c](https://github.com/acteng/atip/pull/83/commits/08aac7c1de8ca08f9586ae8510a58f3632768871)

Sparrow0hawk:
I haven't looked at outputting a HTML report yet no, although will have a look.

Weird initial error wonder what caused that... I shall ponder

tests/test-utils.js
```diff
@@ -0,0 +1,7 @@
+export function getRandomIntInclusive(min, max) {
```
dabreegster, line 1: Thanks for making this fix, it'll hopefully deflake things! I haven't searched around for vitest / playwright docs yet, but I hope there's something a bit more built-in or recommended floating around. Surely this is a problem anyone running tests encounters. But fine for now.

dabreegster (approved):
No urgency for HTML reporting. I think this PR works fine as-is; how about we merge it and iterate as needed? It should make it easier to write more tests, and through the process of doing that, we can figure out what else from the infrastructure / setup is necessary

tests/test-utils.js
Sparrow0hawk: Yeah I couldn't see anything immediately, I was sort of hoping the vite `preview` call would randomly sort out a port, but it just uses a default.

Sparrow0hawk:
I'll convert this to a live PR to merge

Sparrow0hawk:
Thinking back on this I /think/ we could specify port 0 and the OS will give us a random available port. Should avoid needing to generate a random number, will give it a try.

PR #85, opened 2023-02-21
random port assignment for tests using port 0

Sparrow0hawk:
This PR improves upon the previous PR introducing playwright.

In that PR we randomly allocated a port number for tests using a utility function. A nicer way to set random port numbers is to specify port 0 and allow the OS to randomly assign from an available port for each test.


dabreegster (approved):
Very nice! Completely forgot about port 0. :)  I suppose whoever wrote the vitest examples did too. Thanks!

PR #88, opened 2023-03-18
Simplify end-to-end tests.

dabreegster:
- Remove vitest; playwright as a test runner seems fine
- Start the local dev server using common config
- Consolidate the two tests so far

I started writing a more complex test, and wanted to do it interactively. https://playwright.dev/docs/codegen-intro looks fantastic, but I couldn't figure out how to run all the Playwright commands. I went down a rabbit hole to simplify our e2e test infra. `vitest` doesn't seem to give us any benefit for these e2e tests right now, so I removed it. I also fiddled around with the playwright config to dodge the ES module problems. End result is that we can interactively run Playwright tests and take advantage of all the magic there.

I'm hoping github CI will run for this PR, so I can make sure nothing broke there


README.md
```diff
@@ -17,3 +17,11 @@ You'll need
 [npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm). Run
 `npm install` once. Use `npm run dev` to work locally. To auto-format code,
 `npm run fmt`.
+
+## Tests
+
+Run all tests: `npm run test`
+
+To develop a test interactively, `npm run test -- --debug`. You can step
```
dabreegster, line 25: Screenshot from 2023-03-18 18-41-59
This thing is seriously cool

package.json
```diff
@@ -9,19 +9,18 @@
     "preview": "vite preview",
     "fmt": "npx prettier --write *.html *.js *.svelte components/* tests/*",
     "lint": "npx eslint . --ignore-pattern=vite.config.js",
-    "test": "vite build && vitest"
+    "test": "npx playwright test"
   },
   "devDependencies": {
-    "@playwright/test": "^1.30.0",
+    "@playwright/test": "1.31.2",
```
dabreegster, line 15: There was some github issue (I lost track) for a problem waiting forever for the webserver to start, seemingly fixed in this newer version

playwright.config.js
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from "@playwright/test";
+
+export default defineConfig({
+  webServer: {
+    command: "npx vite --port 8080",
```
dabreegster, line 5: This hardcodes the port again.. if we use playwright test runner to manage the webserver, then I'm not sure how to tell it to autopick port 0 but then figure out what to listen for. Think this is OK for now? Using the test runner to manage the server seems much nicer than each test manually doing it

playwright.config.mjs
```diff
@@ -1,92 +0,0 @@
-// @ts-check
-const { defineConfig, devices } = require('@playwright/test');
-
-/**
```
dabreegster, line 4: This looked like an auto-generated config to me, with a bunch of defaults. I pared down to whatever defaults are in `defineConfig`. Happy to restore anything

playwright.config.mjs
```diff
@@ -1,92 +0,0 @@
-// @ts-check
-const { defineConfig, devices } = require('@playwright/test');
```
dabreegster, line 2: I renamed to `.js` and switched to ES module `import` syntax. This was confusing, I had to copy the TypeScript example, even though we're not using TS. But it somehow makes the normal ES6 module stuff work

tests/homepage.spec.js
```diff
@@ -0,0 +1,22 @@
...
+import { test, expect } from "@playwright/test";
+
+test("front page has welcome text", async ({ page }) => {
+  await page.goto("/");
+  await expect(
+    page.getByRole("heading", { name: "Welcome to ATIP v1" })
+  ).toBeVisible();
+});
+
+test("choosing a local authority and clicking start changes the url", async ({
```
dabreegster, line 11: This still uses a separate browser for each test. Per https://playwright.dev/docs/best-practices#write-fewer-tests-but-longer-tests, we could maybe fold some of these tests together later, especially once we get more and the runtime matters

playwright.config.js
Sparrow0hawk: I think this is fine for now

Sparrow0hawk (commented):
This all looks good to me and works nicely!

Sparrow0hawk (approved):

Issue #90, opened 2023-03-19
Reconsider Carbon


dabreegster:
I chose [Carbon](https://carbon-components-svelte.onrender.com/) as a Svelte component library pretty early on, after trying other options only very briefly. I have a few small complaints about it so far (the stylesheet makes regular HTML defaults look less usable, so for example in `components/Instructions.svelte` we had to switch to their `OrderedList` and `ListItem` components).

Also, will we be required to use https://design-system.service.gov.uk/, and if so, how soon?

Robinlovelace:
> Also, will we be required to use https://design-system.service.gov.uk/, and if so, how soon?

Depends how quickly we want to submit Alpha. Around July is a very rough estimate.

dabreegster:
There's no documented way to tell a https://carbon-components-svelte.onrender.com/components/Modal to be larger?! If there's embedded content that expands/contracts, it's not smart at all about handling a max size.
![Screenshot from 2023-03-27 10-24-03](https://user-images.githubusercontent.com/1664407/227900391-6e7d1cc5-38aa-417f-b206-2054c3d80e62.png)

There is no great reason for us to stick with Carbon, very eager to find something else

dabreegster:
Aaaaand just noticing this beauty:
![Screenshot from 2023-03-27 15-59-14](https://user-images.githubusercontent.com/1664407/227979644-d0d16ca1-68f1-4cd9-81aa-3190bb63d460.png)
When we hide the sidebar, all the buttons and text and stuff correctly disappears and isn't interactive. But the Carbon accordions are still there and clickable

dabreegster:
Another issue I just noticed: the Carbon package isn't hermetic; we're reaching out to the network:
![Screenshot from 2023-04-12 12-01-08](https://user-images.githubusercontent.com/1664407/231438319-79feb8c8-b7a3-4b5c-891d-ef5460dc80e4.png)

dabreegster:
If we can put together our own modal popup element, we're pretty much rid of this

PR #92, opened 2023-03-20
Add a button to recenter the map

dabreegster:
Sometimes I accidentally pan the map somewhere far from the authority's boundary, or otherwise want to return to the default view. This PR adds a button to do that:

https://user-images.githubusercontent.com/1664407/226382006-80e0ea60-721a-4622-86a6-daadf5ae85d6.mp4

I tried to pick an icon and placement that's intuitive, but style advice very welcome.

This PR is easiest to understand by reading one commit at a time. Most of the code change first is switching to using a store for the map. See comments below for motivation


App.svelte
```diff
@@ -39,6 +43,21 @@
     showInstructions = !showInstructions;
     showAbout = false;
   }
+
+  let boundaryGeojson;
```
dabreegster, line 47: I lifted this out of `BoundaryLayer`, so we can also pass it down to `ZoomOutMap`. Right now there's not very much global app state plumbed around here; we will definitely need to think about organizing once it grows more.

components/DrawControls.svelte
```diff
@@ -62,8 +62,6 @@
   let routeSnapper;
 
   onMount(async () => {
-    const map = getMap();
```
dabreegster, line 65: All the places previously grabbing map from context can now just import from `stores.js` and subscribe as usual. The `$map` syntax is simplest, since that auto-unsubscribes upon component teardown (https://svelte.dev/tutorial/auto-subscriptions)

components/ZoomOutMap.svelte
```diff
@@ -0,0 +1,19 @@
...
+  function recenter() {
+    $map.fitBounds(geojsonExtent(boundaryGeojson), {
+      padding: 20,
+      animate: true,
+      duration: 500,
+    });
+  }
+</script>
+
+<button type="button" title="Zoom to show entire boundary" on:click={recenter}>
```
dabreegster, line 17: I tried a few variations of https://carbon-components-svelte.onrender.com/components/Button, but very obnoxiously they can't really take an SVG file as input; they need it wrapped as a svelte component. The styling also didn't necessarily seem awesome.

App.svelte
```diff
@@ -48,14 +67,14 @@
     <button type="button" on:click={toggleInstructions}>Instructions</button>
   </div>
   <div slot="sidebar">
-    <h1>{authorityName}</h1>
+    <h1>{authorityName} <ZoomOutMap {boundaryGeojson} /></h1>
```
dabreegster, line 70: The motivation behind switching from context to store for the MapLibre map is here. Context is based on the hierarchy of components -- it's only available to children of the place that sets it. So previously, only the 4 things underneath `Map` (`BoundaryLayer`, `DrawControls`, etc) could get the map. But... we have a button in the sidebar that needs to change the viewport.

I've been thinking through this problem earlier, and wrote up options at https://github.com/dabreegster/osm2svelte#map-and-sidebar-interactions. Option 3 is what we've followed in this codebase so far. We have stores like `currentSidebarHover` and `currentMapHover` that're global to the whole app, and sometimes stuff in the sidebar (not in the same part of the component tree as the Map) will change things here. But that feels like total overkill here -- I don't want to insert a little listener component somewhere under `Map` just to handle this button click. That'd be putting code to handle one piece of behavior in two different places.

I've also been looking through https://github.com/dimfeld/svelte-maplibre for inspiration how to handle some harder problems around only working with fully-loaded maps, which is a bit relevant here. They're using contexts, but also don't really have any examples with controls that live "outside" the `Map` subtree

robinlovelace-ate (approved):
Review is based purely on video, not tested locally and have not read the Svelte code. One minor comment: the button has little/no padding meaning it butts up to the text in the video no? My only comment is that a bit of padding, e.g. 2 pixels, could slightly improve aesthetics. But that is tiny compared with the huge benefit of being able to zoom to your authority after each scheme/navigation.

dabreegster:
Screenshot from 2023-03-20 17-52-24
Does this look a little better?

The grey background style for the button is probably not what we want either. I'm punting a little bit on these kind of style decisions until we have a plan in #90, but we can still adjust stuff immediately

robinlovelace-ate:
Looks fine to me, I see there was a small bit of padding round the original version also, not sure what's different compared with this from the great vid above. 

image

dabreegster:
No rush @Sparrow0hawk, just checking if you have an opinion either on the new feature or the store/context question. The latter is something we can hash out more later, it's very much an ongoing/open issue

Sparrow0hawk (approved):
Tested this and works nicely.

I like the way the logic works here, it definitely makes sense the way we're dealing with the map to have it globally available within the store.

Issue #59, opened 2023-01-05
Get MapTiler API key


dabreegster:
Also, figure out how we're supposed to plumb it around. I can get the key out of Github just fine, but ultimately it winds up in the JS served to the user. Someone could trivially sniff the HTTP request in their browser or check the source. What's common practice here?

Robinlovelace:
Not sure... Any ideas @acalcutt or others on the Maptiler team? Looking at https://github.com/maptiler/tileserver-gl

acalcutt:
I'm only loosely affiliated with Maptiler because I requested access to update tileserver to maplibre and help the community move along some tileserver PR.

 I can't really speak on the API key since I use self-hosted mbtilese here. Typically I see the key specified on the source urls in the style, but they are viewable that way.
 
 Are you trying to using the open source Tileserver-GL or the closed source Maptiler Server?

acalcutt:
Maybe check out https://documentation.maptiler.com/hc/en-us/articles/360020806037-How-to-protect-your-map-key

dabreegster:
We're using our own key now. Leaving this issue open to:

- [ ] use a bit of config to deduplicate the two places in the code using the key
- [ ] Follow that article and set up domain whitelisting for usage of the key (but still let local `npm run dev` work?)
- [ ] Track the usage over a "typical" month and see how long we can manage on free tier

PR #93, opened 2023-03-22
Switch to our own MapTiler free account. #59

dabreegster:
I've been copying around the API key from MapLibre examples, but this morning on conference wifi, I got 403's for it. Seemingly not an issue off of conference wifi, but this is a ticking time-bomb anyway. In the short term, signing up for a free account was super easy, and the request quota should be more than enough. We can keep considering various other options for hosting long-term, but this moves us forward a bit.

As an aside, MapTiler has a v2 of the streets style. I kept the same one so far, but here's a comparison. Current:
Screenshot from 2023-03-22 11-34-10
v2:
Screenshot from 2023-03-22 11-33-50

### DIFF
--- a/components/Map.svelte
+++ b/components/Map.svelte
@@ -16,7 +16,7 @@
     map = new Map({
       container: mapContainer,
       style:
-        "https://api.maptiler.com/maps/streets/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL",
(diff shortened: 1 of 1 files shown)


Robinlovelace:
What's with the different colours, e.g. blue, in the v2?

robinlovelace-ate (approved):

dabreegster:
The blue is their new style for universities, I guess? We can pick a better
style for us later, I've kept the same streets v1 for now

PR #94, opened 2023-03-23
Allow loading multiple files and browsing through them, to convenient…

dabreegster:
…ly explore submissions

I'm exploring submitted ATIPs to understand what people are mapping in practice. To make that less painful, this PR adds a way to load multiple files all at once, then browse through them within the ATIP UI. It's a draft PR because I'm not sure we should add this feature in for real; I haven't thought through what implications it has for a regular user flow. It's only intended right now to help us.

See Slack for a video demo

### DIFF
--- a/src/lib/EntireScheme.svelte
+++ b/src/lib/EntireScheme.svelte
@@ -11,6 +11,10 @@
     baseFilename += `_${schema}`;
   }
 
+  // Strings
+  let otherFiles = [];
+  let browseFileIdx = 0;
+
   // Set up local storage sync
   let loadLocal = window.localStorage.getItem(baseFilename);
   if (loadLocal) {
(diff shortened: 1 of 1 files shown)


components/EntireScheme.svelte
```diff
@@ -4,6 +4,10 @@
 
   export let authorityName;
 
+  // Strings
+  let otherFiles = [];
```
dabreegster, line 8: If we actually decide to merge this, I'd want to clean up the code in a few ways. Maybe store processed JSON instead of strings, stop special-casing the "load 1" vs "load many" cases

components/EntireScheme.svelte
```diff
@@ -68,17 +72,28 @@
...
-        window.alert(`Couldn't load scheme from a file: ${err}`);
-      }
-    };
-    reader.readAsText(e.target.files[0]);
+    otherFiles = [];
+    browseFileIdx = 0;
+
+    let first = true;
+    for (let file of e.target.files) {
+      let immutableFirst = first;
```
dabreegster, line 80: JS can be a tricky language. :\  In Rust, referring to `first` inside the closure we create would move it there. But in JS, it keeps referring to the outer copy, which we change throughout the loop. Hence this copy

dabreegster:
Closing this old PR. Some of the ideas partly live on in the browse scheme page.

Issue #51, opened 2022-12-11
Switching baselayers breaks mapbox-gl-draw and the route tool


dabreegster:
`this.map.getSource(...) is undefined`
`setStyle` appears to blow up a bunch of sources. Disabling satellite for now

dabreegster:
Fixed by #351

PR #96, opened 2023-03-23
Add a control to switch baselayers to satellite. #51

dabreegster:
https://user-images.githubusercontent.com/1664407/227251357-2262a920-539a-4ab1-9102-225bd95a2bd4.mp4

See code comments for discussion points

### DIFF
--- a/components/BaselayerSwitcher.svelte
+++ b/components/BaselayerSwitcher.svelte
@@ -1,0 +1,30 @@
+<script>
+  export let style;
+
+  function changeStyle() {
+    let params = new URLSearchParams(window.location.search);
+    params.set("style", style);
+    let href = `${window.location.pathname}?${params.toString()}${
+      window.location.hash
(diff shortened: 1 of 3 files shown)


components/BaselayerSwitcher.svelte
```diff
@@ -0,0 +1,30 @@
+<script>
+  export let style;
+
+  function changeStyle() {
```
dabreegster, line 4: When I've tried using `setStyle` before to do this without replacing the entire page, I've hit problems. It removes all sources and layers from the map, because maplibre doesn't distinguish the "basemap" from all the things we add on top. https://stackoverflow.com/a/36169495 has a workaround -- when `map.on('style.load'` happens, setup all the sources and layers. There's not an easy way for us to do that yet. So for now, get things working by just jumping to a new page, stashing the style in the URL. Clunky, but working

PR #97, opened 2023-03-26
Color the different features by their type, and add a legend

dabreegster:
Before:
Screenshot from 2023-03-26 16-18-37
After:
Screenshot from 2023-03-26 16-18-52

Hovering over and editing stuff has also changed:

https://user-images.githubusercontent.com/1664407/227785693-ecf46d54-f8a7-4093-a1e8-7c704b92f8bb.mp4

The intention is to start symbolizing (through color, dash style, little icons, etc) different types of interventions. As we flesh out the schema and have 5 types of crossings, we may want to visually distinguish them upfront. This PR paves the way for doing that.


colors.js
```diff
@@ -0,0 +1,10 @@
+export const colors = {
+  // https://colorbrewer2.org/#type=qualitative&scheme=Set1&n=4
```
dabreegster, line 2: These specific choices are probably not great. Blue routes look like rivers? Very happy for design guidance here. It may not be too important to get things exactly right for the 4 categories today, if we'll want to distinguish more stuff in the schema soon.

Also, all of these choices depend very much on the basemap. If we switched to a greyscale style, or satellite, all of this will need to change.

components/DrawControls.svelte
```diff
@@ -22,35 +23,51 @@
...
+      filter: ["all", isPolygon, ["==", "meta", "feature"]],
+      ...drawPolygon(colorByInterventionType, polygonOpacity),
+      // TODO Outline too?
     },
     {
-      id: "base-polygon-fill",
-      filter: isPolygon,
-      ...drawPolygon(color, 0.1),
+      id: "draggable-points",
+      filter: ["all", isPoint, ["!=", "meta", "feature"]],
```
dabreegster, line 64: See https://github.com/mapbox/mapbox-gl-draw/blob/main/docs/API.md#styling-draw for some reference. There's a few weird things happening for us right now. If you draw a new polygon, you can see some of them (the fill color being the fallback black, and intermediate points vanishing). I still need to understand all of the combinations of things mapbox-gl-draw can create.

components/DrawControls.svelte
```diff
@@ -64,6 +81,7 @@
   onMount(async () => {
     drawControls = new MapboxDraw({
       displayControlsDefault: false,
+      userProperties: true,
```
dabreegster, line 84: This was subtle; we don't get `intervention_type` or other props without this! And `user_` gets prefixed

components/EditingLayer.svelte
```diff
@@ -23,23 +23,14 @@
...
-      filter: isPolygon,
-      ...drawPolygon(color, 0.5),
-    });
-    $map.addLayer({
-      id: "editing-lines",
-      source: source,
-      filter: isLine,
-      ...drawLine(color, 1.5 * lineWidth),
-    });
+    // Don't symbolize polygons or linestrings being edited. They'll have
```
dabreegster, line 26: I simplified this; it feels nicer to me now, but opinions welcome, I am far from a designer

components/HoverLayer.svelte
```diff
@@ -36,20 +36,24 @@
...
+      // covered by the 50% opaque color
+      ...drawLine(colors.hovering, lineWidth, 1.0),
     });
     $map.addLayer({
       id: "hover-lines",
       source: source,
       filter: isLine,
-      // TODO I'd like to cover up the base layers, but I can't figure out how to z-order on top of drawControls.
-      ...drawLine(hoverColor, 1.5 * lineWidth, 1.0),
+      // By "accident", this layer is underneath the draw controls. Draw a
```
dabreegster, line 48: It's very tricky to z-order stuff from different Sources in maplibre. When you `addLayer`, you can tell it an ID of another layer to place under. But `mapbox-gl-draw` doesn't plumb through a way to do this, and I think the other layer has to exist first. There's also `moveLayer` for correcting things later. But figuring out when both layers are "ready" is tough, because it depends when the component gets mounted by Svelte and some asyncish thing finishes...

colors.js
dabreegster: Oh yeah, and the route snapper overloads red, blue, and black to mean something else. Conflicts with things here.

Robinlovelace:
:eyes:

colors.js
Robinlovelace: > These specific choices are probably not great. Blue routes look like rivers? Very happy for design guidance here. It may not be too important to get things exactly right for the 4 categories today, if we'll want to distinguish more stuff in the schema soon.

I don't have any better colour suggestions at the moment...

Robinlovelace:
> The intention is to start symbolizing (through color, dash style, little icons, etc) different types of interventions. As we flesh out the schema and have 5 types of crossings, we may want to visually distinguish them upfront. This PR paves the way for doing that.

Big +1 to that, looks great. A step towards more sophisticated differentiation between intervention types.

App.svelte
```diff
@@ -78,6 +79,7 @@
       <DrawControls url={snapperUrl} />
       <HoverLayer />
       <EditingLayer />
+      <Legend />
```
Robinlovelace, line 82: Was this missing previously? Was expecting a matching 

```
<Legend > 
```

Somewhere but maybe that's not how Svelte works...

colors.js
Robinlovelace, line 2: At least the colours are not totally arbitrary.

Robinlovelace (approved):
Great iteration. Can always revisit. Good to keep it moving forward towards v2, Alpha, or whatever the next release is.

App.svelte
dabreegster: `Legend` is a totally new component, so just added here. `<x></x>` is equivalent to `<x />` in HTML generally, that's what's going on here

Robinlovelace: Makes sense, `<Legend />` = `<Legend></Lengend>`, in HTML not Svelte? Learn something every day!

PR #98, opened 2023-03-26
Only zoom to an intervention when clicking it in the sidebar, not fro…

### LINKED ISSUE #64: Don't zoom in when clicking something on the map
When you click an intervention from the left sidebar, we pan+zoom to focus on it -- that seems useful. When you click an object on the map or finish drawing something, we do the same thing, and it feels distracting. `EditingLayer.svelte` does this behavior by listening to the `currentlyEditing` store. Can we fix this without making the derived store super complicated? (And is a derived store even the best way to plumb this state change around?)

dabreegster:
…m the map. Closes #64

Before:

https://user-images.githubusercontent.com/1664407/227789897-8d555438-6540-4b42-8e38-c7dd01a16fee.mp4

After:

https://user-images.githubusercontent.com/1664407/227789906-55965c38-1da7-4a3a-b9f2-af783e851f88.mp4

I've found it very annoying (personally -- this is a design choice we could poll users about if we needed to?) to reset the camera when clicking an object on the map. So this PR makes us only zoom to something if clicked in the sidebar

### DIFF
--- a/components/DrawControls.svelte
+++ b/components/DrawControls.svelte
@@ -18,7 +18,7 @@
     currentMapHover,
     setCurrentlyEditing,
     clearCurrentlyEditing,
-    currentlyEditing,
+    openFromSidebar,
     map,
   } from "../stores.js";
 
@@ -143,7 +143,7 @@
       currentMapHover.set(null);
     });
 
-    currentlyEditing.subscribe((id) => {
+    openFromSidebar.subscribe((id) => {
       if (id) {
         let feature = $gjScheme.features.find((f) => f.id == id);
         // Act like we've selected the object. (This is irrelevant for points.)
--- a/stores.js
+++ b/stores.js
@@ -7,6 +7,8 @@ export const gjScheme = writable(emptyGeojson());
 export const currentSidebarHover = writable(null);
 export const currentMapHover = writable(null);
 
+export const openFromSidebar = writable(null);
+
 // TODO Should we store a map from ID to feature?
 // TODO DrawControls will partly own state. Do we have to listen for every geometry change?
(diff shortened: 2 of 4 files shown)


components/DrawControls.svelte
```diff
@@ -143,7 +143,7 @@
       currentMapHover.set(null);
     });
 
-    currentlyEditing.subscribe((id) => {
+    openFromSidebar.subscribe((id) => {
       if (id) {
         let feature = $gjScheme.features.find((f) => f.id == id);
         // Act like we've selected the object. (This is irrelevant for points.)
```
dabreegster, line 149: Actually we have a pre-existing bug that's still here. If you edit a point from the sidebar, you can't drag the point as you should be able to. Not sure why yet

stores.js
```diff
@@ -7,6 +7,8 @@ export const gjScheme = writable(emptyGeojson());
 export const currentSidebarHover = writable(null);
 export const currentMapHover = writable(null);
 
+export const openFromSidebar = writable(null);
```
dabreegster, line 10: I thought about using event dispatchers, but the flow would be too complicated. InterventionList up to App, but then down to DrawControls and EditingLayer. Doesn't seem to be a nice way of doing that, so we can just use stores.

Robinlovelace:
Nice feature! Taking a look...

Robinlovelace:
Definite UX win IMO. Thumbs up!

stores.js
Robinlovelace: I would be lying if I said I knew what that means. Trust your judgement here!

Robinlovelace (approved):

PR #99, opened 2023-03-27
Experiment with a large modal form, partly generated from a JSON schema

dabreegster:
This isn't intended to merge, just a quick way to experiment with using JSON schemas from https://github.com/acteng/schema/ in ATIP:

https://user-images.githubusercontent.com/1664407/227908327-530f23fe-4f25-4eb5-8bfc-aac7891da765.mp4

# Notes on libraries

I tried a few NPM packages for autogenerating a form from a JSON schema, but didn't have any luck finding one that works well out-of-box. I don't think it's too important to invest in this, because

1) the JSON schema format is just a convenient way for us to iterate on the schema right now; it may not be the final thing we choose to use in the frontend, for any DB schema, etc
2) in ATIP, we will probably always want custom form / controls. As you select a crossing type, we could show some example images inline, have a cost estimate, link to design manuals, etc. Maybe once you specify the "The main source of new space for active travel" for routes, you get a new sub-form with something Streetmix-like, etc.

But anyway, here's what I tried for reference:

- https://github.com/xpluscal/svelte-schemaform: somehow breaks CSS button styling, can't handle empty input to start
- https://github.com/restspace/svelte-schema-form: Requires Saas dependency and the installation somehow breaks
- https://github.com/pyoner/svelte-form: Old, installation breaks

### DIFF
--- a/components/Form.svelte
+++ b/components/Form.svelte
@@ -1,21 +1,33 @@
 <script>
   import {
+    Dropdown,
+    Modal,
     RadioButtonGroup,
     RadioButton,
     TextArea,
     TextInput,
+    NumberInput,
   } from "carbon-components-svelte";
-  import {
-    gjScheme,
-    clearCurrentlyEditing,
-    currentlyEditing,
-  } from "../stores.js";
+  import { gjScheme, clearCurrentlyEditing, currentlyEditing } from "../stores.js";
+  import areaSchema from "../schema/areas.json";
+  import crossingSchema from "../schema/crossings.json";
+  import otherSchema from "../schema/other.json";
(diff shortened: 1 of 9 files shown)


components/Form.svelte
```diff
@@ -50,9 +66,39 @@
...
+
+  {#if intervention_type == "route"}
+    <NumberInput
+      step={0.1}
+      label={routeSchema.properties.width.description}
+      bind:value={details.width}
+    />
+  {/if}
+
+  <!-- TODO Hack for the Modal -->
```
dabreegster, line 102: The Carbon Modal is not smart about dropdowns that may expand, or has any way I can find to specify a height without something like this

PR #100, opened 2023-03-27
Less clunky developer experience, part 1

dabreegster:
Vite does hot-module reloading (HMR), meaning when you edit code, just the necessary Svelte components get reloaded without the whole page refreshing. It's incredibly handy to tune CSS or map drawing styles things and instantly see feedback. But today, if you edit the code in many ATIP components, the page breaks completely.

This PR fixes most of the cases involving sources and layers on the map. MapLibre doesn't let you add a source or layer with a duplicate ID. We could solve this in a few ways:

- Writing an `onDestroy` for all relevant components, and carefully removing things in the correct order. (Layers have to be removed before their source). Boilerplatey, and really easy to get wrong.
- Using Svelte components that wrap Sources and Layers and do the lifetime management properly. This is the approach in https://github.com/dimfeld/svelte-maplibre/. I like it, but it's a bigger change for us to either switch to this library or to mimic this style.
- Use helper methods that first remove sources/layers before adding them, if needed. This is the approach in this PR. Maybe not ideal, but simple.

In the next PR, I want to fix more remaining problems. Everytime we do `someStore.subscribe`, that component can't be reloaded properly; the old callback dangles and so logic will run a bunch of times, or we hold onto stale objects. Svelte auto-subscriptions will help here, but it's not super easy to get right, so I'll save it for the next step.

Going to ASAP merge this because it dramatically makes my editing workflow better, but very happy to address feedback later

### DIFF
--- a/components/BoundaryLayer.svelte
+++ b/components/BoundaryLayer.svelte
@@ -1,30 +1,32 @@
 <script>
   import geojsonExtent from "@mapbox/geojson-extent";
   import mask from "@turf/mask";
-  import { drawPolygon } from "../maplibre_helpers.js";
-  import { onMount, getContext } from "svelte";
+  import {
+    drawPolygon,
+    overwriteSource,
+    overwriteLayer,
+  } from "../maplibre_helpers.js";
+  import { getContext } from "svelte";
   import { map } from "../stores.js";
 
   export let boundaryGeojson;
 
   const setCamera = getContext("setCamera");
 
-  onMount(() => {
-    if (setCamera) {
-      $map.fitBounds(geojsonExtent(boundaryGeojson), {
-        padding: 20,
-        animate: false,
-      });
-    }
+  if (setCamera) {
+    $map.fitBounds(geojsonExtent(boundaryGeojson), {
+      padding: 20,
+      animate: false,
+    });
+  }
 
-    $map.addSource("boundary", {
-      type: "geojson",
-      data: mask(boundaryGeojson),
-    });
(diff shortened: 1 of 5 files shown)


components/BoundaryLayer.svelte
```diff
@@ -1,30 +1,32 @@
...
+    overwriteLayer,
+  } from "../maplibre_helpers.js";
+  import { getContext } from "svelte";
   import { map } from "../stores.js";
 
   export let boundaryGeojson;
 
   const setCamera = getContext("setCamera");
 
-  onMount(() => {
```
dabreegster, line 12: Also, the `onMount` stuff isn't helpful in a few of these components. There's nothing we're sticking in the DOM; we're just interacting with the map

maplibre_helpers.js
```diff
@@ -0,0 +1,72 @@
...
+
+// MapLibre's API isn't idempotent; you can't overwrite an existing source or
+// layer. This complicates Vite's hot-reload feature, unless every component
+// correctly tears down all sources and layers. These methods workaround that
+// lifetime management hassle by overwriting if necessary.
+export function overwriteSource(map, id, source) {
+  if (map.getSource(id)) {
+    // First remove all layers using this source
+    let layers = [];
+    for (let layer of map.getStyle().layers) {
```
dabreegster, line 53: Being paranoid here and not removing layers while we loop over the available layers, because I have no idea how JS works and `removeLayer` is internally implemented

PR #101, opened 2023-03-28
When we expand a form, scroll to ensure it's visible in the sidebar

dabreegster:
This is a small tweak that makes navigation/editing easier. If someone picks an intervention, they probably want to actually see the entire form.

Before:

https://user-images.githubusercontent.com/1664407/228212695-c4638fc0-3b63-486a-b7ee-b6450bfce2c4.mp4

After:

https://user-images.githubusercontent.com/1664407/228212718-85059976-8b7d-42b2-9232-24884957f3e4.mp4

### DIFF
--- a/components/Form.svelte
+++ b/components/Form.svelte
@@ -5,13 +5,28 @@
     TextArea,
     TextInput,
   } from "carbon-components-svelte";
-  import { gjScheme, clearCurrentlyEditing } from "../stores.js";
+  import {
+    gjScheme,
+    clearCurrentlyEditing,
+    currentlyEditing,
+  } from "../stores.js";
 
   export let id;
   export let name;
   export let intervention_type;
   export let description;
   export let length_meters;
+
+  let bottomOfForm;
+
+  currentlyEditing.subscribe((openedId) => {
+    if (id == openedId) {
+      // Wait 1ms before doing this, because it appears the accordion doesn't
+      // expand instantly. This is flaky when clicking the accordion instead of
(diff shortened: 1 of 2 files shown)


components/Form.svelte
```diff
@@ -5,14 +5,29 @@
...
   export let name;
   export let intervention_type;
   export let description;
   export let length_meters;
 
+  let bottomOfForm;
+
+  currentlyEditing.subscribe((openedId) => {
+    if (id == openedId) {
+      // Wait 1ms before doing this, because it appears the accordion doesn't
```
dabreegster, line 24: I can't figure out how to listen to events for when the accordion finishes expanding

components/Form.svelte
```diff
@@ -5,14 +5,29 @@
...
+      // expand instantly. This is flaky when clicking the accordion instead of
+      // the map.
+      setTimeout(() => bottomOfForm?.scrollIntoView({ behavior: "smooth" }), 1);
```
dabreegster, line 27: The `?` is https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining. There could be some crazy race condition where the form vanishes by the time 1ms has passed, so don't blow up if it becomes null

robinlovelace-ate:
UX win: focusses helpfully on the fields the user is most likely to want to edit, reduces risk of people forgetting to add description etc.

robinlovelace-ate (approved):

PR #103, opened 2023-03-28
Show a tooltip when hovering on map features

dabreegster:
https://user-images.githubusercontent.com/1664407/228236482-b7826fcc-3eaa-4c78-a87d-a58076d80d95.mp4

Just experimenting and looking for early UX feedback/ideas. I find it hard to know what stuff on the map is, and tooltips could maybe help. See comments in code for some problems


dabreegster:
I'm going to abandon this PR. The code is very outdated, and it's still a UX question if we want to do this kind of thing generally or not.

robinlovelace-ate:
Sensible decision by the sounds, an avenue worth exploring.

Issue #58, opened 2023-01-05
Rewrite the map control buttons from scratch


dabreegster:
mapbox-gl-draw's built-in buttons for toggling tools don't work in a few ways:

- hard to customize the style
- hard to insert the header above, or the route tool below the group
- the tooltips aren't customizable
- the tooltips lie about keybindings
- not visually obvious when you've selected a tool
- no great way to display popup instructions to help people with the polygon tool, like we do for route snapper

Let's just rip them out and make our own nicer Svelte ones. Possible inspiration is https://terradraw.io, putting them at the top and maybe adding instructions beneath?

dabreegster:
I'm trying to integrate new route-snapper changes and having lots of trouble. mapbox-gl-draw tells us when an object has been clicked by `draw.selectionchange`, but this event is fired for a bunch of other reasons. I can't find a way of getting at the simpler case. We're starting to have complicated sequences... someone clicks a route, we need to tell mapbox-gl-draw to not edit the line-string as usual and instead go to a static mode, and launch the route snapper tool instead. But then changing to static mode deselects the object, and we get another event, and... it's a mess.

We're not using very much of the plugin at this point, and many aspects of the way it works are not ideal. I'm going to prototype something from scratch. We just need to drop a point, move a point, and do the same for polygons. I've also had a look at https://github.com/JamesLMilner/terra-draw, but I'm less confident about how to start building on top of it.

dabreegster:
Thinking through desired UX...

- On the map, the toolbox has a few modes, one of which is active
  - Select: if you click something on the map, you'll open the form for it
  - Edit: if you click something on the map, you'll edit its geometry. For routes and crossings, it'll reopen the route tool. For polygons it'll let you move the whole polygon or vertices. For points, it'll let you drag the point.
  - New point: when you click the map, it'll add a new point
  - New polygon: when you click the map, it'll start adding vertices for a polygon
- The instructions can be shown dynamically
  - Today it's unclear how to finish drawing a polygon. When we have a mode active, we can show the keys/mouse interactions that're valid next steps to do
- This could pave the way for drawing more specific things upfront. Instead of "new point", we could have an expandable menu of point-like things to draw -- modal filters, freeform notes, a spot-fix, etc.

dabreegster:
It's gone!

PR #108, opened 2023-04-02
Overhaul the map drawing UI

dabreegster:
Try out this PR: https://acteng.github.io/atip/draw_v3/scheme.html?authority=West%20of%20England%20Combined%20Authority#9.86/51.476/-2.513

Old/current version: https://acteng.github.io/atip/scheme.html?authority=West%20of%20England%20Combined%20Authority#9.86/51.476/-2.513

----

Towards #58. A bunch of big changes. It's easiest to just run locally and try it out, but here's a video:

https://user-images.githubusercontent.com/1664407/229365933-8c70f1f6-19d3-4ce3-a914-278506fad8b2.mp4

- Distinguish "select" and "edit" mode. The first is about editing the form/properties of an object, and the second is about editing geometry. These need better names! The motivation is that the form will likely soon move from the sidebar to a modal.
- Replace the on-screen toolbox with our own custom thing. This gives us room to put in contextual instructions.
- Get the new route editing functionality working. You can now edit the waypoints of an existing route.

This needs much more work before merging, but I'm at the point where I'd love feedback. Does this feel like the right way forward? What should change? And a specific question: what should we allow doing on the sidebar while in anything other than "Select" mode? While we're editing geometry or drawing a new area, it feels confusing to me to click something on the sidebar. Ideally that would "cancel" any of the geometry editing state and jump back to "select" mode. Or maybe it's more clear to disable and grey out all the controls over there?


Sparrow0hawk:
Having a quick play locally and run across a situation causing a crash:

If you select New Point, but then immediately click New Polygon and immediately click back to New Point (without drawing) it crashes with the following error:
```error
Uncaught (in promise) Error: Source "mapbox-gl-draw-cold" already exists.
```
https://user-images.githubusercontent.com/20887250/229373432-37d5f81e-0078-4eda-a842-9dcaccc9bfa6.mov

Actually, just clicking from New Polygon and immediately clicking back to New Point causes this crash.

Sparrow0hawk:
I like the changes here a lot, I spent time mostly just making routes, adding points and making polygons but will try and have a play with select and edit more too.

My only initial feedback would be is it possible wrap the text when a button in the toolbox is active to avoid hiding more of the map. 

Before New Polygon active:
Screenshot 2023-04-02 at 20 16 33
After New Polygon active:
Screenshot 2023-04-02 at 20 17 08

components/draw/Toolbox.svelte
```diff
@@ -0,0 +1,96 @@
...
+    background-color: #4caf50;
+    border: none;
+    color: white;
+    padding: 15px 32px;
+    text-align: center;
+    text-decoration: none;
+    display: inline-block;
+    font-size: 16px;
+    margin: 4px 2px;
+    cursor: pointer;
```
Sparrow0hawk, line 82: Re my comment about text wrapping I tried adding 
`overflow-wrap: break-word;` here but this doesn't work. 
I think because that requires an absolute width to cause an overflow which isn't what we see here.

dabreegster:
> Actually, just clicking from New Polygon and immediately clicking back to New Point causes this crash.

Thanks for finding the bug! There are two `PointOrPolygonMode` components, both managing a mapbox-gl-draw instance. When we change mode, I'm guessing we set up the new one before tearing down the old one. I'll think about how to fix that, either by forcing something with the ordering... or maybe just having one copy of mapbox-gl-draw somewhere, instead of constantly creating and destroying. Having similar implementation debates for implementing "edit point/polygon."

components/draw/Toolbox.svelte
dabreegster: I think it'd be the `div` above, not the `button`. I'll play around tomorrow. The styling in the toolbox is pretty placeholder right now. I also want to overhaul the route tool, so it doesn't look so wildly different!

dabreegster:
> I'll think about how to fix that, either by forcing something with the ordering... or maybe just having one copy of mapbox-gl-draw somewhere

I went with the latter, because reasoning about ordering is hard in general, and also https://stackoverflow.com/questions/64514581/what-is-the-order-that-reactive-statements-are-executed-in. The bug you found is fixed now, but still lots more fixes to go before this is ready. I'll ping when a review would be helpful.

Though one question in the meantime: if you're in the middle of drawing a route or editing a polygon's geometry, the interactions with the left sidebar are confusing. What should they do? Should they be disabled while editing geometry, or should clicking them just silently cancel the geometry edit and switch back to selection?

components/EntireScheme.svelte
```diff
@@ -82,12 +83,22 @@
...
   function backfill(json) {
-    // Look for any LineStrings without length_meters. Old route-snapper versions didn't fill this out.
+    let idCounter = 0;
     for (let f of json.features) {
+      // Look for any LineStrings without length_meters. Old route-snapper versions didn't fill this out.
       if (f.geometry.type == "LineString" && !f.properties.length_meters) {
         f.properties.length_meters =
           length(f.geometry, { units: "kilometers" }) * 1000.0;
       }
+      // TODO Hack around https://github.com/mapbox/mapbox-gl-js/issues/2716.
```
dabreegster, line 93: This is kind of a nightmare. I just want every feature to have a unique ID, but we can't do something simple like use a `uuid` library, because parts of maplibre **silently** break on non-numeric IDs. But using numeric IDs like this is also confusing:

- this backfill has a bug and might double-assign an ID, the way it's written right now
- `0` acts like `false` in JS silently, and I've been bitten by `if (id)` instead of `id (id != null)` so many times
- mapbox-gl-draw lets us use our own feature IDs... unless it's `0`!

components/draw/AttributeMode.svelte
```diff
@@ -0,0 +1,91 @@
...
+
+  export let mode;
+  export let drawControls;
+  // Maybe null at first
+  export let routeSnapper;
+
+  $: {
+    if (mode == "edit-attribute") {
+      drawControls.changeMode("static");
+      // TODO The stop() implementation should be robust to the source not being ready yet
```
dabreegster, line 20: General frustration: I can't currently wire up local ATIP to local route snapper edits. So small problems like this require me to fix the problem elsewhere and publish a new NPM version. Still trying to find a better workflow.

components/draw/AttributeMode.svelte
```diff
@@ -0,0 +1,91 @@
...
+  } from "../../stores.js";
+
+  export let mode;
+  export let drawControls;
+  // Maybe null at first
+  export let routeSnapper;
+
+  $: {
+    if (mode == "edit-attribute") {
+      drawControls.changeMode("static");
```
dabreegster, line 19: I'm starting to dislike these reactive statements when `mode` changes, but I'm not sure what else to do. All of the different modes effectively need to do something when they become active, and clean up when they become inactive. There are many ways they can become inactive -- for example, click the route tool to start it, but then click something in the sidebar or change to "Edit attributes" mode. The problem with reactive statements is that there's no ordering guarantees at all, so if multiple things are listening to `mode` change, we don't really know what order stuff is happening. Any ideas for detangling this mess?

components/draw/AttributeMode.svelte
```diff
@@ -0,0 +1,91 @@
...
+      });
+      if (results.length > 0) {
+        setCurrentlyEditing(results[0].id);
+      } else {
+        setCurrentlyEditing(null);
+      }
+    }
+  });
+
+  // When the user starts editing something from the sidebar, warp to what's
```
dabreegster, line 69: I moved this from the old `EditingLayer`

components/draw/AttributeMode.svelte
```diff
@@ -0,0 +1,91 @@
+<script>
```
dabreegster, line 1: This was called "Select mode" yesterday, for context

components/draw/GeometryMode.svelte
```diff
@@ -0,0 +1,156 @@
+<script>
```
dabreegster, line 1: This was "Edit mode" yesterday. "Edit attributes" and "edit geometry" sound OK, or any other ideas?

components/draw/GeometryMode.svelte
```diff
@@ -0,0 +1,156 @@
...
+        return gj;
+      });
+
+      // Stay in this mode
+      currentlyEditing = null;
+    }
+  });
+
+  function startEditing(id) {
+    currentHover.set(null);
```
dabreegster, line 107: Maybe it's confusing to abuse the hover layer here. It makes entries in the sidebar light up too. (But... that could be nice; people need to know the name of the thing they're about to edit.) But... then interacting with sidebar entries while in this mode gets confusing.

components/draw/GeometryMode.svelte
```diff
@@ -0,0 +1,156 @@
...
+        ];
+      }
+      routeSnapper.editExisting(feature);
+    } else if (feature.geometry.type == "Polygon") {
+      // The feature already has an ID set, so no need to check what add() returns
+      drawControls.add(feature);
+      drawControls.changeMode("direct_select", { featureId: currentlyEditing });
+    } else if (feature.geometry.type == "Point") {
+      drawControls.add(feature);
+      // We can't direct_select a point
```
dabreegster, line 147: This is clunky. "Edit geometry" mode, then click an existing point (with the intention of moving it). You then have to click (again) and drag it

components/draw/PointOrPolygonMode.svelte
```diff
@@ -0,0 +1,55 @@
...
+  // "point" or "polygon"
+  export let type;
+  export let drawControls;
+  // Maybe null at first
+  export let routeSnapper;
+
+  $: {
+    if (mode == type) {
+      drawControls.changeMode(`draw_${type}`);
+      routeSnapper?.stop();
```
dabreegster, line 19: Here's another kind of weird state-changing example. Anytime we enter a mode, we have to make sure to turn off stuff from every other mode. This feels very "leaky", not encapsulated at all.

dabreegster (commented):
I'm trying to get this PR to MVP so it stops growing in scope, but it's hard. The code complexity is growing again; I wanted to tightly reason about being in exactly one "mode" at a time, but implementation details are leaking everywhere, the transitions between modes happen in odd ways, and the transitions can also be triggered from the sidebar.

Help, please? In figuring out how to structure the code sanely, and/or deciding what the UX should be for sidebar interactions while in anything other than "Edit attributes" mode

dabreegster:
OK, I feel like I've dug out of the hole now and addressed some of the big problems:
- changing between modes is now much more clear in the code, and there are only some smallish bugs
- if we're willing to have feature IDs jump around on save/load, no longer any bugs there

The last big hurdle before I'd call this ready for real review is figuring out what to do about sidebar interactions

dabreegster:
As discussed, for now, if you click something in the sidebar, we jump to editing the form of that thing. The only major clunkiness is that the hover state (both in the map and sidebar) gets kind of stuck based on what's hovered in the sidebar, depending on the mode you're in:

https://user-images.githubusercontent.com/1664407/230404152-cc2214d9-7ed3-4729-a1b6-5c0238974392.mp4

I'm not sure how to improve on this right now, so vote for leaving it as-is. I'll file issues and follow-up with more tightly scoped next steps. In particular:

- Stop using Carbon's accordion. The entire reason we need `properties.editing` and lots of confusion is due to how the library communicate state changes with bound vaariables. And we can't do things like change the style to show hovering on a map object.
- Write our own point and polygon drawing tools. Neither is hard with straight Svelte, and getting rid of mapbox-gl-draw is high on my agenda.
- Take over the controls/button for the route tool, to unify the style
- Figure out how to write tests, so we could automate the quadratic explosion of transitions between modes

But for now, please comment with any show-stopper changes you'd like to see before merging this PR

components/InterventionList.svelte
```diff
@@ -68,7 +72,7 @@
 <svelte:window on:keydown={onKeydown} />
 
 <Accordion>
-  {#each $gjScheme.features as feature, i}
+  {#each $gjScheme.features as feature, i (feature.id)}
```
dabreegster, line 75: https://svelte.dev/tutorial/keyed-each-blocks
I can't tell if this was causing any actual problems earlier or not, but seems good to do this?

Sparrow0hawk:
I can't see the GitHub action adjustments you made to deploy just this branch? I'm sure I'd definitely seen them previously but am now doubting myself slightly (although you definitely did it somewhere because we've still got a version of draw_v3 deployed)

dabreegster:
https://github.com/acteng/atip/commit/a121258ba360dda9af3e30d43cecbdac86dace0a I made the commit to `main` outside of a PR so it took immediate effect, sorry to be confusing!

Sparrow0hawk:
> [a121258](https://github.com/acteng/atip/commit/a121258ba360dda9af3e30d43cecbdac86dace0a) I made the commit to `main` outside of a PR so it took immediate effect, sorry to be confusing!

Ahh gotcha! I think that action hasn't been running for more recent changes of this branch? In the actions log the only time I can see it running was when you pushed to main? I think this is because that updated action yml isn't present in this branch and I /think/ that for an action to run on a particular branch it requires that said branch contains the workflow.

dabreegster:
> I think this is because that updated action yml isn't present in this branch and I /think/ that for an action to run on a particular branch it requires that said branch contains the workflow.

Ahh that makes sense! And actually it means it would've worked to make the workflow change here. I wasn't sure how things worked. I just rebased this PR and force-pushed, so it should deploy properly now..

components/draw/AttributeMode.svelte
```diff
@@ -0,0 +1,91 @@
...
+        $map.fitBounds(geojsonExtent(feature), {
+          padding: 200,
+          animate: true,
+          duration: 500,
+        });
+      }
+
+      // Interrupt any other mode and switch to this one. No need to
+      // setCurrentlyEditing -- the caller already sets the editing property.
+      changeMode(thisMode);
```
Sparrow0hawk, line 84: This is defined up in Toolbox.svelte and piped down.

components/draw/GeometryMode.svelte
```diff
@@ -0,0 +1,169 @@
...
+        return gj;
+      });
+    }
+
+    currentlyEditing = null;
+    currentHover.set(null);
+  }
+
+  // This is different than the stores.js one!
+  let currentlyEditing = null;
```
Sparrow0hawk, line 32: Why do we use a local `currentlyEditing` here rather than the one from `stores.js`?

components/draw/GeometryMode.svelte
```diff
@@ -0,0 +1,169 @@
...
+        delete feature.properties.hide_while_editing;
+        return gj;
+      });
+
+      // Stay in this mode
+      currentlyEditing = null;
+    }
+  });
+
+  $map.on("draw.update", (e) => {
```
Sparrow0hawk, line 106: One slight behaviour different I find when using Edit Geometry mode is if I click a Polygon, edit it and try to finish editing by clicking Enter (which is the behaviour for creating a Polygon) we don't drop out of Edit mode, the only way to do that is by clicking away from the editing area (firing next event on line 121)

Sparrow0hawk: I'd note I don't think this is a blocking issue, just worth noting.

components/draw/GeometryMode.svelte
Sparrow0hawk: I think this is fine, although note the UX you're describing would be nice to have.

components/draw/GeometryMode.svelte
Sparrow0hawk: This is again annoying but don't think it's a blocker at this stage. The UX is annoying with that teeny-tiny little point it shows once selected but I think overriding the style to make the point larger is probably a good way to go.

components/draw/InterventionLayer.svelte
```diff
@@ -0,0 +1,67 @@
...
+    drawPolygon,
+    overwriteSource,
+    overwriteLayer,
+  } from "../../maplibre_helpers.js";
+  import { colors, circleRadius, lineWidth } from "../../colors.js";
+  import { gjScheme, map } from "../../stores.js";
+
+  let source = "interventions";
+
+  overwriteSource($map, source, {
```
Sparrow0hawk, line 17: My head is a little confused by this. 

We define a variable called `source` which we pass into `overwiteSource` here. `overwriteSource` takes 3 args: `map`, `id` and `source`, but we're passing local `source` as the `id` argument.

This is probably just unfamiliarity on my part with the maplibre API and how layers work...

components/draw/PointOrPolygonMode.svelte
```diff
@@ -0,0 +1,79 @@
...
+        // It'd be great to do the callback below right now, but mapbox-gl-draw
+        // will infinitely recurse and blow up if we mutate it inside
+        // draw.create.
+
+        return gj;
+      });
+    }
+  });
+
+  $map.on("draw.modechange", (e) => {
```
Sparrow0hawk, line 50: This handles behaviour immediately after a point is created or a Polygon is finished.

components/draw/Toolbox.svelte
```diff
@@ -0,0 +1,176 @@
...
+      >
+      <AttributeMode {mode} {drawControls} {routeSnapper} />
+    </div>
+    <div>
+      <button
+        type="button"
+        on:click={geometryMode}
+        disabled={mode == "edit-geometry"}>Edit geometry</button
+      >
+      {#if routeSnapper && snapTool}
```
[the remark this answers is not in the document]
Sparrow0hawk: Not sure either but maybe one to get feedback on before deciding.

I.e. do lots of people complaining about this blocking approach so it'd be nicer to handle it quietly

Sparrow0hawk (approved):
This PR looks good to me. I've tested it on a bunch of old files and note no issues I can find with backwards compatibility. The draw controls work nicely and previous issues seem resolved.

I think drawing a bit of a component tree might be worthwhile at this stage, just to help keep track of how this is stacking up. I may if I have time over the long weekend draft something up in Miro and link back here.

I've left comments throughout, some are annotations for me, some replies to you and a couple of questions where i'm not quite getting something.

But lets merge this! Great work @dabreegster 💪

components/draw/GeometryMode.svelte
dabreegster: I've got an in-progress followup to replace point and polygon drawing and improve this. For now, agree, it's not a blocker

components/InterventionList.svelte
dabreegster: Yes, it's looping over a list and enumerating the index. Spotty internet right now, but somewhere nearby in the Svelte tutorial it does that. What I added here was this "keyed each block" feature, which is supposed to help Svelte not rebuild the entire list when anything changes. (Or put things in the wrong order?) No behavior change that I could notice, but it still seems better

dabreegster:
Thanks for the review! I'll merge this and work on followup improvements, or file issues.

> I think drawing a bit of a component tree might be worthwhile at this stage, just to help keep track of how this is stacking up. I may if I have time over the long weekend draft something up in Miro and link back here.

Yes! I was looking at https://github.com/oslabs-beta/svelte-sight/ and similar for automating this. Understanding how props flow up and down the tree could be nice. Mostly we're using the stores as global singletons and not plumbing everything everywhere, though

components/draw/GeometryMode.svelte
dabreegster: I'll change the comment. We need better terminology; "editing" now means "editing geometry" or "editing properties". This is "editing geometry", the one in the store is for properties.

components/draw/GeometryMode.svelte
dabreegster: Agree this is annoying. I'll aim to address this when we replace polygon drawing.

components/draw/InterventionLayer.svelte
dabreegster: The variable `source` is meant as a source ID. It's helpful to name it this way so we can just say `source,` instead of `source: sourceId` when we call `overwriteLayer` later. Should we call the variable `sourceId` maybe to be more clear? Or we used TypeScript, it'd be extremely clear this is being used a string...

MapLibre sources and layers both have a string ID and a definition given by a dictionary.

Sparrow0hawk: Gotcha. That makes sense I think `sourceId` is clearer but totally get why you've called it `source`.

Sparrow0hawk:
I had quick peek at svelte-sight but didn't really have a play. I started drawing something up on [miro](https://miro.com/app/board/uXjVMVVjLdc=/#tpicker-content) although got side tracked and not sure of the best way to illustrate all the flows (i'm just trying to illustrate components and props here)

PR #114, opened 2023-04-11
Split routes

dabreegster:
https://user-images.githubusercontent.com/1664407/231227440-89f0468e-7b08-485a-b500-d372046ce3b6.mp4

Here's a new mode to split routes. The intended flow is that users will first sketch the overall route they want, then later optionally fill in more info about different sections of the route. The specific interventions will likely vary based on available width and other factors. This tool helps them refine a long route into shorter ones.

There's one big task left before this PR is ready: when we split a route, we need to update the route-snapper's stored waypoints, so that we can then edit each piece individually.

Some inspiration for other line splitting tools in the mapbox-gl-draw ecosystem: https://github.com/ReyhaneMasumi/mapbox-gl-draw-split-line-mode, https://github.com/BrunoSalerno/mapbox-gl-draw-cut-line-mode, https://github.com/mhsattarian/mapbox-gl-draw-snap-mode

### DIFF
--- a/components/draw/SplitRouteMode.svelte
+++ b/components/draw/SplitRouteMode.svelte
@@ -71,7 +71,10 @@
       return;
     }
 
-    if (snappedIndex != null) {
+    if (snappedIndex == null) {
+      // We clicked the map, stop the tool
+      changeMode("edit-attribute");
+    } else {
       let result = lineSplit($gjScheme.features[snappedIndex], cursor);
       if (result.features.length == 2) {
         let piece1 = result.features[0];
@@ -97,9 +100,10 @@
           return gj;
         });
       }
+
+      // Stay in this mode, but reset state
+      stop();
     }
-    // TODO Edit one of the pieces?
-    changeMode("edit-attribute");
   });
 
   // Rendering
--- a/components/draw/SplitRouteMode.svelte
+++ b/components/draw/SplitRouteMode.svelte
@@ -156,6 +156,8 @@
 
     let splitDist = distanceAlongLine(original, splitPt);
     let firstPiece = true;
+    // TODO Can we iterate over an array's contents and get the index at the same time?
+    let i = 0;
     for (let waypt of original.properties.waypoints) {
       let wayptDist = distanceAlongLine(
         original,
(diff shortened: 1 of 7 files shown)


components/draw/SplitRouteMode.svelte
```diff
@@ -0,0 +1,143 @@
...
+
+        gjScheme.update((gj) => {
+          // Keep the old ID for one, assign a new ID to the other
+          piece1.id = gj.features[snappedIndex].id;
+          piece2.id = newFeatureId(gj);
+
+          // The properties get lost. Copy everything to both
+          piece1.properties = gj.features[snappedIndex].properties;
+          // "Deep clone"
+          piece2.properties = JSON.parse(JSON.stringify(piece1.properties));
```
dabreegster, line 85: And a reminder, this round-trip is apparently the most straightforward way to "deep clone" something in JS. :\

stores.js
```diff
@@ -64,6 +64,10 @@ export function setCurrentlyEditing(id) {
 // Although this implementation may appear to ID features in order (1, 2, 3,
 // etc), this is NOT an invariant. Do not assume this; it will not be true as
 // soon as a user deletes or reorders an intervention.
+//
+// NOTE! If you call this twice in a row in a `gjScheme.update` transaction
```
dabreegster, line 68: ... You can guess that this happened to me ;) I don't think we'll be adding many features at a time, but if we do, we could allocate and return multiple IDs from here at a time for convenience

dabreegster:
Alright, give https://acteng.github.io/atip/split_line/scheme.html?authority=West of England Combined Authority a go! After you split a route, you should be able to edit its geometry, and everything should work as you expect. If it doesn't, please let me know how to reproduce what you've broken.

robinlovelace-ate:
Giving it a go.

robinlovelace-ate:
OK first UX niggle hit: the route disappeared! I guess this is because I hit "Split route" before "Finish route". As shown below.


https://user-images.githubusercontent.com/122299965/231999979-5d5f0bb2-2fd8-4f49-9019-30a06603cac7.mp4

robinlovelace-ate:
Overall the functionality is good. There are no regressions that I can see. So :+1: on this being merged.

components/draw/SplitRouteMode.svelte
```diff
@@ -0,0 +1,143 @@
...
+    changeMode("edit-attribute");
+  });
+
+  // Rendering
+  let source = "split-route";
+  overwriteSource($map, source, {
+    type: "geojson",
+    data: emptyGeojson(),
+  });
+  // TODO Scissors icon?
```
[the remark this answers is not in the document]
robinlovelace-ate: Yes, that would be awesome!

robinlovelace-ate (approved):
This is an impressive new feature. It has no downsides as far as I can tell. There are a few things I thought of outlined in the points and video below.

- Should the new sub-routes have a default name, e.g. "Untitled Route Section 1" etc? That would encourage people to name the original route and provide a default way of tracking the sections and showing that these are sections that are part of a route.
- Should the split tool stay "on" by default? If shown with scissors that would speed things up but doing all your splitting in one go without having to click on "Split route" again.


https://user-images.githubusercontent.com/122299965/232003372-c17248fe-3cf7-4bc0-948d-4f0097a14964.mp4

There are quite a few branches. We definitely want this. So I suggest merging soon to keep momentum up on the main branch without different branches diverging at this stage.

dabreegster:
> OK first UX niggle hit: the route disappeared! I guess this is because I hit "Split route" before "Finish route".

Added to #124. @Sparrow0hawk and I discussed how mode switching should work. Would a better UX be a warning saying "you're about to discard your changes by changing to a different draw tool. do you want to keep this or not?"

dabreegster:
> Should the new sub-routes have a default name, e.g. "Untitled Route Section 1" etc? That would encourage people to name the original route and provide a default way of tracking the sections and showing that these are sections that are part of a route

Tracking in #119. I think we should do something here, but not sure what. Another change I want to make is auto-naming routes of the form "Main St from X Road to Y Road", in which case splitting needs to do something clever. And yet another idea is letting people group interventions in some logical way -- a route split into segments, with a polygon showing a particular new crossing or junction treatment. Or individual parklets or crossings or modal filters within an overall area polygon.

dabreegster:
Merging, but followup tasks are tracked in the issue

> Should the split tool stay "on" by default?

Done

PR #115, opened 2023-04-11
Take over point and polygon drawing

dabreegster:
This PR completely removes `mapbox-gl-draw`, an endless source of frustration, in favor of our own custom tools for drawing/editing points and polygons (so we can be take full ownership of any frustration!).

It's incomplete; polygon editing needs more work. But the general idea, and complete removal of gl-draw, is ready.


dabreegster:
I think this is good enough to get feedback now:

https://user-images.githubusercontent.com/1664407/231734191-9bfd022c-d1f5-49ff-b913-3d0309bfc5e3.mp4

Some more work needed, none of which I think is blocking:

- [ ] Add cancel buttons (and the escape key) for polygon and point mode. The question is, where should those buttons go for "Edit geometry" mode? (We're weird for editing a route right now; the controls go under the wrong mode's button)
- [ ] Change the cursor to indicate dragging, hovering, etc
- [ ] Guarantee layer ordering in the polygon tool
- [ ] Fix HMR; especially with the polygon tool, live edits to the code usually break it, you have to hard-refresh
- [ ] Consider https://terradraw.io/ instead. Its polygon drawing mode is less usable than ours, in my opinion, but it could be useful to consolidate effort / invest there.

dabreegster:
Preview at https://acteng.github.io/atip/draw_controls_from_scratch/scheme.html?authority=West%20of%20England%20Combined%20Authority#11.53/51.4644/-2.5596

robinlovelace-ate:
https://user-images.githubusercontent.com/122299965/231752075-5cddf036-8dfd-4e0d-961e-da08f36bb8cb.mp4

As shown below, the polygon tool functionality is really smooth.

robinlovelace-ate (approved):

robinlovelace-ate:
https://user-images.githubusercontent.com/122299965/231818497-4163a511-8fc2-4f5e-8e65-2c804b63865e.mp4

robinlovelace-ate:
OK no it's fine: I just missed the instruction to hit Enter when done. That is good UX IMO, I'm just slow!

dabreegster:
> One thought for you @dabreegster, how about "Press Enter or double click on the polygon to finish" for people like me who don't read the docs, or who don't use the keyboard much?

I started #124 to record this and similar things, but will merge this now and continue in smaller followup PRs. I like the idea of double clicking the polygon to finish. geojson.io and others use "double click the last vertex", but we don't really have a concept of "last vertex" in this style, because when you click, we insert a point in the middle of the nearest existing line segment. (Which I _think_ is nicer, but that's also a UX question!)

Thanks for the UX review, I'll merge now. Feedback of technical or design welcome anytime.

Issue #89, opened 2023-03-19
Port to TypeScript


dabreegster:
I'm uncomfortable writing increasingly large/complex software, which'll have multiple contributors, in a language without static types. But so far I've used TypeScript myself very little, and hit a number of snags with it, over in https://github.com/dabreegster/osm2svelte. Do we have any TypeScript experience on the incoming team? I'd like to have a discussion over how soon we should get used to using it, and what to watch out for.

Robinlovelace:
All good questions, one for Slack, let's ask and nudge people this coming week.

Sparrow0hawk:
Interesting counter point to typescript is maybe just using JSDocs.

This is how they do it with SvelteKit - https://www.reddit.com/r/sveltejs/comments/1219z04/thoughts_on_this_twitter_thread_svelte_and_ts/?utm_source=share&utm_medium=ios_app&utm_name=ioscss&utm_content=2&utm_term=1

I don't have a particular dog in this fight (have already started looking at TS anyway).

dabreegster:
Close when we've enabled strict mode (no implicit anys, no nulls)

dabreegster:
https://github.com/gcanti/newtype-ts looking at this

PR #125, opened 2023-04-15
Partially cutover to Typescript

dabreegster:
My development speed has been very much slowed down by lack of a compiler. So, I've initially "ported" ATIP to Typescript. It's actually a gradual process; the TS compiler can handle plain JS and already catch a bunch of possible errors. That's mostly what I've done here; I whittled down `npm run check` to only one error, but not yet started taking true advantage of TS yet. I'll work on that, possibly gradually, in a followup. Ultimately we can enforce it by adding `noImplicitAny: true` in tsconfig, or possibly `strict: true`. #89 

No behavioral changes that I can spot, and the few tests still pass.


components/BoundaryLayer.svelte
```diff
@@ -1,11 +1,11 @@
-<script>
+<script lang="ts">
   import geojsonExtent from "@mapbox/geojson-extent";
   import mask from "@turf/mask";
   import {
     drawPolygon,
     overwriteSource,
     overwriteLayer,
-  } from "../maplibre_helpers.js";
+  } from "../maplibre_helpers";
```
dabreegster, line 8: This is pretty bizarre to me. Actually we could leave it as `.js`, even though we've renamed the file to `.ts`! There's also some tsconfig option we can provide to allow specifying `.ts` explicitly. But AFAICT, it's common practice to just leave off the file extension. Happy to do otherwise; guidance here was unclear and depended on the build process and stuff

components/EntireScheme.svelte
```diff
@@ -81,7 +81,7 @@
     // TODO Should we prompt before deleting the current scheme?
     reader.onload = (e) => {
       try {
-        gjScheme.set(backfill(JSON.parse(e.target.result)));
+        gjScheme.set(backfill(JSON.parse(e.target.result as string)));
```
dabreegster, line 84: It could also be `ArrayBuffer`; the type checker's not smart enough to handle the `readAsText` below, so a type assertion is apparently the normal response

components/InterventionList.svelte
```diff
@@ -75,9 +75,9 @@
   {#each $gjScheme.features as feature, i (feature.id)}
     <AccordionItem
       bind:open={feature.properties.editing}
-      on:click={startEditing(feature.id)}
-      on:mouseenter={sidebarHover(feature.id)}
-      on:mouseleave={sidebarHover(null)}
+      on:click={() => startEditing(feature.id)}
+      on:mouseenter={() => sidebarHover(feature.id)}
+      on:mouseleave={() => sidebarHover(null)}
```
dabreegster, line 80: Because all these event handlers actually do pass in an event object, I was getting some error here before. We can use this syntax to ignore the event param

components/Layout.svelte
```diff
@@ -13,7 +13,11 @@
     </nav>
     <slot name="sidebar" />
   </div>
-  <div class="sidebar-toggle rounded-rect" on:click={toggleSidebar}>&rarr;</div>
+  <button
```
dabreegster, line 16: I finally fixed that A11y warning about a clickable div without key handlers. Making it a button (but still styling it the same way) seems to fix, and allows someone to tab around and press enter (I confirmed this).

components/Map.svelte
```diff
@@ -20,7 +20,7 @@
       style: `https://api.maptiler.com/maps/${style}/style.json?key=MZEJTanw3WpxRvt7qDfo`,
       hash: true,
     });
-    map.addControl(new ScaleControl());
+    map.addControl(new ScaleControl({}));
```
dabreegster, line 23: It takes `object?` but was angry without passing something in

components/draw/GeometryMode.svelte
```diff
@@ -135,11 +135,11 @@
 
     currentlyEditing = id;
 
-    if (feature.geometry.type == "LineString") {
+    if (feature!.geometry.type == "LineString") {
```
dabreegster, line 138: `!` means assert that `feature` isn't `undefined`. This is valid because we only call `startEditing` with IDs of stuff that exists

components/draw/SplitRouteMode.svelte
```diff
@@ -219,7 +219,7 @@
     // TODO Is there a cheaper way to do this?
     let start = lineFeature.geometry.coordinates[0];
     let sliced = lineSlice(start, pointFeature, lineFeature);
-    return length(sliced.geometry, { units: "kilometers" }) * 1000.0;
+    return length(sliced, { units: "kilometers" }) * 1000.0;
```
dabreegster, line 222: Apparently the implementation is flexible about taking `Feature<LineString>` or its geometry, but the type checker got angry otherwise

maplibre_helpers.ts
```diff
@@ -31,7 +32,11 @@ export function drawPolygon(color, opacity) {
   };
 }
 
-export function drawCircle(color, radius, opacity = 1.0) {
+export function drawCircle(
+  color,
+  radius,
+  opacity: DataDrivenPropertyValueSpecification<number> = 1.0
```
dabreegster, line 38: Because we call this method a few times with one of those long `["case", ...]` expressions, we need to do this. Figuring out the name of this type was not easy; I couldn't find the TS types anywhere online in the normal docs. So I just grepped the maplibregl code (conveniently in `node_modules` for `circle-opacity` and figured this out

svelte.config.js
```diff
@@ -0,0 +1,5 @@
+import { vitePreprocess } from "@sveltejs/vite-plugin-svelte";
```
dabreegster, line 1: This is some preprocessing thing needed for `lang=ts` to work

tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
dabreegster, line 1: I mostly copied this from a vite template. I don't really know what all of this means, or if any of it is redundant with the defaults

tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Typecheck JS in `.svelte` and `.js` files by default.
+     * Disable checkJs if you'd like to use dynamic types in JS.
+     * Note that setting allowJs false does not prevent the use
+     * of JS in `.svelte` files.
+     */
+    "allowJs": true,
+    "checkJs": true,
+    "isolatedModules": true,
+    "types": ["vite/client"]
```
dabreegster, line 17: Added so that when we import static assets like SVGs or GeoJSON URLs, tsc understands it

tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "include": ["*.js", "*.ts", "*.svelte", "components/**/*.js", "components/**/*.ts", "components/**/*.svelte", "tests/*"]
```
dabreegster, line 19: This is really important to get right; I got it slightly wrong at first and couldn't figure out why there weren't errors

vite.config.ts
```diff
@@ -4,9 +4,6 @@ import { svelte } from "@sveltejs/vite-plugin-svelte";
 import wasmPack from "vite-plugin-wasm-pack";
 
 export default defineConfig({
-  test: {
```
dabreegster, line 7: The type checker let us know that nothing actually pays attention to this! It's a `vitest` property, but we're not using that anymore; just raw playwright

dabreegster (commented):
I'm going to merge ASAP, otherwise merge conflict hell. Async feedback welcome. I'll do similarly for followup PRs.

PR #127, opened 2023-04-16
Port all code to TypeScript

dabreegster:
This pretty much converts all of our code to TypeScript, leaving no errors. #89 

As the diff shows, this wasn't much work, and gives us great benefits for understanding code, maintaining it, and developing faster without hitting silly preventable bugs.

Still not done; we could enable strict mode, ban implicit nulls, etc.

No behavioral changes


src/App.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
   import "carbon-components-svelte/css/white.css";
 
+  import type { FeatureCollection, Polygon } from "geojson";
```
dabreegster, line 4: For reference (because I couldn't find these anywhere in HTML docs nicely): https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/geojson/index.d.ts

src/App.svelte
```diff
@@ -49,12 +50,12 @@
...
   }
 
-  let boundaryGeojson;
+  let boundaryGeojson: FeatureCollection<Polygon>;
   onMount(async () => {
     boundaryGeojson = await loadAuthorityBoundary();
   });
 
-  async function loadAuthorityBoundary() {
+  async function loadAuthorityBoundary(): Promise<FeatureCollection<Polygon>> {
```
dabreegster, line 58: Things like this are a little curious to me. Nothing checks the return type is correct; TS appears to just trust us

src/ChooseArea.svelte
```diff
@@ -9,9 +9,9 @@
 
   let showAbout = false;
 
-  let inputValue;
+  let inputValue: string;
   let dataList;
```
dabreegster, line 13: I couldn't easily figure out this type, will work on it later. Maybe there's a trick with VS Code to infer it?

src/lib/EntireScheme.svelte
```diff
@@ -89,13 +90,13 @@
...
 
-  function backfill(json) {
+  function backfill(json: Scheme) {
     let idCounter = 1;
     for (let f of json.features) {
       // Look for any LineStrings without length_meters. Old route-snapper versions didn't fill this out.
       if (f.geometry.type == "LineString" && !f.properties.length_meters) {
         f.properties.length_meters =
-          length(f.geometry, { units: "kilometers" }) * 1000.0;
+          length(f, { units: "kilometers" }) * 1000.0;
```
dabreegster, line 99: Another case where the library is flexible in practice, but for some reason it doesn't type-check to do this. It makes more sense to pass in the whole feature anyway

src/lib/InterventionList.svelte
```diff
@@ -1,18 +1,19 @@
...
-  } from "../stores.js";
+  } from "../stores";
 
-  function interventionName(feature) {
+  function interventionName(feature: FeatureUnion): string {
     if (feature.properties.name) {
       return feature.properties.name;
     }
-    var noun = feature.properties.intervention_type;
+    var noun: string = feature.properties.intervention_type;
```
dabreegster, line 16: If we don't specify this, it actually grabs the more specific type and complains when we assign other strings below

src/lib/InterventionList.svelte
```diff
@@ -51,10 +52,10 @@
     currentHover.set(id);
   }
 
-  function onKeydown(e) {
+  function onKeydown(e: KeyboardEvent) {
     const id = $currentlyEditing;
     if (e.key == "Delete") {
-      const tag = e.originalTarget.tagName;
+      const tag = (e.target as HTMLElement).tagName;
```
dabreegster, line 58: https://developer.mozilla.org/en-US/docs/Web/API/Event/originalTarget
Oops, we were using a Firefox-only property before! TS to the rescue.

src/lib/draw/AttributeMode.svelte
```diff
@@ -70,7 +70,9 @@
 
       // Extent of points is defined in a weird way, special-case it
       if (feature.geometry.type == "Point") {
-        $map.flyTo({ center: feature.geometry.coordinates });
+        $map.flyTo({
+          center: feature.geometry.coordinates as [number, number],
```
dabreegster, line 74: See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/045220a0682a1f437c08610db4ad6eaf76b2f9f0/types/geojson/index.d.ts#L38
geojson types use `number[]` even though it should really be 2D

src/lib/draw/SplitRouteMode.svelte
```diff
@@ -75,7 +82,10 @@
       // We clicked the map, stop the tool
       changeMode("edit-attribute");
     } else {
-      let result = lineSplit($gjScheme.features[snappedIndex], cursor);
+      let result = lineSplit(
+        $gjScheme.features[snappedIndex] as Feature<LineString>,
```
dabreegster, line 86: Static analysis doesn't know we picked a linestring (vs a point or polygon), so we assert the type here

src/lib/draw/SplitRouteMode.svelte
```diff
@@ -92,10 +102,20 @@
...
+            piece2,
+            cursor
+          );
 
           // Replace the one LineString we snapped to with the two new pieces
-          gj.features.splice(snappedIndex, 1, piece1, piece2);
+          gj.features.splice(
+            snappedIndex,
+            1,
+            piece1 as OurFeature<LineString>,
```
dabreegster, line 116: We've been dealing with the generic geojson Feature for piece1 and piece2. Above we added in properties, set the ID, etc and made it a valid feature for our definition. So this is another case where static analysis can't figure it out, and we just say "trust us"

src/lib/draw/Toolbox.svelte
```diff
@@ -30,7 +30,7 @@
   let splitRouteMode;
 
   // This must be used; don't manually change mode
-  function changeMode(newMode) {
+  function changeMode(newMode: string) {
```
dabreegster, line 33: I'll make a union type for `mode` later, and make sure we don't typo the string

src/lib/draw/point_tool.ts
```diff
@@ -1,21 +1,29 @@
...
   drawCircle,
 } from "../../maplibre_helpers";
-import { colors, circleRadius } from "../../colors.js";
+import { colors, circleRadius } from "../../colors";
 
+const source = "edit-point-mode";
+
+// Note this uses the geojson Feature, not our specialization in types.ts
 export class PointTool {
+  map: Map;
```
dabreegster, line 15: This kind of thing brings me great comfort, just seeing all the state and types upfront

src/lib/draw/polygon_tool.ts
```diff
@@ -10,10 +12,22 @@ import {
...
+const source = "edit-polygon-mode";
 
 export class PolygonTool {
-  constructor(map) {
+  map: Map;
+  active: boolean;
+  eventListeners: ((f: Feature<Polygon>) => void)[];
+  points: Position[];
+  cursor: Feature<Point> | null;
+  hoverPolyon: boolean;
```
dabreegster, line 25: Now that we're TS, we could get fancier here and express that hovering is either nothing | the polygon | an index point. I'll look into that later

src/lib/draw/polygon_tool.ts
```diff
@@ -246,25 +258,25 @@ export class PolygonTool {
...
-  #polygonFeature() {
+  #polygonFeature(): Feature<Polygon> | null {
     if (this.points.length < 3) {
       return null;
     }
-    let f = {
+    // Deep clone here, or face the wrath of crazy bugs later!
+    let coordinates = [JSON.parse(JSON.stringify(this.points))];
+    coordinates[0].push(JSON.parse(JSON.stringify(coordinates[0][0])));
+    return {
```
dabreegster, line 268: For some reason TS got confused when we did `let f = {...}; return f`, so just switched it around

src/types.ts
```diff
@@ -0,0 +1,49 @@
+import type { Point, LineString, Polygon } from "geojson";
```
dabreegster, line 1: @robinlovelace-ate, since you've been having so much fun playing with the JSON schema format, here's a practical example expressed another way. This is ATIP's current geojson schema expressed using TypeScript, and it gets verified throughout the codebase

src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+export interface Scheme {
+  type: "FeatureCollection";
+  features: FeatureUnion[];
+  // Foreign members
+  scheme_name?: string;
+  authority?: string;
+  origin?: string;
+}
+
+// TODO Can we use a wildcard type, like Feature<? extends Geometry>
```
dabreegster, line 14: I could not figure how to say "an array of a Feature, where the specific Geometry type may vary for each" in TS yet, so this is the hack

src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  geometry: G;
+  properties: InterventionProps;
+}
+
+export interface InterventionProps {
+  intervention_type: "area" | "route" | "crossing" | "other";
+  name: string;
+  description: string;
+
+  // For LineStrings only
```
dabreegster, line 36: We could write the types more explicitly to enforce this, if we want to try that out

src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  length_meters?: number;
+  waypoints?: Waypoint[];
+
+  // Temporary state, not meant to be serialized
+  editing?: boolean;
+  hide_while_editing?: boolean;
+}
+
+export interface Waypoint {
```
dabreegster, line 45: This should be moved over to `route-snapper`

dabreegster (commented):
ASAP merging again

PR #128, opened 2023-04-17
Replace Carbon's accordion with our own. #90

dabreegster:
Along the way, create a few more stores to be more specific about map vs sidebar interactions.

https://user-images.githubusercontent.com/1664407/232451906-bf3fd349-c870-43fc-ae94-60b6cf6040bd.mp4

### DIFF
--- a/src/lib/AccordionItem.svelte
+++ b/src/lib/AccordionItem.svelte
@@ -1,0 +1,95 @@
+<script lang="ts">
+  import { slide } from "svelte/transition";
+  // TODO Make this more generic by taking some of these as props too or forwarding events
+  import {
+    formOpen,
+    mapHover,
+    openFromSidebar,
+    sidebarHover,
+  } from "../stores.js";
+
+  export let id: number;
+  export let label: string;
+
+  $: isOpen = $formOpen == id;
+  const toggle = () => {
+    formOpen.update((current) => {
+      if (current == id) {
+        return null;
+      } else {
+        return id;
+      }
+    });
+    if ($formOpen == id) {
+      // Always set this to null first, to force subscribers to see the update.
+      // It's possible to open something from the sidebar, close it (by
+      // clicking on the map or using the sidebar), then reopen the same thing.
+      openFromSidebar.set(null);
+      openFromSidebar.set(id);
+    }
+  };
+
+  // TODO Design choice: While a form is open and we hover on something else in
+  // the map, should we underline the item in the sidebar? For now, yes.
+  $: underlineRemotely = $mapHover == id;
+
+  let contents: HTMLDivElement;
+  function scroll() {
+    contents?.scrollIntoView({ behavior: "smooth" });
+  }
+</script>
+
+<button
+  on:click={toggle}
+  on:mouseenter={() => sidebarHover.set(id)}
+  on:mouseleave={() => sidebarHover.set(null)}
+  aria-expanded={isOpen}
+  class:underlined={underlineRemotely}
+  ><svg
+    style="tran"
+    width="20"
+    height="20"
+    fill="none"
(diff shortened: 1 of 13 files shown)


src/lib/AccordionItem.svelte
```diff
@@ -0,0 +1,95 @@
...
+    if ($formOpen == id) {
+      // Always set this to null first, to force subscribers to see the update.
+      // It's possible to open something from the sidebar, close it (by
+      // clicking on the map or using the sidebar), then reopen the same thing.
+      openFromSidebar.set(null);
+      openFromSidebar.set(id);
+    }
+  };
+
+  // TODO Design choice: While a form is open and we hover on something else in
```
dabreegster, line 32: The form is visibly open, so we don't need to underline its label. This differs from the hover effect being fixed on the map while a form is open -- we don't want 2 styles for "currently active object" and "possibly different object we're hovering on". Unless someone says otherwise. :)

src/lib/AccordionItem.svelte
```diff
@@ -0,0 +1,95 @@
...
+    stroke-width="2"
+    viewBox="0 0 24 24"
+    stroke="currentColor"><path d="M9 5l7 7-7 7" /></svg
+  >
+  {label}
+</button>
+{#if isOpen}
+  <div
+    bind:this={contents}
+    transition:slide={{ duration: 100 }}
```
dabreegster, line 64: The effect may be a little choppy; we spend 100ms sliding the accordion down, then scroll to make sure the form is in view. Any better ideas?

src/lib/InterventionList.svelte
```diff
@@ -26,74 +21,36 @@
...
-
-  function startEditing(id: number) {
-    console.log(`Clicked ${id} from the sidebar; going to edit-attribute`);
-    // Always set this to null first, to force subscribers to see the update.
-    // It's possible to open something from the sidebar, close it (by clicking
-    // on the map or using the sidebar), then reopen the same thing.
-    openFromSidebar.set(null);
-    openFromSidebar.set(id);
-
-    // Remove the editing property from everything else, so that the Accordion is hidden
```
dabreegster, line 43: All of this weird hack goes away

src/lib/draw/AttributeMode.svelte
```diff
@@ -17,12 +16,12 @@
 
   export function start() {}
   export function stop() {
-    setCurrentlyEditing(null);
+    formOpen.set(null);
   }
 
   // Calculate hover
   $map.on("mousemove", (e) => {
-    if (mode == thisMode && $currentlyEditing == null) {
```
dabreegster, line 25: This used to be necessary for behavior (don't hover another object when a form is open). Now we take care of that in `HoverLayer` better. We could still skip this calculation as an optimization, but it doesn't seem to matter much

src/lib/draw/HoverLayer.svelte
```diff
@@ -54,13 +61,15 @@
     "interventions-points"
   );
 
-  currentHover.subscribe((id) => {
+  // When a form is open, ignore regular map and sidebar interactions
```
dabreegster, line 64: I feel like this is a simplification -- let's not abuse what object is hovered from the map, sidebar, etc. Calculate all of those things in a straightforward way. Then when we need to make a decision based on multiple values, we can just combine them

dabreegster (commented):
Another thing fixed. Before, the collapsed sidebar still had elements visible and clickable!
Screenshot from 2023-04-17 11-07-15

Now we're good:
Screenshot from 2023-04-17 11-07-35

Going to merge this now, but async feedback welcome

robinlovelace-ate:
Amazing work Dustin!

PR #129, opened 2023-04-18
Eliminate Carbon

dabreegster:
This PR finishes removing the Carbon component system, which was a hasty original choice. #90 

In the short term, this gives us much easier full control over everything. Longer term, if we need to adopt the govuk frontend styles, this frees us up to do it.

Some style changes visible below. Very happy to make any changes.

https://user-images.githubusercontent.com/1664407/232755298-db7dddf3-3138-49eb-9aaa-ebcb4ba75e7c.mp4

### DIFF
--- a/src/ChooseArea.svelte
+++ b/src/ChooseArea.svelte
@@ -99,9 +99,6 @@
 <div class="left">
   <h1>Welcome to ATIP v1</h1>
   <button type="button" on:click={() => (showAbout = !showAbout)}>About</button>
-  <br />
-  <br />
-  <br />
 
   <p>Select Transport Authority or Local Authority District:</p>
   <div>
--- a/src/ChooseArea.svelte
+++ b/src/ChooseArea.svelte
@@ -3,7 +3,6 @@
   import "maplibre-gl/dist/maplibre-gl.css";
   import geojsonExtent from "@mapbox/geojson-extent";
   import About from "./lib/About.svelte";
-  import "carbon-components-svelte/css/white.css";
 
   import authoritiesUrl from "../assets/authorities.geojson?url";
 
--- a/src/lib/Form.svelte
+++ b/src/lib/Form.svelte
@@ -1,10 +1,4 @@
 <script lang="ts">
-  import {
-    RadioButtonGroup,
-    RadioButton,
-    TextArea,
-    TextInput,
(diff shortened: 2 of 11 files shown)


src/ChooseArea.svelte
```diff
@@ -100,9 +99,6 @@
 <div class="left">
   <h1>Welcome to ATIP v1</h1>
   <button type="button" on:click={() => (showAbout = !showAbout)}>About</button>
-  <br />
```
dabreegster, line 103: Carbon's style overrides did weird things with whitespace, forcing hacks like this previously

src/lib/Form.svelte
```diff
@@ -21,33 +15,56 @@
   }
 </script>
 
-<TextInput labelText="Name" bind:value={name} />
+<label
```
dabreegster, line 18: https://stackoverflow.com/questions/774054/should-i-put-input-elements-inside-a-label-element
No need to repeat IDs and `label for`

src/lib/Modal.svelte
```diff
@@ -0,0 +1,64 @@
+<script lang="ts">
+  import { onMount } from "svelte";
+
+  export let title: string;
+  export let open = false;
+</script>
+
+<!-- svelte-ignore a11y-click-events-have-key-events -->
```
dabreegster, line 8: The outer `background` div is fullscreen and adds the fade. If you click it, we close the modal. There's the a11y warning because it's not a `<button>`. I think that's actually fine from an a11y perspective; there's an explicit close button inside the modal. If you were tabbing over everything and selected the entire screen, it'd be confusing as a button anyway

src/lib/Modal.svelte
```diff
@@ -0,0 +1,64 @@
...
+<div
+  class="background"
+  on:click={() => (open = false)}
+  class:open
+  class:closed={!open}
+>
+  <div
+    class="content centered-fullscreen"
+    on:click|stopPropagation={() => null}
```
dabreegster, line 17: Empty click handler just to stop the click from counting on the outer div

dabreegster (commented):
Merging, but reviews later welcome

robinlovelace-ate:
Amazing work Dustin!

PR #130, opened 2023-04-19
Use TypeScript more

dabreegster:
We have no errors with the current `tsconfig.json`, but adding `strict: true` yields around 200 problems. This PR whittles it down to 38. #89 

No behavioral changes

### DIFF
--- a/src/lib/draw/AttributeMode.svelte
+++ b/src/lib/draw/AttributeMode.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import geojsonExtent from "@mapbox/geojson-extent";
+  import { bbox } from "../../maplibre_helpers";
   import {
     map,
     gjScheme,
@@ -67,15 +67,15 @@
       // being edited. (Don't do this when clicking the object on the map.)
       let feature = $gjScheme.features.find((f) => f.id == id);
 
-      // Extent of points is defined in a weird way, special-case it
+      // Padding around a point looks odd; special case it by keeping the current zoom
       if (feature.geometry.type == "Point") {
         $map.flyTo({
           center: feature.geometry.coordinates as [number, number],
+          duration: 500,
         });
       } else {
-        $map.fitBounds(geojsonExtent(feature), {
+        $map.fitBounds(bbox(feature), {
           padding: 200,
-          animate: true,
           duration: 500,
         });
       }
--- a/src/lib/draw/AttributeMode.svelte
+++ b/src/lib/draw/AttributeMode.svelte
@@ -30,12 +30,7 @@
           "interventions-polygons",
         ],
       });
-      // TODO ? syntax
-      var newHoverId = null;
-      if (results.length > 0) {
-        newHoverId = results[0].id;
-      }
-      mapHover.set(newHoverId);
+      mapHover.set((results[0]?.id as number) || null);
     }
   });
   $map.on("mouseout", () => {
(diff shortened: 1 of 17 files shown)


src/lib/draw/AttributeMode.svelte
```diff
@@ -29,12 +30,7 @@
           "interventions-polygons",
         ],
       });
-      // TODO ? syntax
-      var newHoverId = null;
-      if (results.length > 0) {
-        newHoverId = results[0].id;
-      }
-      mapHover.set(newHoverId);
+      mapHover.set((results[0]?.id as number) || null);
```
dabreegster, line 33: Kind of awkward, but this is as clean as I can make it. The `as number` is necessary because the ID might be `string` according to MapLibre's API. But because we only deal with numeric IDs, the cast is safe

src/lib/draw/GeometryMode.svelte
```diff
@@ -123,21 +126,22 @@
   function startEditing(id: number) {
     mapHover.set(null);
 
-    let feature;
+    let maybeFeature: FeatureUnion | null = null;
```
dabreegster, line 129: I tried leaving off the `= null` part entirely, but then got weird errors about it being uninitialized when we use it; maybe it didn't understand the store.update bit

src/lib/draw/InterventionLayer.svelte
```diff
@@ -53,19 +56,20 @@
...
-    colors.area,
-    "route",
-    colors.route,
-    "crossing",
-    colors.crossing,
-    "other",
-    colors.other,
-    "white",
-  ];
+  const colorByInterventionType: DataDrivenPropertyValueSpecification<string> =
```
dabreegster, line 59: For some reason this didn't type-check, had to insist on it here

src/lib/draw/SplitRouteMode.svelte
```diff
@@ -50,13 +51,16 @@
       $map.unproject(e.point).distanceTo($map.unproject(nearbyPoint)) / 1000.0;
 
     // Are we snapped to anything?
-    let candidates = [];
+    let candidates: [number, Position, number][] = [];
```
dabreegster, line 54: I'd love to have Rust-style wrapper types and say something like [Index, Position, Distance] here, but...

src/lib/draw/SplitRouteMode.svelte
```diff
@@ -50,13 +51,16 @@
...
     for (let [index, f] of $gjScheme.features.entries()) {
       if (f.geometry.type == "LineString") {
         let snapped = nearestPointOnLine(f.geometry, cursor, {
           units: "kilometers",
         });
-        if (snapped.properties.dist <= thresholdKm) {
+        if (
+          snapped.properties.dist != undefined &&
```
dabreegster, line 61: The property is optional in the declared types, but scanning through the implementation, not sure when it's possible to be undefined. Couldn't figure out how to express this type narrowing more succinctly; I really dislike that the truthiness check in JS/TS treats 0 and "" as false

src/lib/draw/SplitRouteMode.svelte
```diff
@@ -82,36 +86,37 @@
       // We clicked the map, stop the tool
       changeMode("edit-attribute");
     } else {
+      // TODO Can we avoid using ! everywhere here?
```
dabreegster, line 89: I dunno why tsc can't figure out that `snappedIndex != null` in this branch...

src/lib/draw/types.ts
```diff
@@ -0,0 +1,7 @@
+export type Mode =
```
dabreegster, line 1: It catches typos!

Issue #131, opened 2023-04-21
Handle incorrect App parameters


dabreegster:
We get `authorityName` from URL params and require it to match something in `assets/authorities.geojson`. Near
https://github.com/acteng/atip/blob/9bd25471db3470553015c1dd3f0d9a0fd5be0294/src/App.svelte#L29 we insist everything exists and is fine, but we should cleanly handle problems with an error message/page or auto-changing to some default

Leaving this open as a good starter bug for Svelte

ELBuenFercho:
Hello ! I'll do my best to work on it, thanks  (is it necessary to assign it to myself?) thanks !!!

dabreegster:
Awesome! Happy to chat over Slack/Teams/etc about if you like. First we should decide how to handle the error (what we show to the user), then can implement it and write a Playwright test for it

robinlovelace-ate:
Testing this with the following link: https://acteng.github.io/atip/scheme.html?authority=nonexistentauthority#10.91/50.6717/-1.3282

robinlovelace-ate:
This is what I see, what's wrong with this from user perspective? 

![image](https://github.com/acteng/atip/assets/122299965/09f55f3d-4c4b-4fda-9484-371a8d42caf3)

Pete-Y-CS:
I guess we'd want to redirect to the select authority page and display an error 
> the authority requested does not exist

dabreegster:
Sounds good to me. You reckon we show the error first, then redirect? Or the other way around? Maybe something like a modal that says the error, isn't closeable by clicking away, then only has one button to "go back and try again" or something?

Pete-Y-CS:
Good question. I was thinking redirect and show the error on the select authority because I guess it's kinda where they need to know there was an error in this case: they need to know to select a predefined authority

dabreegster:
That sounds sensible to me. Just thinking through the implementation of that a bit... using a query param to plumb the incorrect authority name?

Probably the most common way of arriving in this error state right now is from the scheme browse page, because we haven't yet correctly retained the authority field from the original GJ input files when merging. Redirecting to the homepage in that case doesn't make sense, but it'd be simpler to just detect the error there, display a warning, and stay on the page.

Pete-Y-CS:
Hmm good point. I had planned on using a query param (in fact it's already done: 
![image](https://github.com/acteng/atip/assets/133757676/9ff66b09-85b6-4782-8c01-db7a7b8691bc)
)

[...]

PR #132, opened 2023-04-21
More TS fixes. Down to 5 strict errors!

dabreegster:
Home-stretch! The last few problems are harder to deal with, though. #89

No behavioral changes.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -26,7 +26,8 @@
   let prod = window.location.hostname.includes("atip.uk");
 
   const params = new URLSearchParams(window.location.search);
-  let authorityName: string = params.get("authority");
+  // TODO Add validation and some kind of error page
+  let authorityName: string = params.get("authority")!;
   let style: string = params.get("style") || "streets";
 
   // TODO Slight hack. These files are stored in an S3 bucket, which only has
@@ -56,9 +57,9 @@
   async function loadAuthorityBoundary(): Promise<FeatureCollection<Polygon>> {
     const resp = await fetch(authoritiesUrl);
     const body = await resp.text();
-    const geojson = JSON.parse(body);
+    const geojson: FeatureCollection<Polygon> = JSON.parse(body);
     geojson.features = geojson.features.filter(
-      (feature) => feature.properties.name == authorityName
+      (feature) => feature.properties?.name == authorityName
     );
     return geojson;
   }
--- a/src/ChooseArea.svelte
+++ b/src/ChooseArea.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import type { FeatureCollection } from "geojson";
   import maplibregl from "maplibre-gl";
   import "maplibre-gl/dist/maplibre-gl.css";
   import { bbox } from "./maplibre_helpers";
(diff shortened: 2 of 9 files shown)


src/App.svelte
```diff
@@ -26,7 +26,8 @@
   let prod = window.location.hostname.includes("atip.uk");
 
   const params = new URLSearchParams(window.location.search);
-  let authorityName: string = params.get("authority");
+  // TODO Add validation and some kind of error page
```
dabreegster, line 29: Leaving this for a new starter! #131

src/ChooseArea.svelte
```diff
@@ -20,24 +21,24 @@
...
+    json.features = json.features.filter((f) => f.properties!.level == "TA");
 
     let map = new maplibregl.Map({
       container: "map",
       style:
         "https://api.maptiler.com/maps/streets/style.json?key=MZEJTanw3WpxRvt7qDfo",
     });
 
-    let hoverId = null;
+    let hoverId: number | null = null;
```
dabreegster, line 41: We know the IDs are numbers, and not strings, because we use `generateIds: true`

src/ChooseArea.svelte
```diff
@@ -72,9 +73,9 @@
       });
 
       map.on("mousemove", layer, (e) => {
-        if (e.features.length > 0) {
+        if (e.features!.length > 0) {
```
dabreegster, line 76: Maybe we should use error handling, like optional chaining with `?`, more? But I've never seen when these events fire without features

src/lib/draw/point_tool.ts
```diff
@@ -10,12 +10,17 @@ import { colors, circleRadius } from "../../colors";
 
 const source = "edit-point-mode";
 
+// Properties are guaranteed to exist
+type FeatureWithProps<G extends Geometry> = Feature<G> & {
+  properties: { [name: string]: any };
```
dabreegster, line 15: I'm still confused why we need the extra `{}` here to express that it's a map from string to any, but... this is what type-checks

src/lib/draw/polygon_tool.ts
```diff
@@ -16,15 +23,19 @@ import { colors, circleRadius } from "../../colors";
...
   map: Map;
   active: boolean;
-  eventListeners: ((f: Feature<Polygon>) => void)[];
+  eventListeners: ((f: FeatureWithProps<Polygon>) => void)[];
   points: Position[];
   cursor: Feature<Point> | null;
-  hoverPolyon: boolean;
-  hoverPoint: number | null;
-  dragging: boolean;
+  // The number is an index into points
```
dabreegster, line 37: I collapsed a bunch of state here. One field to say what we're hovering on, and to say if we're dragging

src/lib/draw/polygon_tool.ts
```diff
@@ -306,7 +307,9 @@ function pointsToLineSegments(points: Position[]): Feature<LineString>[] {
         type: "LineString",
         coordinates: [points[points.length - 1], points[0]],
       },
+      properties: {},
     });
   }
-  return lines;
+  // TODO TS is failing here
```
dabreegster, line 313: And I don't know why :(

src/maplibre_helpers.ts
```diff
@@ -60,12 +64,16 @@ export function drawCircle(
...
+  map: Map,
+  id: string,
+  source: SourceSpecification
+) {
   if (map.getSource(id)) {
     // First remove all layers using this source
     let layers = [];
     for (let layer of map.getStyle().layers) {
-      if (layer.source == id) {
+      if ("source" in layer && layer.source == id) {
```
dabreegster, line 76: `LayerSpecification` has a bunch of union cases, and one of them has no `source`. Apparently this is the canonical way to deal

src/maplibre_helpers.ts
```diff
@@ -78,15 +86,19 @@ export function overwriteSource(map, id, source) {
   map.addSource(id, source);
 }
 
-export function overwriteLayer(map, layer, beforeId: string | null = null) {
+export function overwriteLayer(
+  map: Map,
+  layer,
```
dabreegster, line 91: You'll note I'm not saying `LayerSpecification` here yet, because the errors absolutely explode :) One of the hard things remaining

dabreegster:
We have our first playwright flake :(
https://github.com/acteng/atip/actions/runs/4763337365/jobs/8466648609?pr=132
It runs fine for me locally, triggering it through GH again with a small cleanup commit

PR #133, opened 2023-04-21
Overhaul toolbox controls

### LINKED ISSUE #113: Route tool button should be the same size/prominence as the other buttons
As shown here: 

image

dabreegster:
Two big changes here! First up, I've stopped using https://github.com/dabreegster/route_snapper/blob/main/route-snapper/lib.js, the JS wrapper around the WASM library. Instead I've ported it over to `route_tool.ts`, converted to TS, and made some improvements to better integrate with ATIP. This feels like the right approach long-term; we want tighter control over everything here, and it's slower and more annoying to iterate over in the other repo, publishing new NPM packages, etc.

Second, I've detangled the on-screen controls for each of the modes a bit. This lets us use the same controls whether we're creating a new point/polygon/route or modifying the geometry of an existing one! The new directory structure should make things pretty clear.

Behavioral changes intentional; feels like polish to me. Fixes #113.

https://user-images.githubusercontent.com/1664407/233630674-97403b46-efd1-43bf-a4fd-e730d71f9cac.mp4


src/lib/draw/route/route_tool.ts
```diff
@@ -0,0 +1,258 @@
+import { JsRouteSnapper } from "route-snapper";
```
dabreegster, line 1: Here's the port of the JS library. Pretty similar logically.

src/lib/draw/route/route_tool.ts
```diff
@@ -0,0 +1,258 @@
...
+  // Activate the tool with blank state.
+  start() {
+    // If we were already active, don't do anything
+    // TODO Or... error? Why'd this happen?
+    if (this.active) {
+      return;
+    }
+
+    this.active = true;
+    // Otherwise, shift+click breaks
```
dabreegster, line 159: This was new to add, not sure why it wasn't a problem before

PR #138, opened 2023-04-24
Manage callbacks better

dabreegster:
As the edits near `README` describe, we have to be a bit careful about unregistering any document/map/etc event handlers when a Svelte component gets torn down. Otherwise as you edit a component, every previous version of the callback keeps running! This PR fixes most remaining problems; see comments below for the strategy.

The last big remaining problem is the custom observable pattern used in all the `*tool.ts` classes with `addEventListener`. We need to clean those up too.

### DIFF
--- a/src/lib/draw/AttributeMode.svelte
+++ b/src/lib/draw/AttributeMode.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+  import { onDestroy } from "svelte";
+  import { type MapMouseEvent } from "maplibre-gl";
   import type { Mode } from "./types";
   import { bbox } from "../../maplibre_helpers";
   import {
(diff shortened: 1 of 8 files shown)


src/lib/draw/AttributeMode.svelte
```diff
@@ -55,32 +94,7 @@
         formOpen.set(null);
       }
     }
-  });
-
-  openFromSidebar.subscribe((id) => {
```
dabreegster, line 60: Also changed from subscribe to `$` to get automatic cleanup

src/lib/draw/AttributeMode.svelte
```diff
@@ -21,7 +23,44 @@
   }
 
   // Calculate hover
-  $map.on("mousemove", (e) => {
+  $map.on("mousemove", onMouseMove);
```
dabreegster, line 26: This is an example of the change within a component. Each event handler should be a function, so we can refer to it for `on` and `off`

src/lib/draw/events.ts
```diff
@@ -0,0 +1,42 @@
+import type { Map } from "maplibre-gl";
+
+// Use this to create and destroy event handlers on the map and document. When
```
dabreegster, line 3: This is a bit fancier, for classes. We can't just say `map.on("click", this.onClickHandler)`, because we need to bind the current object as `this`. That binding step messes up calling `off` and having it match. There's maybe some simpler things we could do by defining all methods as arrow functions; see https://osmus.slack.com/archives/C01G3D28DAB/p1682086599507859 for discussion

src/lib/draw/point/point_tool.ts
```diff
@@ -78,10 +85,10 @@ export class PointTool {
   stop() {
     this.cursor = null;
     this.active = false;
-    this.#redraw();
+    this.redraw();
   }
 
-  #redraw() {
+  private redraw() {
```
dabreegster, line 91: Also switching to more clear TS syntax

Issue #124, opened 2023-04-14
Smaller feedback on drawing tools


dabreegster:
I'll use this to record smaller followup tasks raised in PRs. We can split into separate issues when it's useful.

- [ ]  Add cancel buttons (and the escape key) for polygon and point mode. The question is, where should those buttons go for "Edit geometry" mode? (We're weird for editing a route right now; the controls go under the wrong mode's button)
- [ ] Change the cursor to indicate dragging, hovering, etc
- [ ] Guarantee layer ordering in the polygon tool
- [ ] Fix HMR; especially with the polygon tool, live edits to the code usually break it, you have to hard-refresh
- [ ] Finish drawing a polygon by double clicking the last vertex (similar to delete?) or double clicking the polygon itself

PR #139, opened 2023-04-24
Consistent controls to stop drawing

dabreegster:
This adds a clickable cancel button to all modes, and lets pressing the escape key also work. #124 

Weird caveat: listening for the escape key only works after clicking on the page. Not sure why.


src/lib/draw/KeyHandler.svelte
```diff
@@ -0,0 +1,19 @@
+<script lang="ts">
+  interface Cancelable {
```
dabreegster, line 2: The magic of duck typing!

Issue #136, opened 2023-04-24
Specialize ATIP for planning


dabreegster:
Specialize ATIP for land use planning applications. The changes needed are:

- #118 
- color the polygons to match a type:  preapp / outline / reserved matters / local plan
- Assuming we don't need linear routes or points, we can maybe simplify the toolbox.
- Change the form to instead have:
  - ATE reference type
  - application type (the thing we're coloring?)
  - Number of Units / Floorspace (GFA) / Area (ha)
  - ATE Triage / Prioritisation: No Comment / Standing Advice / Toolkit Assessment
  - ATE Recommendation: No Comment / No Objection / Standing Advice / Deferral / Approve subject to conditions and/or obligations / Refusal
- Depending on how this form looks, maybe this is also a forcing function to try a full-screen or larger modal for the form contents, instead of the sidebar.

How should we implement this? My strawman proposal is to have a separate `/planning_scheme.html` page glued to a variation of `App.svelte` (a copy or we parameterize it)? We should be able to directly plug in most components, or parameterize things to handle the current ATIP form or the other.

robinlovelace-ate:
Good shout. I think this deserves its own milestone as separate from the core Design functionality.

dabreegster:
Planning mode is live and has everything mentioned here, except for area snapping. Since there's an issue for that and no further info on what's needed here, closing

robinlovelace-ate:
Latest on this shown in quick demo below. 


https://user-images.githubusercontent.com/122299965/234585002-905b4c71-92da-4271-acbf-ca5e805c4c95.mp4

PR #140, opened 2023-04-24
Specialized planning mode form, toggled by a URL query param

dabreegster:
For #136. It's just normal ATIP, but with a different form, and only a polygon tool.

https://user-images.githubusercontent.com/1664407/234032078-f7ab007f-6eef-44a5-8181-8e92eb3a0104.mp4

UX thoughts:
- Should we have buttons to switch between regular ATIP mode and planning mode, or should planning folks know to go to a special URL? We should probably loudly change the style or label the page as planning mode, regardless.
- The about & instructions popups maybe need to change
- Is the form easy to use? Is it too small cramped in the sidebar; should we have a bigger modal pop up?

Implementation / code maintenance thoughts:
- The current approach is to introduce a `?planning` URL query param and toggle a bunch of behavior based on that. This was the easiest way to make this change, but is it the most maintainable?
- Should we instead make a separate `.html` with its own specialized copy of `App`? The immediate problem is that most components are the same; only a few need to specialize right now. Maybe we revisit as both ATIP modes evolve?
- Alternatively, do we bite the bullet and make everything in ATIP much more generic? Do we want to support multiple schemas or data entry forms?

### DIFF
--- a/src/lib/EntireScheme.svelte
+++ b/src/lib/EntireScheme.svelte
@@ -4,9 +4,15 @@
   import type { Scheme } from "../types";
 
   export let authorityName: string;
+  export let planningMode: boolean;
+
+  let baseFilename = authorityName;
+  if (planningMode) {
+    baseFilename += "_planning";
+  }
 
   // Set up local storage sync
-  let loadLocal = window.localStorage.getItem(authorityName);
+  let loadLocal = window.localStorage.getItem(baseFilename);
   if (loadLocal) {
     try {
       gjScheme.set(backfill(JSON.parse(loadLocal)));
@@ -19,7 +25,7 @@
     if ($gjScheme) {
       console.log(`GJ changed, saving to local storage`);
       window.localStorage.setItem(
-        authorityName,
+        baseFilename,
         JSON.stringify(geojsonToSave())
       );
     }
@@ -51,7 +57,7 @@
 
   function exportToGeojson() {
     let geojson = geojsonToSave();
-    var filename = authorityName;
+    var filename = baseFilename;
     geojson.authority = authorityName;
     // we could probably be more sophisticated here and set version more centrally
     geojson.origin = "atip-v1";
--- a/src/lib/PlanningForm.svelte
+++ b/src/lib/PlanningForm.svelte
@@ -1,0 +1,116 @@
+<script lang="ts">
+  import { gjScheme, deleteIntervention, formOpen } from "../stores";
+  import type { InterventionProps, PlanningProps } from "../types";
+
+  export let id: number;
+  export let allProperties: InterventionProps;
+
+  // TODO Should we start with these defaults, or make everything optional?
+  allProperties.planning ||= {
+    name: "",
+    notes: "",
+    reference_type: "preapp",
+    size: 0,
+    size_units: "number of units",
+    triage: "No Comment",
+    recommendation: "No Comment",
+  };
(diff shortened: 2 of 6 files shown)


src/lib/EntireScheme.svelte
```diff
@@ -4,9 +4,15 @@
   import type { Scheme } from "../types";
 
   export let authorityName: string;
+  export let planningMode: boolean;
+
+  let baseFilename = authorityName;
```
dabreegster, line 9: Keep the two auto-saved files separate

src/lib/PlanningForm.svelte
```diff
@@ -0,0 +1,116 @@
...
+    "outline",
+    "reserved matters",
+    "local plan",
+  ];
+  const sizeUnits = [
+    "number of units",
+    "floorspace (Gross Floor Area)",
+    "area (Hectare)",
+  ];
+  const triages = ["No Comment", "Standing Advice", "Toolkit Assessment"];
```
dabreegster, line 31: There's repetition here with `types.ts`. It's not onerous to keep in sync right now, and TS should help us with accidental typos or mismatches. But maybe there's a way to do TS build-time reflection and get all string cases of an enum?

dabreegster (commented):
Go try it: https://acteng.github.io/atip/scheme.html?planning&authority=Derby

robinlovelace-ate:
I'm finally taking a look at this...

robinlovelace-ate:
In terms of UX...

> 
> * Should we have buttons to switch between regular ATIP mode and planning mode, or should planning folks know to go to a special URL? We should probably loudly change the style or label the page as planning mode, regardless.

Sounds reasonable and will not affect the core 'Design' mode

> * The about & instructions popups maybe need to change
> * Is the form easy to use? Is it too small cramped in the sidebar; should we have a bigger modal pop up?

No comment...

> 
> Implementation / code maintenance thoughts:
> 
> * The current approach is to introduce a `?planning` URL query param and toggle a bunch of behavior based on that. This was the easiest way to make this change, but is it the most maintainable?

Query param sounds fine.

> * Should we instead make a separate `.html` with its own specialized copy of `App`? The immediate problem is that most components are the same; only a few need to specialize right now. Maybe we revisit as both ATIP modes evolve?

Not sure.

> * Alternatively, do we bite the bullet and make everything in ATIP much more generic? Do we want to support multiple schemas or data entry forms?

Big question, and I don't have enough info to make an informed decision. Overall: looks great.

robinlovelace-ate (approved):
I'm not qualified to comment on tech but overall thumbs up on idea, assuming there are no additional costs/impacts.

dabreegster:
I added coloring by reference type:
Screenshot from 2023-04-26 13-01-28

and a way to switch between the two ATIP modes:
Screenshot from 2023-04-26 13-01-48

The code maintenance cost is not high right now. I'd still like to get @Sparrow0hawk's opinion on how we might structure this longer-term, but for now, I'm happy enough with it, so I'll merge.

src/lib/draw/InterventionLayer.svelte
```diff
@@ -70,6 +72,20 @@
       colors.other,
       "white",
     ];
+  // For planning mode
+  const colorByReferenceType: DataDrivenPropertyValueSpecification<string> = [
```
dabreegster, line 76: There's boilerplate here and above; we can refactor if we do this kind of thing often

Issue #118, opened 2023-04-11
Polygon snapping?


robinlovelace-ate:
There are three main options:

option 1: broadly, like the existing route snapping, but as soon as you have 3 points, it connects the first and last. You can no longer click normally to add a totally new point, just drag existing points to make new waypoints

option 2: like route snapper, but some button to say "close it off and make an area". then you can only drag to adjust

option 3: make it more like the current polygon tool, where a click will insert a vertex in the nearest line segment. like that, but try to snap

Which to go for? Comments below welcome (requires signing up for a GitHub account).

Previous outdated comment below.

<details>

~~Worth implementing? I know it would be of use to the spatial planning team although they may be looking for a non OSM source of snap points. It could be handy when planning traffic management areas but not sure. So pitching as a question.~~

dabreegster:
Prototype at https://github.com/dabreegster/route_snapper/tree/cordon. We can feed it any line-strings we want, if there's a desire for non-OSM based

robinlovelace-ate:
Ace. I tentatively suggest this is part of the v2 road map.

robinlovelace-ate:
Have set it to v2. We can always revisit.

dabreegster:
This is working, with no urgent next steps known. https://github.com/dabreegster/route_snapper/issues/27 has improvements

PR #143, opened 2023-04-27
New polygon snap tool

dabreegster:
#118

https://user-images.githubusercontent.com/1664407/234880094-2674b4cd-ba4e-4d7c-a2a2-60834e62226a.mp4

For now, the new tool is distinct from the previous freehand polygon tool. In the future, we could try to merge the two, allowing easy freehand points alongside snapping. But for now, it's too tough.

As a note, this was a super quick change to crank out, because TS had my back. I had one or two bugs (forgetting to glue something up) and the compiler told me. I'm very happy with the decision to use it.

### DIFF
--- a/src/lib/draw/route/RouteControls.svelte
+++ b/src/lib/draw/route/RouteControls.svelte
@@ -6,7 +6,7 @@
 
   // TODO Save this in the route
   let avoidDoublingBack = false;
-  $: routeTool.setConfig({
+  $: routeTool.setRouteConfig({
     avoid_doubling_back: avoidDoublingBack,
   });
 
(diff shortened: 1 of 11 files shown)


src/lib/draw/route/RouteControls.svelte
```diff
@@ -6,7 +6,7 @@
 
   // TODO Save this in the route
   let avoidDoublingBack = false;
-  $: routeTool.setConfig({
+  $: routeTool.setRouteConfig({
```
dabreegster, line 9: Sliiight bug here. If we make a snapped area, then make a snapped route, avoid doubling back will be true from the last call, but the checkbox won't be ticked. I think we need to add a property to the output geojson feature to indicate this option and keep it synced.

PR #145, opened 2023-05-03
Start a v2 route schema using auto-generation

dabreegster:
Here's the new schema in action:

https://user-images.githubusercontent.com/1664407/235950184-1842f930-2ae5-4251-b125-8899658979ee.mp4

For this example, the generated JSON looks like:

```
      "properties": {
        "intervention_type": "other",
        "v2": {
          "name": "heya",
          "RouteType": {
            "OnRoad": {
              "Type": {
                "NoSeparation": "ModalFilters"
              }
            }
          },
          "Users": {
            "SharedUseWithSeparation": {
              "width_footpath": 9,
              "width_cyclepath": 9
            }
          }
        }
      },
```

## Design decisions

First, the easy one -- in #140 we added the second schema and I wondered how to support this long term. This PR clearly moves in the direction of ATIP supporting multiple schemas long-term, not just as some kind of temporary hack.

I started by manually writing the Svelte component to gather all of the properties, like how `PlanningForm.svelte` looks right now. But the boilerplate quickly becomes overwhelming -- especially because the form becomes complicated with nested enum cases. (Example: when someone says the user of the route is shared with separation, we need to ask them additional width questions that only make sense for that case.)

So I decided to express the schema in JSON. I didn't use the standard [JSON schema](https://json-schema.org/) language and tooling, because

1) All the existing libraries for auto-generating a form didn't work well and don't seem easy to fix up for our purposes. See #99 for some notes
2) It's a complicated language and is kind hard to read the results, especially when all of the nested types are lifted/flattened into one definitions list.

The bespoke schema definition magic is quite simple -- the form autogeneration is 85 lines of Svelte, and the TS type generator is 55 lines. I don't know if investing in something standard is a better choice longer-term, but for the moment, I can iterate much faster with this full control.

## UX design

The form is pretty minimally styled right now. From Robin's early feedback, I added some transitions when a sub-form appears for some enum cases. Ideas or examples of similar forms welcome.

Another reason to play with our own schema definition format is that I'd like, in the relatively short term, to add a bunch of metadata about all of these enum cases. For `OnRoad` types, displaying a reference picture, description, approximate cost, "when should you use this?", etc in some kind of tooltip would be awesome. There's room in the current format to add that kind of stuff in however best suits us.

## TODO

I'm going to merge this ASAP but async feedback welcome as usual. Some next steps:

- [ ] Use TS for the new auto-generation stuff. Fix the `| string` hack for `NumberInput`
- [ ] Convert the planning and v1 forms to use this JSON magic
- [ ] Write more of the v2 schema -- not just routes!
- [ ] Improve form UX -- maybe the sidebar needs to be bigger, or we need modals?


src/lib/forms/AutogenerateForm.svelte
```diff
@@ -0,0 +1,85 @@
...
+  import type { Field } from "./types";
+
+  // This component creates a form to collect the property described by spec, and put the result in value
+  export let spec: Field;
+  export let value;
+
+  // Blank out initial values if needed
+  let oneOfCase = "";
+  if (typeof spec == "string") {
+  } else if ("members" in spec) {
```
dabreegster, line 12: These're so much less clear than Rust's `if let` or `match`. I think I'll switch to https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

src/lib/forms/AutogenerateForm.svelte
```diff
@@ -0,0 +1,85 @@
...
+    value = oneOfCase;
+  }
+  function otherOneOf() {
+    value = {};
+    value[oneOfCase] = {};
+  }
+</script>
+
+{#if typeof spec == "string"}
+  <h1>Error: Field component instantiated for plain string, why? {spec}</h1>
```
dabreegster, line 36: This should be unreachable

src/lib/forms/AutogenerateForm.svelte
```diff
@@ -0,0 +1,85 @@
...
+{:else if "members" in spec}
+  {#each spec.members as x}
+    <div>
+      <h3>{x.name}</h3>
+      <svelte:self spec={x} bind:value={value[x.name]} />
```
dabreegster, line 41: This is how recursion works: https://svelte.dev/tutorial/svelte-self

src/lib/forms/AutogenerateForm.svelte
```diff
@@ -0,0 +1,85 @@
...
+        <input
+          type="radio"
+          bind:group={oneOfCase}
+          on:change={otherOneOf}
+          value={x.name}
+        />
+        {x.name}<br />
+      </label>
+
+      {#if oneOfCase == x.name && typeof value == "object"}
```
dabreegster, line 67: The second part is necessary to make sure the `otherOneOf` onchange handler has a chance to run and set up the value as a dictionary if needed. Otherwise when you switch an enum from a string to a struct case, it breaks

src/lib/forms/FormV2.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
```
dabreegster, line 1: No logic here at all -- it just uses the autogenerated form for the schema

src/schemas/v2_route.json
```diff
@@ -0,0 +1,67 @@
+{
```
dabreegster, line 1: This is the thing meant to be edited by a human. @robinlovelace-ate , without me really describing the schema language somewhere, is this intuitive enough to figure out how to write? To me this is much more clear than all the equivalent JSON Schema examples I could come up with

src/App.svelte
```diff
@@ -92,27 +87,28 @@
     </div>
     <br />
     <label>
-      ATIP mode:
-      <select bind:value={mode} on:change={changeMode}>
-        <option value="design">Design</option>
+      ATIP schema:
+      <select bind:value={schema} on:change={changeSchema}>
+        <option value="v1">Design (v1)</option>
```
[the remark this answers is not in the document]
robinlovelace-ate: "Infrastructure taxonomy"?

src/schemas/v2_route.json
robinlovelace-ate: Taking a look...

robinlovelace-ate:
Liking the backwards compatibility: 
image

robinlovelace-ate:
In terms of naming, I'm erring back towards schema:

> a representation of a plan or theory in the form of an outline or model

PR #146, opened 2023-05-03
Update App.svelte: Design -> Schema

robinlovelace-ate:
As per comments here https://github.com/acteng/atip/pull/145#issuecomment-1533573887 I think schema is a more useful word than Design here, even though it does affect the UI design.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -89,9 +89,9 @@
     <label>
       ATIP schema:
       <select bind:value={schema} on:change={changeSchema}>
-        <option value="v1">Design (v1)</option>
-        <option value="v2">Design (v2)</option>
-        <option value="planning">Planning</option>
+        <option value="v1">Schema (v1)</option>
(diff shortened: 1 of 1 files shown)


src/App.svelte
```diff
@@ -89,9 +89,9 @@
     <label>
       ATIP schema:
```
dabreegster, line 90: We already label all of these things as schemas. So "ATIP Schema: Schema (housing)" feels redundant -- why not just "ATIP Schema: housing"?

robinlovelace-ate: > We already label all of these things as schemas. So "ATIP Schema: Schema (housing)" feels redundant -- why not just "ATIP Schema: housing"?

Oh yeah, good point.

robinlovelace-ate: Updated.

dabreegster (approved):
LGTM, thanks!

PR #147, opened 2023-05-03
Update v2_route.json

robinlovelace-ate:
This makes the human readable text more consistent (Capital first letter in field heading) and clear.

### DIFF
--- a/src/schemas/v2_route.json
+++ b/src/schemas/v2_route.json
@@ -2,7 +2,7 @@
   "name": "Route",
   "members": [
     {
-      "name": "name",
+      "name": "Intervention name",
       "type": "one-liner"
     },
     {


src/schemas/v2_route.json
```diff
@@ -2,7 +2,7 @@
   "name": "Route",
   "members": [
     {
-      "name": "name",
+      "name": "Intervention name",
```
dabreegster, line 5: Right now these are used both as keys in the resulting JSON and in the form description. Maybe it's time to distinguish those two things already. For example, below, `FullSeparation` could become something like `{"value": "FullSeparation", "short_name": "Full separation", "description": "Solid raised kerb at least 1m in width separating the path from the carriageway. Eg Blackfriars road. Minimum width X meters. Should be used when possible. Costs about $123 per km"

robinlovelace-ate: +1 to `value`, `short_name` and `description` keys in the JSON schema spec.

Issue #134, opened 2023-04-23
Nicer toolbox design


dabreegster:
Currently we're here:
![233829877-5b7a01b5-865b-446d-b309-6ca7cbab8300](https://user-images.githubusercontent.com/1664407/233853284-19850d7f-6dce-4fd5-b9f8-d6a782a8ffe7.png)
@robinlovelace-ate suggested scaling everything down to take less space, for starters.

Other possible ideas for improving the design:

- [x] Icons for each tool
- [ ] Using consistent styling to indicate keybindings and what mouse action to take
- [x] A way to toggle off all the hints and keep things small, once someone knows it

Any other ideas?

dabreegster:
Fiddling around with size/color is easy: edit https://github.com/acteng/atip/blob/f5cb5d9a5a6e1c4b66ea4e607b780ab67b468c7e/src/lib/draw/Toolbox.svelte#L147 and other bits in the same file. `npm run dev` will auto-reload and show the changes

robinlovelace-ate:
Great stuff and agreed: the sizing of the toolbox deserves to be in the wider context of the UI design. Size is definitely an element, there are more, especially given the ambition to become GDS compliant. Looking forward to having a fiddle with `Toolbox.svelte` and seeing the results of others trying things out. Will mark as a good first issue as this is probably a good way into the ATIP codebase and as a 'help wanted' issue because UX/UI ideas are v. welcome.

robinlovelace-ate:
An additional issue with the current design: clashing buttons.

![image](https://user-images.githubusercontent.com/122299965/236023964-a99aafb6-f43b-44ab-b827-ab3b7d51dc98.png)

robinlovelace-ate:
Reproducible example: https://acteng.github.io/atip/scheme.html?authority=Isle+of+Wight&schema=v2#10.47/50.6717/-1.3282

dabreegster:
Right, and this problem is only going to get worse as we add more controls. Maybe all the instructions should be collapsible, as a start...

PR #151, opened 2023-05-04
Let long drawing control instructions be hidden. #134

dabreegster:
https://user-images.githubusercontent.com/1664407/236215962-b81e0f49-40e3-481d-a6fd-cde5f385a350.mp4

This is a small improvement to the very long instructions eating up all the toolbox space. Plenty of room for improvement. Maybe whether the instructions are shown should be remembered globally or per control?

### DIFF
--- a/src/lib/draw/CollapsibleCard.svelte
+++ b/src/lib/draw/CollapsibleCard.svelte
@@ -1,0 +1,54 @@
+<script lang="ts">
+  // TODO Lots of overlap with AccordionItem
+  import { slide } from "svelte/transition";
+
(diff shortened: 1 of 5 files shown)


src/lib/draw/CollapsibleCard.svelte
```diff
@@ -0,0 +1,54 @@
+<script lang="ts">
+  // TODO Lots of overlap with AccordionItem
```
dabreegster, line 2: Mostly just copied. We could refactor, or pick a better component library (possibly govuk based)

Issue #137, opened 2023-04-24
Add dedicated crossings identification functionality


robinlovelace-ate:
As outlined by @chris-ate this could be a priority. 

Reasoning:

- Crossings represent a major pinch point for active travel, especially for walking, and especially affecting less confident people e.g. school pupils
- The majority of walking and cycling crashes happen at junctions in many places

Thoughts on user story

- [ ] When sketching a new route it would be interesting to see the minimum number of times people have to cross a junction along the route
- [ ] Crossings could pop-up automatically or maybe less-intrusively after the user opts in
- [ ] Visual differentiation between crossing types, e.g. different colours, could help give an indication of severity of the crossing, with highway type being a reasonable starting point

In terms of how that could look, imagine sketching the top path in the route below, it would be useful to flag that there are many crossings, meaning many opportunities for left-hook crashes.

![image](https://user-images.githubusercontent.com/122299965/236026597-66c9eb11-b7f1-47ed-ad05-4388f40c3848.png)

Source: https://www.openstreetmap.org/relation/6386965#map=17/53.80788/-1.64609

dabreegster:
Any more context here?

robinlovelace-ate:
> Any more context here?

Just added 👍

dabreegster:
Thanks! This sounds like a good thing to tack onto the wishlist in #69. That issue has ideas and a prototype for how to implement it. We have to make some UX and data storage choices that're a bit interesting... for example, after we draw a route (which is recorded in the GeoJSON file) and we calculate crossings with a bunch of metadata, should we save that in the geojson file too? Or recalculate it later, treating it like a contextual layer?

dabreegster:
Dupe of #480

Issue #69, opened 2023-02-07
Auto-generate road name and speed limits for a drawn route


dabreegster:
As you draw a snapped route, we can give some info about it to help people decide its quality over possible alternatives.

- the amenities along it (broken down by categories -- groceries, health facilities, schools, etc)
- a measure of "directness" / number of turns / mental complexity to follow it (see https://github.com/a-b-street/abstreet/issues/1013#issuecomment-1287138031)

There are design tradeoffs to nudging people onto quieter back-streets vs past the local high street. We don't need to be opinionated about them, but we could just start exposing the information and see how people might use it.

dabreegster:
Another request: how many "critical" junctions crossed

dabreegster:
Reviving this. Further ideas from Kit and others:

- Amenities: high streets, rail/bus stations, childcare, medical facilities, schools, etc
- Speed limit along the route
- Use the route endpoints to automatically fill out a name for the route -- "Xyz St to Abc St". Many people in submitted ATIPs did this manually!
- Current lane allocation and/or width
- Census stats within a nearby walkshed

I'd like to start discussing how we might implement this.

# Strawman implementation idea

The A/B Street map model ([outdated docs](https://a-b-street.github.io/docs/tech/map/index.html), [code](https://github.com/a-b-street/abstreet/tree/main/map_model/src)) has much of the required data already. Or for things missing or inaccurate (like the current classification of OpenStreetMap's amenity tagging into higher-level categories), fixes that fit ATIP's purposes would be very much welcome there too.

I can think of a few approaches here:

1) Bake more detailed data into the file used by https://github.com/dabreegster/route_snapper/.
2) Load a serialized map file in the browser and interact with the contents using a WASM/JS API. We'd feed in a drawn LineString (or an area's polygon or anything else), do basic map matching to relate it to the roads defined in the map model, and return any details we need.
3) Create a read-only API that does the same, but host it remotely and make API calls over the network.

(1) and (2) are similar. Both would mean shipping and deserializing a larger file to the browser. For the large authority boundaries we've got, this could impact initial load time. (Which is unfortunate -- if we were shipping people a desktop app that they download/update only occasionally, that cost is amortized, and we have the full power of threads, etc....) I could estimate how much, if it's helpful.

[...]

PR #154, opened 2023-05-09
Create a new RouteInfo service, and use it to auto-fill the name of a route

dabreegster:
This is a big step towards #69, #137, and all other ideas where we start to understand routes drawn in ATIP not just as geospatial line-strings, but as deeply rich semantic structures. First, the demo. After drawing a route in the v1 mode, you can press a new Autofill button, and generate a name for the route based on street names:

https://github.com/acteng/atip/assets/1664407/98b5c747-27ea-456c-a91d-f63ab89562dd

# Technical design decisions

We want to understand and visualize many things about the route someone draws and things nearby:

- speed limits along the sections on carriageways
- crossings
- existing lane configuration, to help with "will-it-fit" checks, determining if part of a route shares a bus lane, etc
- what buildings and other amenities exist along the route
- possible turning conflicts at big junctions

I'm proposing two design decisions to do that kind of thing:

1) Don't attempt to do this analysis in JS or TS, or on top of "raw" OpenStreetMap data. We could try to express some of this as Overpass queries and post-process data coming back. I think that's doable for some of the simpler questions we have, but will quickly break down in complexity.

2) Use A/B Street's map model as a starter. Years of development and polish have gone into the map importer there, handling various problems in OSM. The map model API has been re-shaped over time to handle traffic simulations, walkshed analyses, low-traffic neighbourhood analysis, etc. I don't think it'll do everything we need in its current form (crossings as they're tagged in OSM are a big gap right now), but I'd push for investing in more improvements there, rather than starting something from scratch.

As an alternative to this approach, we could consider doing all of the route analysis "offline" -- read in the ATIP GeoJSON files, write R or Python scripts to calculate some of the things listed above. This would limit interactivity in ATIP -- maybe someone could later load the results of that and match things up, but the method started in this PR gives instant results. The results of the analysis will require us to develop custom UIs to display things like turn conflicts, and I think the Svelte components are the right place to do that frontend work. We can treat the Rust layer as a backend and keep the two pieces coupled.

## Future-proofing

Because initially deserializing the A/B Street map file is slow, running using web workers is immediately desirable, because it happens off the main thread. I found https://github.com/GoogleChromeLabs/comlink, which makes dealing with web workers very easy. This gives us another advantage as well -- the Svelte layer has to deal with async queries upfront. This means we could just as easily make HTTP requests to some new remote API somewhere to answer the same questions. So if we wanted, we could implement the `RouteInfo` service with a Python server running somewhere, and use whatever backend methods/data we want.

## File management

Edit: https://github.com/actenglabs/abstreet-to-atip/pull/7 imports all of the maps for the UK. They're already in S3, aside from 4 areas with bugs.


route_info/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+#[wasm_bindgen]
+impl RouteInfo {
+    #[wasm_bindgen(constructor)]
+    pub fn new(map_bytes: &[u8]) -> Result<RouteInfo, JsValue> {
+        // Panics shouldn't happen, but if they do, console.log them.
+        console_error_panic_hook::set_once();
+
+        web_sys::console::log_1(&format!("Got {} bytes, deserializing", map_bytes.len()).into());
+
+        let map: Map = bincode::deserialize(map_bytes).map_err(err_to_js)?;
```
dabreegster, line 20: So after we pass in a binary map file (produced in abstreet-to-atip and soon hosted in S3), we deserialize it. Then we also build a quadtree to snap to an intersection, since all the queries will make use of that

route_info/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+            closest_intersection.add_polygon(i.id, &i.polygon);
+        }
+
+        Ok(RouteInfo {
+            map,
+            closest_intersection,
+        })
+    }
+
+    // TODO How can we glue up anyhow?
```
dabreegster, line 33: These TODOs are to improve the code; they're not bugs or feature critical

route_info/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+    // TODO Can we take geojson::Feature so that it shows up in TS?
+    #[wasm_bindgen(js_name = nameForRoute)]
+    pub fn name_for_route(&self, raw_line_string: JsValue) -> Result<String, JsValue> {
+        let feature: geojson::Feature = serde_wasm_bindgen::from_value(raw_line_string)?;
+        // TODO Duplicate non-adjacent points; make a more permissive version of PolyLine.
+        //let polyline = PolyLine::from_geojson(&feature, Some(self.map.get_gps_bounds())).map_err(err_to_js)?;
```
dabreegster, line 39: A/B Street's geometry library has methods to much more easily parse GeoJSON, but they're too strict about duplicate points. We could improve this later to simplify code here.

route_info/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+            let pt2 = LonLat::new(last[0], last[1]);
+            (
+                pt1.to_pt(self.map.get_gps_bounds()),
+                pt2.to_pt(self.map.get_gps_bounds()),
+            )
+        } else {
+            return Err(err_to_js("couldn't get line-string"));
+        };
+
+        // Find the closest intersections to the route endpoints
```
dabreegster, line 56: The route snapper just gives raw GPS coordinates. Here we do a super simple version of [map matching](https://en.wikipedia.org/wiki/Map_matching) to find the closest intersection in the street network to each of the route's raw endpoints.

route_info/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+        let lang = None;
+        let (i1, _) = self
+            .closest_intersection
+            .closest_pt(pt1, threshold)
+            .ok_or_else(|| err_to_js("no intersection close to first pt"))?;
+        let (i2, _) = self
+            .closest_intersection
+            .closest_pt(pt2, threshold)
+            .ok_or_else(|| err_to_js("no intersection close to last pt"))?;
+        let i1_name = self.map.get_i(i1).name(lang, &self.map);
```
dabreegster, line 67: Then over in A/B Street, there's an existing method to print a nice name for an intersection, based on road names: https://github.com/a-b-street/abstreet/blob/dc74f7de5455b122c5636b34eb0d83b80116a20b/map_model/src/objects/intersection.rs#L229

(This can even use alternate languages, when a street is tagged with multiple names!)

src/App.svelte
```diff
@@ -54,6 +60,25 @@
...
+
+    // If you get "import declarations may only appear at top level of a
+    // module", then you need a newer browser.
+    // https://caniuse.com/mdn-api_worker_worker_ecmascript_modules
+    //
+    // In Firefox 112, go to about:config and enable dom.workers.modules.enabled
+    //
+    // Note this should work fine in older browsers when doing 'npm run build'.
+    // It's only a problem during local dev mode.
+    interface WorkerConstructor {
```
dabreegster, line 72: This is a small amount of setup / TS boilerplate to make Comlink wrap our class

src/App.svelte
```diff
@@ -3,6 +3,10 @@
   import { onMount } from "svelte";
   import authoritiesUrl from "../assets/authorities.geojson?url";
   import type { Schema } from "./types";
+  import bristolUrl from "../assets/bristol.bin?url";
+  import * as Comlink from "comlink";
+  import workerWrapper from "./worker?worker";
```
dabreegster, line 8: The `?worker` syntax is Vite build magic... which only works in some browsers when doing `npm run dev`. See comments below

dabreegster (commented):
I'm working on generating the map data files everywhere, will followup soon.

Two UX questions immediately:

1) Should we have a second progress bar somewhere to show the second file loading?
2) Where should this auto-fill button go? It's just in the old v1 form for now. It's only relevant for routes, so maybe also in v2.

Sparrow0hawk:
Just glancing at this and will try give it a proper look through tomorrow. Quick question, is there a reason `route_info` has to live in this repo? Or could it live in its own repo like `route-snapper` and we just `npm install` it?

dabreegster:
> Or could it live in its own repo like route-snapper and we just npm install it?

No hard limit, but in the short-term, I think it's more convenient to keep it here. The local development workflow is more of a hassle when you have to edit a few local repos in sync, merge PRs in sync, etc. I've hit that annoyance already with route-snapper; https://github.com/dabreegster/route_snapper/blob/main/CHANGES.md is the amusing result. I feel like we're in the exploratory phase right now, and the API of the route info service will change constantly. I expect to write new Svelte components tightly coupled to the data it returns (like to display speed limit broken down along a route).

So I'd propose keeping it here for now, and splitting it out into a standalone repo once the dust has settled a bit.

robinlovelace-ate:
Regarding your proposals


> 1. Don't attempt to do this analysis in JS or TS, or on top of "raw" OpenStreetMap data. We could try to express some of this as Overpass queries and post-process data coming back. I think that's doable for some of the simpler questions we have, but will quickly break down in complexity.

Agreed. Seeing the routes as part of the network along which people travel makes sense and is the most future proof option. I see it as hard to build and cannot imagine building such functionality, hence had not thought of it, but fully approve of the thinking here if it's actually doable, which I believe is.

> 2. Use A/B Street's map model as a starter. Years of development and polish have gone into the map importer there, handling various problems in OSM. The map model API has been re-shaped over time to handle traffic simulations, walkshed analyses, low-traffic neighbourhood analysis, etc. I don't think it'll do everything we need in its current form (crossings as they're tagged in OSM are a big gap right now), but I'd push for investing in more improvements there, rather than starting something from scratch.

robinlovelace-ate:
> As an alternative to this approach, we could consider doing all of the route analysis "offline" -- read in the ATIP GeoJSON files, write R or Python scripts to calculate some of the things listed above. This would limit interactivity in ATIP -- maybe someone could later load the results of that and match things up, but the method started in this PR gives instant results. The results of the analysis will require us to develop custom UIs to display things like turn conflicts, and I think the Svelte components are the right place to do that frontend work. We can treat the Rust layer as a backend and keep the two pieces coupled.

If the business logic is written in Rust wouldn't that allow us to get best of both worlds, using the Rust code in the browser via Wasm and in the backend via a CLI like `cargo install --git https://github.com/dabreegster/odjitter` ?

robinlovelace-ate:
> Two UX questions immediately:
> 
> 1. Should we have a second progress bar somewhere to show the second file loading?

No strong feelings. Having 2 progress bars is not common and may look a bit complex but that's just a random thought without having seen it. As a sketch, where would the 2nd progress bar go compared with this?

image



> 2. Where should this auto-fill button go? It's just in the old v1 form for now. It's only relevant for routes, so maybe also in v2.

In the left hand panel just below each question would make sense to me on first glance.

src/App.svelte
robinlovelace-ate: This is magic to me ✨

robinlovelace-ate (approved):
I'm not qualified to comment on the technical details here. This looks like some hardcore modern TypeScript and Svelte code, combined with a nice bit of tightly integrated high-performance yet concise Rust of good measure. Based on the video and thinking expressed in the PR comments, which greatly helps understand the intention, it's a thumbs-up from me pending a few questions in my previous comments.

dabreegster:
Thanks for asking great, critical questions! This is a longer-term decision starting in this PR, so very nice to get some discussion before diving farther in.

> Isn't A/B Street's map model itself based on osm2streets now? That seems like a smaller, easier to update, and more agile codebase to build on.

Yes. For Alex's benefit, some context: about a year ago, we started splitting https://github.com/a-b-street/osm2streets out from the rest of A/B Street. That's made it easier to spin out smaller projects that just need to understand the street network part of OSM data. It's also a chance to more rapidly iterate on that code without having to immediately keep all A/B Street apps up-to-date. The decision to split has paid off very well (unlike the one here of ATIP and the route snapper). I'm trying to slowly move more and more of A/B Street logic into osm2streets, taking the chance to tackle it with fresh perspective and rewrite -- things like turns, representing edits to the network, pathfinding.

I started basing off A/B Street [map model](https://github.com/a-b-street/abstreet/tree/master/map_model/src/), which is a layer after osm2streets, partly based on convenience/rapid prototyping speed, and partly because the map model layer also has [buildings](https://github.com/a-b-street/abstreet/blob/main/map_model/src/objects/building.rs) snapped to roads. That could be helpful if we want to show amenities along a route.

But...

> Ease of maintenance and contribution?

I think this is a strong argument. The osms2streets codebase is comparatively much cleaner, better documented, better tested, and less burdened with older assumptions. Things like zebra and unmarked crossings are a total mess in the map model layer, with US-specific assumptions baked pretty deeply in and crossing data tagged in OSM ignored. We'll likely want to revamp that relatively soon to understand the pedestrian experience. osm2streets is a much better place to do that.

And for buildings, there are also some old hacks (like losing any association between an OSM relation of a university campus and the 100 individual buildings in it) that'd be helpful to improve for ATIP.

Additionally, storing files with just the osm2streets StreetNetwork will be smaller. Map model files have more stuff crammed in, and we don't need it all.

So, I propose:
- merge this PR as a first, already working step
- immediately followup by cutting over to osm2streets instead, and moving+rewriting building, crossing, turn, etc logic there as needed

> Isn't A/B Street's map model overkill for some things? A design goal for ATIP is simplicity and A/B Street's map model is quite complex I thought. As an example, all A/B Street routes are represented in 2D while ATIP routes are 1D.

osm2streets and A/B Street are the same here. Roads and routes are "either" 1D or 2D. Through the API, you can access just 1D road center-lines and glue them together into a longer route. Or you can use the road's estimated width and see more of a 2D perspective. We won't make use of that yet.

> If the business logic is written in Rust wouldn't that allow us to get best of both worlds

+1, also true. I would summarize it like this:

- Rust: run as an offline CLI tool, directly in the browser, or as a remote API server
- anything else: offline CLI tool or as a remote API server

> Having 2 progress bars is not common and may look a bit complex

Agreed, and once the tool is loaded, there's not even an immediately actionable thing for the user to do with it. So actually, how about by the autofill button nested in a route's form? If the RouteInfo API isn't ready to use yet, that button can be disabled with a tooltip saying it's loaded, or we just put the progress bar there. Most people won't even see it, unless they open a route before it finishes loading.

robinlovelace-ate:
> So, I propose:
> 
> * merge this PR as a first, already working step
> * immediately followup by cutting over to osm2streets instead, and moving+rewriting building, crossing, turn, etc logic there as needed

👍 to that.

robinlovelace-ate:
> Through the API, you can access just 1D road center-lines and glue them together into a longer route. Or you can use the road's estimated width and see more of a 2D perspective. We won't make use of that yet.

Very cool, I didn't realise that. The fact that a multi-lane road can simultaneously be represented as a 1D linestring and as a complex polygon is amazing. Great that we'll benefit from the ability to almost instantly switch between the two representations. Big 👍

dabreegster:
Just noting we'll see Github actions failures until this is merged and the other pending branches are rebased. I tried making the wasm-pack step be conditional on whether the `route_info` directory exists (and thus we're in a branch where this new code exists and needs building), but when switching between all branches in the build script, it leaves behind some files there and falsely tries to build. Seems simplest to me to push past this and make sure everything's fine after merging, instead of making the script more complicated for a short time.

Issue #157, opened 2023-05-10
Addition of optional free text Description field to Schema v2


robinlovelace-ate:
With v1 the two free text fields are Intervention name and Description:

![image](https://github.com/acteng/atip/assets/122299965/8696bcfc-c94f-463b-867e-76560ee27367)

Under v2 it's:

![image](https://github.com/acteng/atip/assets/122299965/5300e4cc-559b-4751-bd28-7a5b4e090681)

I think it's worth keeping an optional Description field in both for backwards compatibility.

PR #158, opened 2023-05-10
Add freeform description to v2 schema. #157

dabreegster:
Screenshot from 2023-05-10 10-58-01

### DIFF
--- a/src/schemas/v2.ts
+++ b/src/schemas/v2.ts
@@ -4,6 +4,7 @@ export type Intervention = Route | Crossing;
 
 export interface Crossing {
   name?: string;
+  description?: string;
   CrossingType?: CrossingType;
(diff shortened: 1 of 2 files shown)


src/schemas/v2.ts
```diff
@@ -4,6 +4,7 @@ export type Intervention = Route | Crossing;
 
 export interface Crossing {
   name?: string;
+  description?: string;
```
dabreegster, line 7: Reminder this is autogenerated. I edited the JSON, then `npm run generate-form-ts`

robinlovelace-ate (approved):
Big 👍. Wouldn't be an overstatement to say that I find it amazing and super cool that the UI can be adjusted in this way with only 10 lines of code changed in config file that everyone can see. Goodbye UI hacks with this approach!

Issue #162, opened 2023-05-12
Deleting a route gives console error and prevents further drawing


Sparrow0hawk:
Whilst testing most recent changes pushed to `main` I found I could trigger an unrecoverable bug that requires a refresh to fix.

## User interaction

I'm using Firefox 112.0.2 (64-bit) on macOS Ventura.

https://github.com/acteng/atip/assets/20887250/8600479c-44ac-43cb-921f-7b5e2dd49d93

## Steps to reproduce

You can reproduce this by:

1. Clicking New Route
2. Drawing a route and clicking Enter to stop drawing
3. Use the new autofill button to fill scheme name
4. Click Save
5. Click delete

This triggers a console error:

```
Uncaught (in promise) TypeError: h is undefined
```

And any subsequent attempts to draw or use a different intervention tool will allow you to draw but on clicking Enter do not add the intervention.

PR #163, opened 2023-05-12
Adding some tests for adding route intervention

Sparrow0hawk:
This PR adds some tests for adding route interventions and updated playwright version to get access to [`--ui`](https://playwright.dev/docs/test-ui-mode) mode.

These tests include one test that aims to capture the bug described in #162 and so I expect it to fail checks.

Test code was produced using codegen which is pretty nice.

### DIFF
--- a/tests/add-delete-add.spec.ts
+++ b/tests/add-delete-add.spec.ts
@@ -5,7 +5,7 @@ test('testing adding interventions, then deleting one, then adding another', asy
     test.fixme();
     await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
     // wait for the map to load and interventions panel to appear
-    await page.getByText('Edit attributes Click an intervention to fill out its attributes Edit geometry N').waitFor();
+await page.getByText('Click an intervention to fill out its attributes').waitFor();
     // wait for router snapper to load so we can use route tool
     await page.getByText('Route tool loading...').waitFor({ 'state': 'hidden' });
 
--- a/tests/add-delete-add.spec.ts
+++ b/tests/add-delete-add.spec.ts
@@ -1,0 +1,39 @@
+import { test, expect } from '@playwright/test';
+
+test('testing adding interventions, then deleting one, then adding another', async ({ page }) => {
+    await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
+    await page.getByRole('region', { name: 'Map' }).waitFor();
+    await page.getByRole('button', { name: 'New route' }).click();
+    await page.getByRole('region', { name: 'Map' }).click({
+        position: {
+            x: 522,
+            y: 468
+        }
+    });
+    await page.getByRole('region', { name: 'Map' }).click({
+        position: {
+            x: 192,
+            y: 513
+        }
+    });
+    await page.getByRole('button', { name: 'Finish' }).click();
+    await page.getByRole('button', { name: 'Save' }).click();
+    await page.getByRole('button', { name: '1) Untitled route' }).click();
+    await page.getByRole('button', { name: 'Delete' }).click();
+    await page.getByRole('button', { name: 'New route' }).click();
+    await page.getByRole('region', { name: 'Map' }).click({
+        position: {
+            x: 196,
+            y: 375
+        }
+    });
+    await page.getByRole('region', { name: 'Map' }).click({
+        position: {
+            x: 481,
+            y: 399
+        }
+    });
+    await page.getByRole('button', { name: 'Finish' }).click();
+
+    await expect(page.getByRole('button', { name: '1) Untitled point' }).isVisible())
+});
--- a/tests/add-delete-add.spec.ts
+++ b/tests/add-delete-add.spec.ts
@@ -2,6 +2,11 @@ import { test, expect } from '@playwright/test';
 
 test('testing adding interventions, then deleting one, then adding another', async ({ page }) => {
     await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
+    // wait for the map to load and interventions panel to appear
+    await page.getByText('Edit attributes Click an intervention to fill out its attributes Edit geometry N').waitFor();
+    // wait for router snapper to load so we can use route tool
+    await page.getByText('Route tool loading...').waitFor({ 'state': 'hidden' });
+
     await page.getByRole('region', { name: 'Map' }).waitFor();
     await page.getByRole('button', { name: 'New route' }).click();
     await page.getByRole('region', { name: 'Map' }).click({
--- a/tests/add-delete-add.spec.ts
+++ b/tests/add-delete-add.spec.ts
@@ -1,6 +1,8 @@
 import { test, expect } from '@playwright/test';
 
 test('testing adding interventions, then deleting one, then adding another', async ({ page }) => {
+    // mark this test to fail
+    test.fixme();
     await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
     // wait for the map to load and interventions panel to appear
     await page.getByText('Edit attributes Click an intervention to fill out its attributes Edit geometry N').waitFor();
--- a/tests/add-route.spec.ts
+++ b/tests/add-route.spec.ts
@@ -1,0 +1,23 @@
+import { test, expect } from '@playwright/test';
+
+test('testing add a route and save it', async ({ page }) => {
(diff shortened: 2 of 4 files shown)


tests/add-delete-add.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { test, expect } from '@playwright/test';
+
+test('testing adding interventions, then deleting one, then adding another', async ({ page }) => {
+    await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
+    await page.getByRole('region', { name: 'Map' }).waitFor();
```
Sparrow0hawk, line 5: Do an explicit `.waitFor()` here to make sure the map loads else I found it got itself in a pickle and timed out.

Sparrow0hawk:
Although that's it failing slightly before it should. This could potentially be because it's trying to do a step before something is visible (although I thought playwright handled this).

tests/add-route.spec.ts
```diff
@@ -2,7 +2,11 @@ import { test, expect } from '@playwright/test';
 
 test('testing add a route and save it', async ({ page }) => {
     await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
-    await page.getByRole('region', { name: 'Map' }).waitFor();
+    // wait for the map to load and interventions panel to appear
+    await page.getByText('Edit attributes Click an intervention to fill out its attributes Edit geometry N').waitFor();
```
Sparrow0hawk, line 6: Playwright doesn't know to wait for the map to load or for route snapper to finish loading. So we add some explicit waits in here for text elements that relate to those features.

I'm open to a potentially better way of waiting for these things (spent a non-trivial amount of time trying to work out how to wait for the console log message from router-snapper but failed), but as a first cut these seem to behave.

tests/add-route.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    });
+    await page.getByRole('region', { name: 'Map' }).click({
+        position: {
+            x: 192,
+            y: 513
+        }
+    });
+    await page.getByRole('button', { name: 'Finish' }).click();
+    // wait to make sure intervention attributes appear
+    await page.getByRole('button', { name: 'Save' }).waitFor();
```
dabreegster, line 25: I thought this was implied by https://playwright.dev/docs/actionability?

tests/add-route.spec.ts
dabreegster: I haven't tried it out yet, but I checked maplibre's own tests for inspiration: https://github.com/birkskyum/maplibre-gl-js/blob/c82f26ff6e3b1d4fb735db2c5b71d654d3319972/test/integration/browser/browser.test.ts#L51
They await a DOM element that maplibre inserts: `const canvas = await page.$('.maplibregl-canvas');`
We may be able to try something similar. Worth noting they just switched from Playwright to Puppeteer, but I think this particular technique didn't change: https://github.com/maplibre/maplibre-gl-js/pull/2494

In the meantime, I'm happy waiting for text like this as a proxy. We control it, so we could use some other way to mark that the route snapper is loaded. Maybe it slightly violates the principle of black-box testing, but do something blunt like `window.routeSnappedReady = true` and wait for that.

But I would say just trim down the text to not include the buttons? It reads a bit strangely in the test. How about just
```suggestion
    await page.getByText('Click an intervention to fill out its attributes').waitFor();
```

tests/add-delete-add.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+test('testing adding interventions, then deleting one, then adding another', async ({ page }) => {
+    await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
+    // wait for the map to load and interventions panel to appear
+    await page.getByText('Edit attributes Click an intervention to fill out its attributes Edit geometry N').waitFor();
+    // wait for router snapper to load so we can use route tool
+    await page.getByText('Route tool loading...').waitFor({ 'state': 'hidden' });
+
+    await page.getByRole('region', { name: 'Map' }).waitFor();
+    await page.getByRole('button', { name: 'New route' }).click();
```
dabreegster, line 11: Not necessary yet, but maybe as we have more of these, some simple helper functions would be nice, so we can express something like `createRoute();  await page.getByRole('button', { name: '1) Untitled route' }).click(); await page.getByRole('button', { name: 'Delete' }).click(); createRoute()`. And maybe something to wait for the map + route tool to be ready too

tests/add-delete-add.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            y: 375
+        }
+    });
+    await page.getByRole('region', { name: 'Map' }).click({
+        position: {
+            x: 481,
+            y: 399
+        }
+    });
+    await page.getByRole('button', { name: 'Finish' }).click();
```
dabreegster, line 41: I'm getting the timeout here -- same for you?

dabreegster (approved):
This is awesome, thanks! And the new playwright `--ui` is so convenient, surprised it's so new.

We could merge this new test in now if you like. https://playwright.dev/docs/api/class-test#test-fail-1 lets us mark that the test is expected to fail because the implementation is wrong. Then when we fix the bug, we should just have to remove this bit, and all should be well

tests/add-route.spec.ts
dabreegster: Started https://osmus.slack.com/archives/C01G3D28DAB/p1684002466097189 to ask about maplibre+playwright

tests/add-route.spec.ts
Sparrow0hawk: Me too but I found it would time out at this step. Now this could resolved by adding the waits for the map and route-snapper I just haven't tested it.

tests/add-delete-add.spec.ts
Sparrow0hawk: Yes apologies this should timeout because of issue in #162

tests/add-delete-add.spec.ts
Sparrow0hawk: Yep, that would definitely be nice, most of the code here is generated from a codegen session but definitely up for abstracting this into helper functions.

tests/add-route.spec.ts
dabreegster: A quick reply from Slack! We could do this:

https://github.com/birkskyum/maplibre-gl-js/blob/c82f26ff6e3b1d4fb735db2c5b71d654d3319972/test/integration/browser/browser.test.ts#LL32C1-L50C1

Sparrow0hawk:
> We could merge this new test in now if you like. https://playwright.dev/docs/api/class-test#test-fail-1 lets us mark that the test is expected to fail because the implementation is wrong. Then when we fix the bug, we should just have to remove this bit, and all should be well

I'm up for merging this went with [`test.fixme()`](https://playwright.dev/docs/api/class-test#test-fixme-2) as seems appropriate in this case.

tests/add-route.spec.ts
dabreegster: ... figuring out how to get ahold of `map`. Maybe something like `window.map = ...` in `Map.svelte` isn't the worst hack in the world

Sparrow0hawk: > But I would say just trim down the text to not include the buttons? It reads a bit strangely in the test. How about just

This is code generated from codegen, I'll give it a try with your suggestion locally and see if I can include.

dabreegster:
> I'm up for merging this went with [test.fixme()](https://playwright.dev/docs/api/class-test#test-fixme-2) as seems appropriate in this case.

Feel free to merge! We can continue to iterate on how to wait for stuff and refactoring helpers, as we write more of these tests. Thanks for adding these

tests/add-delete-add.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { test, expect } from '@playwright/test';
+
+test('testing adding interventions, then deleting one, then adding another', async ({ page }) => {
+    // mark this test to fail
+    test.fixme();
+    await page.goto('/scheme.html?authority=Derby#16.84/52.906457/-1.504519');
+    // wait for the map to load and interventions panel to appear
+    await page.getByText('Edit attributes Click an intervention to fill out its attributes Edit geometry N').waitFor();
```
Sparrow0hawk, line 8:
```suggestion
await page.getByText('Click an intervention to fill out its attributes').waitFor();
```

tests/add-route.spec.ts
Sparrow0hawk: Migrating the ideas about the map here to an issue to address in a new PR

PR #165, opened 2023-05-14
Reset openFromSidebar store, fixing #162

dabreegster:


src/stores.ts
```diff
@@ -49,11 +49,13 @@ export function newFeatureId(gj: Scheme): number {
...
 export function deleteIntervention(id: number) {
+  console.log(`Deleting intervention ${id}`);
   gjScheme.update((gj) => {
     gj.features = gj.features.filter((f) => f.id != id);
     return gj;
   });
   formOpen.set(null);
   mapHover.set(null);
   sidebarHover.set(null);
+  openFromSidebar.set(null);
```
dabreegster, line 60: Why is this necessary? https://github.com/acteng/atip/blob/b1b384908390f21b976777fed5e882c8caa80c76/src/lib/draw/AttributeMode.svelte#L37 is the place `openFromSidebar` gets used. The intention here is to only rerun this reactive block when that store changes. But Svelte doesn't work by specifying the variables to watch; when anything in the reactive statement changes, the whole thing reruns. So what's happening here is that `gjScheme` gets updated, deleting the feature. `openFromSidebar` is still set to something, and so looking up the feature by ID fails. We fix by remembering to clear this other store too -- aka, clean up all references to the deleted object.

tests/add-delete-add.spec.ts
```diff
@@ -46,7 +44,5 @@ test("testing adding interventions, then deleting one, then adding another", asy
   });
   await page.getByRole("button", { name: "Finish" }).click();
 
-  await expect(
-    page.getByRole("button", { name: "1) Untitled point" }).isVisible()
-  );
+  await page.getByRole("button", { name: "1) Untitled route" }).isVisible();
```
dabreegster, line 47: I was getting `Error: expect: Property 'then' not found.` We don't need this extra `expect`.

Sparrow0hawk (approved):
Good catch!

This resolved my issues when testing locally and the add-delete-add test now passes 💪

Issue #102, opened 2023-03-28
UX problems with routes


dabreegster:
The way we draw currently, you can't see the basemap underneath drawn routes. So the road names aren't visible! https://maplibre.org/maplibre-gl-js-docs/example/geojson-layer-in-stack/ could help.
![Screenshot from 2023-03-28 13-27-18](https://user-images.githubusercontent.com/1664407/228235893-ee717982-c956-4161-ade3-fc5e10ae43a6.png)

And when two separate routes meet at a point, the boundary between them is unknowable until you click one and see waypoints:
![Screenshot from 2023-03-28 13-28-14](https://user-images.githubusercontent.com/1664407/228236097-01cba560-205f-45ac-b652-2b7651da0dc8.png)

dabreegster:
For the second, maybe just adding circular end caps with a faint outline would help. I tried in Felt and geojson.io, and both have the same problem right now

dabreegster:
Both issues fixed. We still saw some people in the UX study get confused after successfully using the split route tool; the circle outline wasn't clear enough. But in lieu of better design ideas, marking done for v2

PR #166, opened 2023-05-15
Centralize layer z-ordering.

dabreegster:
Before, when we had points, lines, and polygons overlapping, the z-ordering was not great -- you couldn't really select the smaller object:
Screenshot from 2023-05-15 13-35-34

This PR defines an ordering for all of the layers in ATIP, so we can explicitly control everything from one place. The behavior is nicer now:

https://github.com/acteng/atip/assets/1664407/16d3c795-5509-45da-8953-89b157c449fa

As background for why getting z-ordering between different layers in MapLibre is tough, see https://github.com/mapbox/mapbox-gl-js/issues/7016.

# Next steps

I originally wanted to also make sure that we don't cover up road names: #102. We now draw beneath the road labels, but because they're faint and black in the default style, it's... still pretty useless against the blue routes:
Screenshot from 2023-05-15 13-38-52
If anyone has color or symbolizing suggestions either for the colors we're using, or to change the road labels, it's easy to try things out!


src/maplibre_helpers.ts
```diff
@@ -86,20 +84,39 @@ export function overwriteSource(
...
 }
 
-export function overwriteLayer(
-  map: Map,
-  layer,
-  beforeId: string | undefined = undefined
-) {
+// The layer.id here MUST be present in layerZorder.
+export function overwriteLayer(map: Map, layer) {
   if (map.getLayer(layer.id)) {
```
dabreegster, line 89: Starting a slightly orthogonal conversation here: using these two `overwrite` methods is kind of gross; see https://github.com/acteng/atip#hmr-best-practices for why. I see people using Svelte components to specify GeoJSON sources and layers in https://svelte-maplibre.vercel.app/examples/geojson_polygon, https://www.npmjs.com/package/@mapcomponents/react-maplibre, etc. https://imfeld.dev/writing/svelte_domless_components and other articles on the same blog explain the reasoning here. I think we should consider switching to this style too.

Also, if you edit this file and let Vite auto-reload, you'll see it crash! You have to hard refresh. It's because while some component gets recreated the maplibre `Map` is null, and we try to `addSource`. We need to figure out how to robustly add stuff to the map, or defer it to when the map is loaded, and handle all the weird Svelte component orders

robinlovelace-ate:
> If anyone has color or symbolizing suggestions either for the colors we're using, or to change the road labels, it's easy to try things out!

I think given the current way of doing it, and without complexity of rendering our own road names (v. low priority, probably not in scope), it seems like a balance between line opacity and seeing road name. For me, given that this is aimed at transport planning professionals who know their areas well, I would favour visibility of ATIP layers, I think we have that now.

robinlovelace-ate:
Lighter shade of blue?

dabreegster:
> Lighter shade of blue?

Screenshot from 2023-05-15 22-05-46
Yeah actually this looks pretty nice to me. I don't remember if the 4 colors originally came from some kind of palette that's important to keep as "qualitative / discrete", so maybe we ought to adjust everything. This is `#BCF5F9`, set in `colors.ts`. I'll wait for Alex's and Pete's opinion, but otherwise I'd be happy switching to this as an incremental improvement

robinlovelace-ate:
> I'd be happy switching to this as an incremental improvement

Same, although my one comment would be maybe a touch darker would be better, that is a super light blue.

PR #169, opened 2023-05-16
New speed limit layers

dabreegster:
For #69. To try this out: https://acteng.github.io/atip/speed_limit/scheme.html?authority=West%20of%20England%20Combined%20Authority

# The new features

After drawing a route, you can now add a layer showing the current speed limits along that route, according to OSM:

https://github.com/acteng/atip/assets/1664407/2a13f15b-a827-4f82-8142-812710304c5e

And you can also color _all_ roads with known data:

https://github.com/acteng/atip/assets/1664407/ccfd8940-8ca4-400e-af79-d42463ef0366

# Data quality

This is [maxspeed](https://wiki.openstreetmap.org/wiki/Key:maxspeed) from OSM. As you can see, it's often not tagged at all, especially on smaller streets:
Screenshot from 2023-05-18 15-27-00
I also spot suspicious differences between two sides of the A38 dual carriageway -- 40mph vs 70mph!
Screenshot from 2023-05-18 15-27-52

The help tooltip warns the user about data quality problems. We can consider integrating with something like https://github.com/westnordost/osm-legal-default-speeds in the future to infer speeds for unknown roads, or to sanity check explicitly tagged speeds.

# Design problems

I chose colors and 4 buckets of speeds (in mph) pretty arbitrarily. This is very easy to change, if anyone in ATE has an opinion.
Screenshot from 2023-05-18 15-29-21

There's a fair bit of clunkiness with the new speed limit layer interacting with other controls. For example, you can still hover and click on any interventions drawn:

https://github.com/acteng/atip/assets/1664407/0d696ff1-b8b9-49ef-a203-a7458318d845

Should we hide everything else on the map and disable all the toolboxes on the right while a speed limit layer is active? Or is that overkill?

# Implementation details

Context for Pete: #154. We load a per-area file containing https://github.com/a-b-street/osm2streets data, load it async in a web worker, and call Rust code through WASM.

Related backend changes are already done: https://github.com/a-b-street/osm2streets/pull/217, https://github.com/acteng/abstreet-to-atip/commit/2a5dc68e4152f93acbf7c3a84e6605aacfccb561


index.html
```diff
@@ -11,7 +11,7 @@
   <body>
     <div id="app"></div>
     <script type="module">
-      import ChooseArea from "./src/ChooseArea.svelte";
+      import ChooseArea from "./src/pages/ChooseArea.svelte";
```
dabreegster, line 14: Oh yeah, I forgot to mention in the PR description.. I also renamed/organized a bunch of Svelte components as part of this PR. Apologies; it makes the diff larger. I can split into a separate PR if useful. I just wound up getting annoyed at the too-flat structure in the middle of working on this PR.

route_info/src/lib.rs
```diff
@@ -17,8 +26,9 @@ impl RouteInfo {
     pub fn new(input_bytes: &[u8]) -> Result<RouteInfo, JsValue> {
         // Panics shouldn't happen, but if they do, console.log them.
         console_error_panic_hook::set_once();
+        console_log::init_with_level(log::Level::Info).unwrap();
```
dabreegster, line 29: Any `info!` and similar logs wind up in browser console

route_info/src/lib.rs
```diff
@@ -39,23 +49,7 @@ impl RouteInfo {
...
-            let last = pts.last().unwrap();
-            let pt2 = LonLat::new(last[0], last[1]);
-            (
-                pt1.to_pt(&self.network.gps_bounds),
-                pt2.to_pt(&self.network.gps_bounds),
-            )
-        } else {
-            return Err(err_to_js("couldn't get line-string"));
-        };
+        let (pt1, pt2) = self.linestring_endpoints(&feature).map_err(err_to_js)?;
```
dabreegster, line 52: Refactored into `map_matching.rs`

route_info/src/lib.rs
```diff
@@ -88,8 +82,82 @@ impl RouteInfo {
...
+    pub fn speed_limit_for_route(&self, raw_waypoints: JsValue) -> Result<String, JsValue> {
+        let raw_waypoints: Vec<RawRouteWaypoint> = serde_wasm_bindgen::from_value(raw_waypoints)?;
+        let waypoints = self.parse_waypoints(raw_waypoints).map_err(err_to_js)?;
+
+        let mut features = Vec::new();
+        for pair in waypoints.windows(2) {
+            if let (Waypoint::Snapped(_, i1), Waypoint::Snapped(_, i2)) = (&pair[0], &pair[1]) {
+                if let Some(path) = self.geometric_path(*i1, *i2) {
+                    // Add one LineString per Road where speed_limit is defined.
+                    // We could try to glue together contiguous LineStrings where the speed is the
```
dabreegster, line 99: A previous commit in this PR did things in a much more complicated way. It's preserved in git if we ever need something like it again.

src/lib/common/HelpIcon.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+  import icon from "../../../assets/help.svg";
+
+  export let contents: string;
+</script>
+
+<img src={icon} title={contents} alt={contents} />
```
dabreegster, line 7: In Firefox, the tooltip takes a second to appear. And maybe the help icon looks clickable and it isn't, and that's an issue? Happy to do something better if there are specific ideas/examples

src/lib/forms/FormV1.svelte
```diff
@@ -68,7 +69,8 @@
   Length: {prettyPrintMeters(length_meters)}
   <br /><button type="button" on:click={() => autoFillDetails()}
     >Auto-fill details</button
-  >
+  ><br />
+  <RouteInfoLayers {routeInfo} {id} />
```
dabreegster, line 73: TODO later: use all of this for v2 as well

src/lib/layers/ContextualLayers.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
```
dabreegster, line 1: Making room for more map-wide overlays here! The hard part, IMO, is UX decisions raised in the PR description about disabling other layers/controls while one of these is active.

src/lib/layers/SpeedLimits.svelte
```diff
@@ -0,0 +1,94 @@
...
+    ["get", "speed_limit"],
+    colors[0],
+    speedLimitSteps[0],
+    colors[1],
+    speedLimitSteps[1],
+    colors[2],
+    speedLimitSteps[2],
+    colors[3],
+  ];
+  // NOTE! There's only ever one source and layer with this name. This component
```
dabreegster, line 43: This is the first time in ATIP we do something like this; all other sources and layers live for the duration of the app. So we need the `onDestroy` here.

src/lib/layers/SpeedLimits.svelte
```diff
@@ -0,0 +1,94 @@
...
+  import type { Feature } from "../../types";
+  import { lineWidth } from "../../colors";
+  import { map, gjScheme } from "../../stores";
+  import { type Remote } from "comlink";
+  import { type RouteInfo } from "../../worker";
+  import MapTooltips from "../common/MapTooltips.svelte";
+  import DiscreteLegend from "../common/DiscreteLegend.svelte";
+  import HelpIcon from "../common/HelpIcon.svelte";
+
+  export let routeInfo: Remote<RouteInfo>;
```
dabreegster, line 23: Plumbing is getting kind of annoying. It's logically a global; maybe we should use a Svelte store.

dabreegster (commented):
And just recording some older thoughts about how to implement this. I was also thinking about the Rust layer returning segments described by linear referencing. Something like "from 0 to 500m, the speed is 30mph. from 500 to 700m, it's unknown. from 700 to 1km, it's 50mph". Using linear referencing then raises the question of which geometry to use -- the raw linestring from route-snapper? Or the version in the osm2streets `StreetNetwork` that we reconstruct?

In the end, all of that thinking is complicated. At the end of the day, we want to draw on the map with tooltips, so we need geometry anyway. We can just continue to communicate in GeoJSON and return more things to draw.

robinlovelace-ate:
This looks amazing Dustin. The fact that you can see *all* roads is a bonus!

Speed is absolutely critical yet under-considered in network design. In unpublished work I did ages ago 20 mph seemed to be highly correlated with active travel. So makes sense for this to be the first 'data layer'.

One question in terms of UI/UX: this does add complexity. So if we wanted to make a stripped down version for certain use cases, or if it's deemed 'tmi' for next survey, is it easy to turn off? Imagine so given the modular way ATIP is being built but thought I'd ask and assume we will not turn it off as the feature is awesome.

robinlovelace-ate (approved):
Approved based on outcome not code.

dabreegster:
> is it easy to turn off?

Yes, we can "feature gate" things. An example is disabling most draw controls in planning mode: https://github.com/acteng/atip/blob/fed65194027c3410090fa186a61205bbd9098635/src/lib/draw/Toolbox.svelte#L145. Hopefully we can iterate on the UX and make this less confusing before v2 ships. Is there something in the UI that feels particularly complex that we could de-emphasize by default?

robinlovelace-ate:
> Is there something in the UI that feels particularly complex that we could de-emphasize by default?

Not right now for me. That question is probably best asked of users in inspections.

robinlovelace-ate:
Just spotted an issue. Easy to fix this before merging @dabreegster ? 

https://github.com/acteng/atip/issues/177

dabreegster:
Fixed #177 now:
Screenshot from 2023-05-19 12-22-15
Screenshot from 2023-05-19 12-22-39

route_info/src/map_matching.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl RouteInfo {
+    /// Extract the endpoints of a LineString in map-space.
+    pub(crate) fn linestring_endpoints(&self, feature: &geojson::Feature) -> Result<(Pt2D, Pt2D)> {
+        // TODO Duplicate non-adjacent points; make a more permissive version of PolyLine.
+        //let polyline = PolyLine::from_geojson(feature, Some(&self.network.gps_bounds))?;
+
+        if let Some(geojson::Geometry {
+            value: geojson::Value::LineString(ref pts),
+            ..
```
Sparrow0hawk, line 16: I think I understand `if let` but can you run me through this @dabreegster? And also the `..`?

Sparrow0hawk: This is how you unpack the `pts` from a GeoJSON feature I think

dabreegster: We have a [geojson::Feature](https://docs.rs/geojson/latest/geojson/struct.Feature.html) struct. It's got a `geometry: Option<Geometry>` field. Inside there is a [value enum](https://docs.rs/geojson/latest/geojson/enum.Value.html) with a bunch of cases. The `if let` is like Rust's [match](https://doc.rust-lang.org/rust-by-example/flow_control/match.html) statement, but for handling just one case. So we're saying "if this GJ feature has geometry that's defined, and the value inside is the LineString enum case". If so, we're grabbing the points in that enum case and calling them `pts`. `ref` means to immutably borrow -- we don't need to "move" or "take ownership" of the data in there, just have a read-only pointer to it.

The array lookup stuff below is based on the type that we extracted: https://docs.rs/geojson/latest/geojson/type.LineStringType.html

dabreegster: https://doc.rust-lang.org/rust-by-example/flow_control/match/destructuring/destructure_enum.html is a better reference for `match` / `if let`

route_info/src/map_matching.rs
```diff
@@ -0,0 +1,81 @@
...
+                    .ok_or_else(|| anyhow!("no intersection close to waypoint"))?;
+                waypoints.push(Waypoint::Snapped(pt, i));
+            } else {
+                waypoints.push(Waypoint::Free(pt));
+            }
+        }
+        Ok(waypoints)
+    }
+
+    /// Calculates a path between intersections, ignoring all semantics of the roads (lane types,
```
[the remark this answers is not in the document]
Sparrow0hawk: And this is because the .bin files from route-snapper could plausibly be created with an older version of OSM data?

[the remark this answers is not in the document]
dabreegster: Correct. The tension here is that we're using two different files. The route-snapper uses a [simple one](https://github.com/dabreegster/route_snapper/blob/main/route-snapper-graph/src/lib.rs), and osm2streets uses a richer one with much more structure. The route-snapper one is built from osm2streets and the pathfinding implemented in this PR matches everything, so the results _should_ be the same, if the same input .osm was used to generate both.

In the short term, I think doing this simple matching and generating both files in https://github.com/acteng/abstreet-to-atip/ off the same OSM input will be fine. We might hit some weird edge case.

Longer term, it'd be nice to only load one file, instead of the two. The route snapper could dynamically create its in-memory graph from the osm2streets `StreetNetwork`. That'd make our lives easier -- less chance of things getting out of sync, less files to generate and manage. The two different files are an artifact of the order things have been developed.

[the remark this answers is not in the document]
dabreegster: If we hit edge cases, it'll be visibly obvious to the user. They see the route they draw in blue. When they overlay the speed limit, if it doesn't match up with their route, it'll be pretty clear something has gone wrong.

src/lib/layers/SpeedLimits.svelte
Sparrow0hawk: Feels like it should be a store? Especially if we're going to draw out more contextual layers like this.

src/worker.ts
```diff
@@ -32,6 +33,22 @@ export class RouteInfo {
 
     return this.inner.nameForRoute(linestring);
   }
+
+  speedLimitForRoute(waypoints: Waypoint[]): string {
+    if (!this.inner) {
+      throw new Error("Need to loadFile before using RouteInfo");
```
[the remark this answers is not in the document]
dabreegster: Done. I'd like to maybe surface a second progress bar (but where?) and disable the controls that try to invoke route info until it's ready. Not sure how yet, so a better error message is a good interim

dabreegster:
Thanks for the review! I'll merge this now, though async feedback from Pete is still welcome. This is very much just a first cut still.

> At this stage I'm comfortable with an arbitrary choice and think we should iterate on this based on some user feedback

SGTM! Already updated these arbitrary choices thanks to Robin's feedback (#177 for context). Agreed about color schemes; we need to revisit these generally.

> Is it to assist someone drawing an intervention along a road based on speed (if so do we make the speed limits more transparent when in draw route mode?)? Or are we providing it as an extra layer that surfaces some data that someone wants to check about an existing plotted intervention (in which case maybe disable everything).

Good question... I don't know. The one case I'm imagining is to understand context _near_ an intervention. Speed along a route is one thing, but if some of the cross streets to the route are high-speed, maybe that matters. I could see a use for editing geometry while this layer is on, in which case we do need to adjust opacity and z-order.

For now I'll leave things as they are, and we can iterate as we get feedback / design advice.

src/worker.ts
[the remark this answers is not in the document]
dabreegster: UX-wise that's a little confusing to me, since the snap tool usually finishes loading first (smaller file) and the tool is usable before the route info is ready. I was thinking about putting the progress bar near the two dropdowns that let you enable speed limit, but maybe that's also a bit confusing.

I also want to overhaul how we do progress bars; the one for the route snapper is scary old non-TS code. I started https://github.com/acteng/atip/tree/progress, but could use help polishing / figuring out all the edge cases. Oddly enough, I've had trouble finding a simple library on NPM to do this kind of thing!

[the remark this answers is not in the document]
dabreegster: #181

Issue #160, opened 2023-05-11
Minor feature request: change wording above drawing tools to be more generic or change them depending on the schema


robinlovelace-ate:
As shown below ATIP currently says "Click on an intervention" which is not right for reporting criticals

![image](https://github.com/acteng/atip/assets/122299965/f89105e9-d850-464d-b2a8-24f8bc4531bb)

Suggested text

> Click on a tool to document the location of a critical issue

(or similar)

robinlovelace-ate:
Reproducible example: https://acteng.github.io/atip/scheme.html?authority=Isle+of+Wight&schema=criticals#14.72/50.63716/-1.40675

dabreegster:
We use "intervention" in many places in both the code and user-visible app. "0 interventions" on the left, in the instructions, in the legend, etc.

Should we show a per-schema noun? For criticals, "critical issue." For planning, what should it be -- "planning application" or "application"?

Or should we aim for a kind of generic word and just say "object"?

robinlovelace-ate:
> Or should we aim for a kind of generic word and just say "object"?

For now a generic word like object could be fine:

> Click an object to fill out its attributes 

works for all modes. :+1: to the suggested fix, good thinking, and will help reduce code complexity and number of lines of code, for now...

robinlovelace-ate:
Reopening because we've had clear communication that Object is not good for the main use case. I suggest reverting this or making the word "Object" change depending on the mode.

dabreegster:
So what specifically do we want to change it to? When the schema is v1/v2 (not criticals or planning), do we again use "intervention"?

For planning mode, maybe "development"?

For critical issue mode, maybe "issue"?

robinlovelace-ate:
#256 has a fix for this I think.

robinlovelace-ate:
> So what specifically do we want to change it to? When the schema is v1/v2 (not criticals or planning), do we again use "intervention"?

Yes for sure, that's what Lynsey and the inspections team wanted. In hindsight we shouldn't have changed it to 'object' in the first place, apologies for jumping the gun on that decision which we're now reverting.

> 
> For planning mode, maybe "development"?

Development is perfect

> 
> For critical issue mode, maybe "issue"?

I vote `Critical issue` to be specific.

dabreegster:
Closed by the PR above

PR #171, opened 2023-05-17
Use 'object' instead of 'intervention' some places in the user-visibl…

dabreegster:
…e text. #160

See #160 for context. We still use "intervention" many places in the code -- ideally we'd be consistent about the terminology everywhere. "Object" feels a bit generic, and "object on the map" is too verbose. If there are ideas for what to do, we can adjust here, otherwise this PR will just adjust a bit of the user-visible text.

### DIFF
--- a/src/lib/EntireScheme.svelte
+++ b/src/lib/EntireScheme.svelte
@@ -159,7 +159,7 @@
 <br />
 
 <div>
-  <span>{$gjScheme.features.length} interventions</span>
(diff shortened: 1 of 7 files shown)


robinlovelace-ate (approved):
Looks good. I think it may be good to mention interventions somewhere in schema-dependent text, but that's another issue, will see how it looks post merge and open another issue if I spot anything else. Great work Dustin!

PR #174, opened 2023-05-17
Clean up custom event listeners

dabreegster:
This is the final round of fixing our code to properly teardown event listeners in Svelte components. When editing components locally with HMR, these callbacks were previously being duplicated from past loads.

To verify this works, I added logging statements in PointMode's callbacks, edited the component repeatedly, and watched the console when clicking. Before this PR, multiple log statements would stack up. With this change, we just have one, as expected.

CC @yongrenjie, who may hit similar problems in a new Svelte app. Any callback created by a Svelte component has to get cleaned up correctly, or local edits will get confusing.


dabreegster:
Argh, actually I repeated the test and now it's looking flaky. Not sure why, and also not sure if we could automate a test here somehow, so moving to draft for now.

dabreegster:
Cleaning up old PRs+branches. Enough has changed with the new event plumbing that we'd need to re-think this anyway. The problems remain, but they're not so bad during live development.

Issue #172, opened 2023-05-17
Loading a geojson file directly from a URL parameter


Sparrow0hawk:
A brief feature idea from a discussion today:

If we have a list of scheme .txt files, it would be nice to click a link that automatically opens ATIP with that .txt file loaded. 

## Considerations

- This would only work for public URLs 
- Is this even possible/practical
- How would this interact with existing data in local storage?
- Is there a better way to do this that isn't directly in the ATIP code base

dabreegster:
We could support this pretty easily -- just `fetch` the file, then load it up same as https://github.com/acteng/atip/blob/41be5dbc720ffa6907391ff5cd9d41cc4850582e/src/lib/EntireScheme.svelte#L98.

> How would this interact with existing data in local storage?

It would blindly overwrite it, the same as clicking "Load file" and opening something locally. We may want to revisit that generally; right now local storage only keeps one file per (authority, schema) combo. But also, local storage has something like a 5MB limit per website, so ATIP power users are going to hit problems at some point! I'd love to figure out how to manage more / larger files generally, and have some kind of cloud-ish uploading/sharing.

robinlovelace-ate:
Does this need to be documented? Great new feature, @becfleming-ate was asking about this so will be pleased I imagine!

dabreegster:
Good point! https://github.com/acteng/atip#query-parameters

PR #175, opened 2023-05-17
Load from a URL specified in query params. #172

dabreegster:
Quite straightforward! Just add `geojsonUrl=xyz`

I tried to add a test for this:
```
test("loading from URL", async ({ page }) => {
  await page.goto(
    "/scheme.html?authority=Isle of Wight&geojsonUrl=https://gist.github.com/dabreegster/e3aed0cffc59e8ac47e2e303641bac
39/raw/b03d1afc64d4d998b7b564671f257022b012031b/Isle%2520of%2520Wight_Holiday.txt"
  );
  await page.getByRole("button", { name: "1) Hike part 1" }).click();
});
```

But hit CORS errors; we can't fetch from Github gists like this. Not sure how to host a file locally with Playwright. If anyone's got another idea for a stable-over-time URL to stash a tiny txt file, we could use it. (Is it worth putting in the S3 bucket?)

### DIFF
--- a/src/lib/EntireScheme.svelte
+++ b/src/lib/EntireScheme.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { onMount } from "svelte";
   import length from "@turf/length";
   import {
     gjScheme,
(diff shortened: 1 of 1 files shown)


dabreegster:
I'll merge now so @becfleming-ate can use it. You can generate a URL like `https://acteng.github.io/atip/scheme.html?authority=XYZ&geojsonUrl=https://abc.com/foo.txt`. Let me know if you hit any problems with that!

PR #182, opened 2023-05-22
Add a tool to open Streetview

### LINKED ISSUE #111: Add Streetview convenience
Embed or open in another tab easily. Bing, Google, Mapillary, anything else

dabreegster:
Closes #111

https://github.com/acteng/atip/assets/1664407/463c5832-1da6-421d-b725-e88a6b57e9d4

The UX is slightly clunky, but it feels useful to me. See issue for details about licensing -- I think we're fine to open a totally new tab. Since we're not embedding imagery or showing two maps at the same time, we don't need an API key or anything else.


assets/street_view.svg
dabreegster, line 1: Somehow this is a standard Material icon! https://fonts.google.com/icons?icon.query=streetview

src/lib/draw/StreetViewMode.svelte
```diff
@@ -0,0 +1,131 @@
+<script lang="ts">
```
dabreegster, line 1: Mostly copied this code  from `SplitRouteMode.svelte`

src/lib/draw/StreetViewMode.svelte
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  function onClick() {
+    if (mode != thisMode) {
+      return;
+    }
+
+    let [lon, lat] = cursor.geometry.coordinates;
+    if (imagery == "google") {
+      window.open(
```
dabreegster, line 72: The URL is mostly self explanatory, but some magic params lifted from https://nptscot.github.io/#8/55.953/-3.138. I couldn't find a clear guide anywhere explaining these params and letting us, for example, always start facing north

src/lib/draw/StreetViewMode.svelte
```diff
@@ -0,0 +1,131 @@
...
+        `http://maps.google.com/maps?q=&layer=c&cbll=${lat},${lon}&cbp=11,0,0,0,0`,
+        "_blank"
+      );
+    } else if (imagery == "bing") {
+      window.open(
+        `https://www.bing.com/maps?cp=${lat}~${lon}&style=x`,
```
dabreegster, line 78: https://learn.microsoft.com/en-us/bingmaps/articles/create-a-custom-map-url?redirectedfrom=MSDN partly explains params, but `style=x` is... magic I discovered somewhere years ago and put in the A/B Street parking mapper tool. Still works, and the other ones documented here do not, so...

src/lib/draw/StreetViewMode.svelte
```diff
@@ -0,0 +1,131 @@
...
+        "_blank"
+      );
+    } else if (imagery == "mapillary") {
+      window.open(
```
dabreegster, line 82: This one's a bit clunky because you have to additionally click on a specific photo, but I don't think anything faster is possible without properly integrating with their API

src/lib/draw/StreetViewMode.svelte
```diff
@@ -0,0 +1,131 @@
...
+  // The escape key isn't registered at all for keypress, so use keydown
+  function onKeyDown(e: KeyboardEvent) {
+    if (mode == thisMode && e.key == "Escape") {
+      changeMode("edit-attribute");
+      e.preventDefault();
+    }
+  }
+</script>
+
+{#if mode == thisMode}
```
dabreegster, line 109: UX issue: the user's choice here is lost everytime they switch tools. We could store somewhere globally and plumb through. The other option so far is the route tool's "avoid doubling back." Maybe it's time to start something for this? And should we save it as a cookie or local storage?

src/lib/draw/Toolbox.svelte
```diff
@@ -171,6 +175,16 @@
     {/if}
     <SplitRouteMode bind:this={splitRouteMode} {mode} {changeMode} />
   </div>
+  <div>
```
dabreegster, line 178: I put this new feature as a drawing/editing tool. That's kind of misleading, because it's just for read-only info. But the controls are mutually exclusive with everything else -- it takes over some keybindings and map controls, so the toolbox is kind of our place to do that for now.

dabreegster (commented):
Preview at https://acteng.github.io/atip/streetview/scheme.html?authority=Derby

dabreegster:
Ping @Sparrow0hawk -- how's the usability of this feel to you, and any licensing/credit issues I might've unforeseen?

src/lib/draw/StreetViewMode.svelte
Sparrow0hawk: I'd be inclined, as the behaviour is different to the other two, to potentially drop this until we've looked at integrating their API. Just as the UX is slightly different to what you might expect.

src/lib/draw/StreetViewMode.svelte
Sparrow0hawk: I think the idea of starting something like `userSettings` might be useful. Haven't got any opinions at present about the best location for that and happy to be steered by others

src/lib/draw/StreetViewMode.svelte
```diff
@@ -0,0 +1,131 @@
...
+  </label>
+  <br />
+  <label>
+    <input type="radio" bind:group={imagery} value="mapillary" />
+    Mapillary
+  </label>
+
+  <ul>
+    <li><b>Click</b> on the map to open imagery</li>
+    <li>Press <b>Escape</b> to cancel</li>
```
Sparrow0hawk, line 127: Again not sure of the perfect wording here.
```suggestion
    <li>Press <b>Escape</b> to exit this mode</li>
```

src/lib/draw/StreetViewMode.svelte
```diff
@@ -0,0 +1,131 @@
...
+    Bing Streetside
+  </label>
+  <br />
+  <label>
+    <input type="radio" bind:group={imagery} value="mapillary" />
+    Mapillary
+  </label>
+
+  <ul>
+    <li><b>Click</b> on the map to open imagery</li>
```
Sparrow0hawk, line 126: I'd recommend wording changes just to make clear this is going to open a new tab
```suggestion
    <li><b>Click</b> on the map to open a new tab with a 3rd party street view provider</li>
```
I don't know if that is too wordy...

src/lib/draw/Toolbox.svelte
Sparrow0hawk: I think this is fine for now.

Sparrow0hawk (changes requested):
I'm happy overall with changes here and think this is ready to merge bar sorting out the wording of this tool.

I think opening a new tab (and clearly stating their third party providers) means we aren't in contravention of any license of usage terms and also don't need to attribute providers anymore than we have already.

dabreegster:
Thanks for the feedback! Everything addressed, and rebased, so merging

PR #183, opened 2023-05-22
Refactor code to create layers

dabreegster:
Part of #89. Previously, we used one big `overwriteLayer` call, but couldn't specify `LayerSpecification` on one of the parameters without getting some bizarre TS errors. After fighting this for a while, I decided it's more clear to split the function into 3 smaller ones specifically for circles, polygons, and lines. That makes the parameters easier to specify, anyway.

This should be a pure refactoring PR, no behavioral change (that I've spotted!)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@turf/nearest-point-on-line": "^6.5.0",
         "@types/geojson": "^7946.0.10",
         "comlink": "^4.4.1",
+        "just-flush": "^2.3.0",
         "maplibre-gl": "^2.4.0",
         "route-snapper": "^0.1.13",
         "svelte": "^3.54.0"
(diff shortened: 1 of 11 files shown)


dabreegster:
Oh yeah, and to mention the ultimate goal: expressing MapLibre sources and layers as Svelte components would be great. It makes it trivial to create/destroy them along with a larger component, and is probably more readable. It'd look something like:

```
<Source id="boundary" data={mask(boundaryGeojson)}>
  <PolygonLayer id="boundary" source="boundary" color="black" opacity={0.5} />
</Source>
```

I have a WIP in `source_layer_helpers` inspired by https://github.com/dimfeld/svelte-maplibre/, but it's not ready yet. I think this refactor is still an incremental step towards this, because it switches to nice named params

dabreegster:
Merging this one as non-controversial, but happy for async feedback

PR #184, opened 2023-05-23
Add a layer to render lane details along a route

dabreegster:
Now that we have osm2streets.org hooked up to ATIP, it's relatively easy for us to attempt something kind of cool: rendering detailed lanes along a route! CC @BudgieInWA. Preview at https://acteng.github.io/atip/layer_lanes/scheme.html?authority=Derby.

https://github.com/acteng/atip/assets/1664407/dbab2d42-beeb-4093-a60b-ab475b88d568

# Design questions

First of all, there are loads of caveats with the rendering. The video above shows a few glaring problems! A big question should be how people will use this; does it solve any concrete problem right now? Is the danger of misinterpretation high and we should work towards this kind of thing eventually, but only add it once it's stabler? My motivation is partly to make osm2streets work even more visible, to drive improvements and feature requests (as a starter -- specializing the marking styles and default lane widths for the UK!).

The blue route underneath is still visible (and clickable), as with speed limits. From far away, this looks strange, and when zoomed in, it can also interfere.

It'd be cool to add a tooltip when you hover on lanes, maybe estimating the current width shown.

# Scope

Right now it only renders things along the route. (And in the example video, the snapping is a bit wrong, so it fails to match up the route -- same issue with the speed limit layer). Maybe it'd be more useful to show a bit of surrounding context, like all of the roads that connect to anything along the route?

I originally tried rendering the entire area, but even in the smallest cases, we hit major performance problems. The entire pipeline is inefficient at that scale -- rendering something in-memory to GeoJSON, serializing it as a string, possibly copying the string (I'm still not totally sure how things are transferred around by web workers), parsing it in the browser, adding it to MapLibre (which internally tiles it and converts to MVT). In A/B Street, there's none of that overhead, and still we have to lazily render details as we pan around while zoomed in. I have some ideas for how to achieve the same thing in ATIP (by lazily rendering something tile-like, or possibly by drawing directly to the OpenGL context from the web worker) but it's nothing easy to do quickly.

### DIFF
--- a/src/lib/layers/LaneDetails.svelte
+++ b/src/lib/layers/LaneDetails.svelte
@@ -4,6 +4,9 @@
   import LanePolygons from "./lane_details/LanePolygons.svelte";
   import LaneMarkings from "./lane_details/LaneMarkings.svelte";
 
+  // Show along a route if specified, or show all otherwise
+  export let id: number | undefined;
+
   let gj1;
   let gj2;
   let gj3;
@@ -11,17 +14,25 @@
 
   onMount(async () => {
     try {
-      console.log("GIMME");
-      let raw = await $routeInfo.renderLaneDetails();
+      let raw;
+      if (id) {
+        let linestring = $gjScheme.features.find(
+          (f) => f.id == id
+        ) as Feature<LineString>;
+        raw = await $routeInfo.renderLaneDetailsForRoute(
+          linestring.properties.waypoints
+        );
+      } else {
+        raw = await $routeInfo.renderAllLaneDetails();
+      }
       gj1 = JSON.parse(raw[0]);
       gj2 = JSON.parse(raw[1]);
       //gj3 = JSON.parse(raw[2]);
       //gj4 = JSON.parse(raw[3]);
     } catch (e) {
-      window.alert(`Couldn't calculate speed limits for route: ${e}`);
+      window.alert(`Couldn't render lane details: ${e}`);
     }
   });
-
 </script>
 
 {#if gj1}
--- a/src/lib/layers/LaneDetails.svelte
+++ b/src/lib/layers/LaneDetails.svelte
@@ -12,6 +12,8 @@
   // Show along a route
   export let id: number;
 
+  // renderLaneDetailsForRoute returns 4 layers in a certain order. They're
+  // labeled when used below.
   let gj1;
   let gj2;
   let gj3;
--- a/src/lib/layers/LaneDetails.svelte
+++ b/src/lib/layers/LaneDetails.svelte
@@ -1,0 +1,30 @@
+<script lang="ts">
+  import { onMount } from "svelte";
+  import { gjScheme, routeInfo } from "../../stores";
+  import LanePolygons from "./lane_details/LanePolygons.svelte";
+  import LaneMarkings from "./lane_details/LaneMarkings.svelte";
+
+  let gj1;
+  let gj2;
+  let gj3;
+  let gj4;
+
+  onMount(async () => {
+    try {
+      console.log("GIMME");
+      let raw = await $routeInfo.renderLaneDetails();
+      gj1 = JSON.parse(raw[0]);
+      gj2 = JSON.parse(raw[1]);
+      //gj3 = JSON.parse(raw[2]);
+      //gj4 = JSON.parse(raw[3]);
+    } catch (e) {
+      window.alert(`Couldn't calculate speed limits for route: ${e}`);
+    }
+  });
+
+</script>
+
+{#if gj1}
+  <LanePolygons gj={gj1} />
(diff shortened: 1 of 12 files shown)


src/lib/layers/lane_details/IntersectionMarkings.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
```
dabreegster, line 1: These files are ported from Leaflet styles: https://github.com/a-b-street/osm2streets/blob/main/street-explorer/js/layers.js

Ben and I have talked about some ideas for changing the output of the rendering layer to be more semantic (describing something as a dashed center line, and letting the API user figure out how to render that). But nothing concrete yet.

src/lib/layers/lane_details/IntersectionMarkings.svelte
```diff
@@ -0,0 +1,23 @@
...
+  import Layer from "./Layer.svelte";
+  import { caseHelper } from "../../../maplibre_helpers";
+  import type { GeoJSON } from "geojson";
+
+  export let gj: GeoJSON;
+
+  let style = {
```
dabreegster, line 8: We could maybe use the functions added in #183 , but for now, easier to do this

src/lib/layers/lane_details/IntersectionPolygons.svelte
```diff
@@ -0,0 +1,28 @@
...
+  export let gj: GeoJSON;
+
+  let style = {
+    type: "fill",
+    filter: ["==", ["get", "type"], "intersection"],
+    paint: {
+      "fill-color": caseHelper(
+        "intersection_kind",
+        {
+          MapEdge: "#696",
```
dabreegster, line 15: The feature properties for all of the layers are not well-documented anywhere right now. I know them from working with osm2streets, but if I can define / clarify anything, let me know. The things here are from https://github.com/a-b-street/osm2streets/blob/0d6f0f973b3fa8dbdc8223aa6f412a8d997252fc/osm2streets/src/intersection.rs#L48

src/lib/layers/lane_details/LanePolygons.svelte
```diff
@@ -0,0 +1,36 @@
...
+    paint: {
+      "fill-color": caseHelper(
+        "type",
+        // TODO Generate TS types from osm2streets
+        {
+          Driving: "black",
+          Parking: "#333333",
+          Sidewalk: "#CCCCCC",
+          Shoulder: "#CCCCCC",
+          Biking: "#0F7D4B",
```
dabreegster, line 19: This is an old US assumption! Is there a standard color for cycle lanes in the UK? I feel like I've seen both blue and red and just markings in London alone...

src/lib/layers/lane_details/Layer.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
+  import { onMount, onDestroy } from "svelte";
+  import type { GeoJSON } from "geojson";
+  import { map } from "../../../stores";
+  import { overwriteSource, overwriteLayer } from "../../../maplibre_helpers";
+
+  // This component manages a source with exactly one layer. Both are torn down
```
dabreegster, line 7: Kind of towards the idea mentioned in #183

dabreegster:
Ping @Sparrow0hawk -- do you think we should put this in v2, or hold off till it has more polish?

src/lib/layers/lane_details/LanePolygons.svelte
Sparrow0hawk: Page 70 of [LTN1/20](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/951074/cycle-infrastructure-design-ltn-1-20.pdf) doesn't obligate a colour choice in the UK, but does mention green and red surfaces. 

My vote on this would be green!

src/lib/layers/LaneDetails.svelte
```diff
@@ -0,0 +1,49 @@
...
+  let gj2;
+  let gj3;
+  let gj4;
+
+  onMount(async () => {
+    try {
+      let linestring = $gjScheme.features.find(
+        (f) => f.id == id
+      ) as Feature<LineString>;
+      let raw = await $routeInfo.renderLaneDetailsForRoute(
```
Sparrow0hawk, line 27: Note for me. 

This is where the magic happens and we call the new function for rendering lane details from the webWorker

Sparrow0hawk (approved):
I'm happy with the changes here, I think these could be included at this stage in the v2 milestone and think this is definitely a target area for feedback in the next round of usability testing.

dabreegster:
Thanks for the review! I rebased and handled an older refactor. I'll merge this now, and I'm sure we'll have many followup things to tune as people try it out

PR #185, opened 2023-05-23
Small UX tweaks

dabreegster:
From watching UX sessions, two small things jumped out, easy to fix:

The "Auto-fill details" button is too far away from where people type in the name, and its effect is unclear. So move it:
Screenshot from 2023-05-23 14-40-50

In the split tool, someone was confused by "Click near a route". I just meant the mouse doesn't need to be exactly on the route before it snaps, but that's obvious anyway when trying to use it. So adjust the wording to "Click on a route"

### DIFF
--- a/src/lib/draw/route/SplitRouteMode.svelte
+++ b/src/lib/draw/route/SplitRouteMode.svelte
@@ -268,7 +268,7 @@
 
 {#if mode == thisMode}
   <ul>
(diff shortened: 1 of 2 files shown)


Pete-Y-CS (approved):

dabreegster:
Thanks for the review! Merging since it's pretty uncontroversial

PR #186, opened 2023-05-23
Double click to finish a route

dabreegster:
This came up several times in the UX study -- people expected to double click to end a route. For now, I've made double click always try to finish things -- even if someone double clicks a waypoint somewhere in the middle. Is that reasonable, or should we have different behavior?

Try at https://acteng.github.io/atip/double_click_to_end/scheme.html?authority=Derby

This PR is related to https://github.com/dabreegster/route_snapper/pull/30. We'd have to do this PR regardless; the changes in the other one only affect users of the `lib.js`, which we stopped using a while ago.

### DIFF
--- a/src/lib/draw/route/RouteControls.svelte
+++ b/src/lib/draw/route/RouteControls.svelte
@@ -23,7 +23,7 @@
     <li>Hold <b>Shift</b> to draw a point anywhere</li>
     <li><b>Click and drag</b> any point to move it</li>
     <li><b>Click</b> a red waypoint to delete it</li>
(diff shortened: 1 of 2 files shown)


robinlovelace-ate:
Tested with route tool, the main one it will be useful for. Definitely lets merge this. Comments on testing:

- Should double click save freehand tool objects? I vote yes.
- This is a great feature!

robinlovelace-ate (approved):

dabreegster:
Good catch on double click for the freehand polygon tool! I'll followup after the conference with that fix

PR #187, opened 2023-05-23
Rewrite instructions

### LINKED ISSUE #116: Improve instructions
There is little point in creating detailed instructions while the project is not being put out there and while changes are expected. I think this will come into it's own when we have a stable 'v2'.

dabreegster:
Closes #116. I tried to reduce the level of detail, encouraging people to use each tool's instructions or help button.

Some of the instructions may depend on the schema (eg, planning or criticals). Do we want to customize anything, or give more specific guidance about the level of detail in a scheme we're asking for?

### DIFF
--- a/src/lib/sidebar/Instructions.svelte
+++ b/src/lib/sidebar/Instructions.svelte
@@ -7,26 +7,28 @@
 <Modal title="Instructions" bind:open>
   <h2>Drawing objects on the map</h2>
   <p>
-    Use the tools on right to sketch new objects on the map. Each tool has instructions that appear when you click on its icon.
+    Use the tools on right to sketch new objects on the map. Each tool has
+    instructions that appear when you click on its icon.
   </p>
   <p>
-    Modify existing objects on the map using <b>Edit geometry</b> and clicking
-    something.
+    Modify existing objects on the map using <b>Edit geometry</b> and clicking something.
   </p>
 
   <h2>Filling out data</h2>
   <p>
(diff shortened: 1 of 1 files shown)


src/lib/sidebar/Instructions.svelte
```diff
@@ -5,61 +5,32 @@
...
-  <p>
-    Repeat the steps above for each substantial infrastructure intervention in
-    the scheme.
-  </p>
-
+  <h2>Drawing objects on the map</h2>
   <p>
-    To make further edits to attribute data, or to delete an intervention, click
-    on its name in the left hand panel or its geometry in the map.
+    Use the tools on right to sketch on the map. Each tool has a <b>Help</b> button
```
Pete-Y-CS, line 10: I like this change a lot

src/lib/sidebar/Instructions.svelte
```diff
@@ -5,61 +5,32 @@
...
-    scheme name, and start a new set of interventions representing a new scheme.
+    You can modify the object on the map using <b>Edit geometry</b> and clicking
+    something.
   </p>
 
+  <h2>Filling out data</h2>
   <p>
-    To load a GeoJSON file that you have already saved from the tool, click the
-    "Load from GeoJSON" button and select the file.
+    While you're in <b>Edit attributes</b> mode, you can select an object to describe
```
Pete-Y-CS, line 20: maybe a slight rephrase to `you can select an object on the left to describe it` would  be slightly clearer

robinlovelace-ate:
Taking a look, in prep for demo today!

PR #195, opened 2023-05-25
Double click to finish a freehand polygon

dabreegster:
Robin noticed this bug yesterday. Thanks to OSMUS for advice: https://osmus.slack.com/archives/C01G3D28DAB/p1685016451984109

We have a remaining bug double clicking in the snap polygon tool, but the fix for that needs to happen in the route-snapper repo; I'll do it separately.


dabreegster:
Aha, apparently the order double click events are fired is not guaranteed by the browser. I suggest we continue with this approach for the moment until we think of something better, or at least make sure this works in the common browser people're using (Edge?)

https://github.com/maplibre/maplibre-gl-js/pull/2593

robinlovelace-ate (approved):

PR #196, opened 2023-05-25
WIP: State changing bugs 142

Pete-Y-CS:
Work in progress: added an initial somewhat messy go at confirmation modals. This is just a first step towards slightly more robust handling of the rest of the state when taking actions in a component, which can care about state in distantly related components, either to modify that state or change behaviour based on that state

### DIFF
--- a/src/pages/BrowseSchemes.svelte
+++ b/src/pages/BrowseSchemes.svelte
@@ -22,18 +22,15 @@
   const schema = "v1";
 
   interface Scheme {
-    file_name: string;
-    authority_code: string;
-    authority_name: string;
-    internal_scheme_id: string;
-    scheme_priority: number;
+    scheme_reference: string;
+    authority_or_region: string;
     num_features: number;
   }
 
-  // unique by file_name
+  // unique by scheme_reference
   let schemes: Scheme[] = [];
   let filterText = "";
-  // by file_name
+  // by scheme_reference
   let showSchemes: Set<string> = new Set();
 
   // Stats about filtered schemes
@@ -56,12 +53,12 @@
             .toLowerCase()
             .includes(filterNormalized)
         ) {
-          showSchemes.add(feature.properties.atip_file_name);
+          showSchemes.add(feature.properties.scheme_reference);
         }
       }
     } else {
       for (let scheme of schemes) {
-        showSchemes.add(scheme.file_name);
+        showSchemes.add(scheme.scheme_reference);
       }
     }
 
@@ -71,7 +68,7 @@
         return null;
       }
       for (let feature of gj.features) {
-        if (showSchemes.has(feature.properties.atip_file_name)) {
+        if (showSchemes.has(feature.properties.scheme_reference)) {
           delete feature.properties.hide_while_editing;
         } else {
           feature.properties.hide_while_editing = true;
@@ -83,7 +80,7 @@
     // Recalculate stats
     counts = { area: 0, route: 0, crossing: 0, other: 0 };
     for (let feature of $gjScheme?.features) {
-      if (showSchemes.has(feature.properties.atip_file_name)) {
+      if (showSchemes.has(feature.properties.scheme_reference)) {
         counts[feature.properties.intervention_type]++;
       }
     }
@@ -105,22 +102,19 @@
   }
 
   function addSchemeToSidebar(gj: GeoJSON) {
-    let byFilename = {};
+    let byScheme = {};
 
     for (let feature of gj.features) {
       let p = feature.properties;
-      byFilename[p.atip_file_name] ||= {
-        file_name: p.atip_file_name,
-        authority_code: p.authority_code,
-        authority_name: p.authority_name,
-        internal_scheme_id: p.internal_scheme_id,
-        scheme_priority: p.scheme_priority,
+      byScheme[p.scheme_reference] ||= {
+        scheme_reference: p.scheme_reference,
+        authority_or_region: p.authority_or_region,
         num_features: 0,
       };
-      byFilename[p.atip_file_name].num_features++;
+      byScheme[p.scheme_reference].num_features++;
     }
 
-    schemes = Object.values(byFilename);
+    schemes = Object.values(byScheme);
   }
 
   function tooltip(props: { [name: string]: any }): string {
@@ -132,7 +126,7 @@
     let gj = {
       type: "FeatureCollection",
       features: $gjScheme.features.filter(
-        (f) => f.properties.atip_file_name == scheme.file_name
+        (f) => f.properties.scheme_reference == scheme.scheme_reference
       ),
     };
     $map?.fitBounds(bbox(gj), { padding: 20, animate: false });
@@ -142,15 +136,18 @@
     let gj = {
       type: "FeatureCollection",
       features: $gjScheme.features.filter(
-        (f) => f.properties.atip_file_name == scheme.file_name
+        (f) => f.properties.scheme_reference == scheme.scheme_reference
       ),
     };
-    let filename = scheme.authority_name;
+    let filename = scheme.authority_or_region;
     // Assuming the schema is always v1
 
     // Put the file in local storage, so it'll be loaded from the next page
     window.localStorage.setItem(filename, JSON.stringify(gj));
-    window.open(`scheme.html?authority=${scheme.authority_name}`, "_blank");
+    window.open(
+      `scheme.html?authority=${scheme.authority_or_region}`,
+      "_blank"
+    );
   }
 </script>
 
@@ -183,16 +180,12 @@
 
     <ul>
       {#each schemes as scheme}
-        {#if showSchemes.has(scheme.file_name)}
+        {#if showSchemes.has(scheme.scheme_reference)}
           <CollapsibleCard
-            label={`${scheme.internal_scheme_id}: ${scheme.num_features} features`}
+            label={`${scheme.scheme_reference}: ${scheme.num_features} features`}
           >
             <ul>
-              <li>Filename: {scheme.file_name}</li>
-              <li>Priority: {scheme.scheme_priority}</li>
-              <li>
-                Authority: {scheme.authority_name} ({scheme.authority_code})
-              </li>
+              <li>Authority or region: {scheme.authority_or_region}</li>
               <li>
                 <button type="button" on:click={() => showScheme(scheme)}
                   >Show on map</button


src/lib/sidebar/EntireScheme.svelte
```diff
@@ -55,12 +57,17 @@
     }
   }
 
+  function openClearAllDialogue() {
+    displayClearAllConfirmation = true;
+  }
+
+  function cancelClearAll() {
```
dabreegster, line 64: As a more general code style question, when should we try and inline the event handler, and when should we use separate functions like this? I think I've been pretty inconsistent about it so far. I kind of like this style, because the function name can describe "what" instead of "how". Though "displayClearAllConfirmation = false" is also pretty self-descriptive

Pete-Y-CS: As a general rule my preference is towards more lines/words if it makes it more readable. But you're right that here it doesn't necessarily do that.  

The only other thing I was thinking was that we might want to eventually to do some other cleanup at the same time, but I can't think what that would be so don't really feel too strongly either way in this case.

dabreegster: More explicitly readable SGTM. If you find other parts of the code where things are too terse / acronymish, let's clean it up!

Pete-Y-CS:
@dabreegster @Sparrow0hawk taking this out of WIP, but will have a quick look through myself in case there's anything I should obviously change. Main thing to note is that I've used VSCode auto formatter, we might want to decide about coding standards/a shared linter because doing this has introduced some changes which are just formatting. You can get auto linting (or at least a check that your code was appropriately linted) automated during git commits, it worked like this at a previous place I worked at, could be nice.

Pete-Y-CS:
I feel like the behaviour I have isn't quite right. With a bit of playing around I wonder what the best way of determining, not just which mode we're in, but whether we're _actually_ using that tool just yet. For example if you click `Edit geometry` but haven't yet selected something to edit, maybe the buttons shouldn't be disabled. In a similar vein maybe draw point shouldn't disable the buttons. 
What do you think about:
- The behaviour
-  What state tells you whether the tool is actually in use

dabreegster:
Trying it out now...

> VSCode auto formatter

I've been using [npm run fmt](https://github.com/acteng/atip/blob/main/docs/developer_guide.md#installation), which uses prettier. I know nothing about prettier config; we could make it match up to whatever VSCode does

Pete-Y-CS:
> Trying it out now...
> 
> > VSCode auto formatter
> 
> I've been using [npm run fmt](https://github.com/acteng/atip/blob/main/docs/developer_guide.md#installation), which uses prettier. I know nothing about prettier config; we could make it match up to whatever VSCode does

I don't have super strong feelings about formatting/how we auto format, just that we are aligned on it 🤝

dabreegster:
The behavior now feels decent to me... maybe it could be confusing to someone why the buttons are disabled. When they're disabled, we could add `title="Change to edit attributes mode to use this"` or something to help?

Maybe disabling `Export to GeoJSON` is a bit overkill; it's a read-only method. Loading and clearing mutate state, which is why they're tricky to leave active.

> What state tells you whether the tool is actually in use

For `GeometryMode`, `currentlyEditing` being null or not is the relevant thing. For other modes, I'm not sure if it's as clear when things are active or not.

src/lib/common/ConfirmationModal.svelte
```diff
@@ -0,0 +1,27 @@
+<script lang="ts">
+    import Modal from "./Modal.svelte";
+    import { createEventDispatcher } from 'svelte';
```
dabreegster, line 3: `npm run fmt` should change this to use double quotes. I'd vote for that for consistency for now. Happy to switch to something else, as long as I can trigger it from the command line somehow

src/lib/draw/Toolbox.svelte
```diff
@@ -46,6 +47,8 @@
   let splitRouteMode: SplitRouteMode;
   let streetViewMode: StreetViewMode;
 
+  const unsubscribe = currentMode.subscribe((value) => mode = value);
```
dabreegster, line 50: Why don't we remove the local var `mode` on line 40 then and just use the store everywhere? We can more or less just use `$currentMode` in place of `mode`, but I haven't thought through every use of it yet

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -9,16 +9,22 @@
     openFromSidebar,
   } from "../../stores";
   import type { Scheme, Schema } from "../../types";
+  import ConfirmationModal from "../common/ConfirmationModal.svelte";
+  import { currentMode } from "../../stores";
```
dabreegster, line 13: You can include this in the list of imports from the same file a few lines up.

I really wish prettier (or similar) would auto-format / order / group imports...

src/types.ts
```diff
@@ -57,3 +57,12 @@ export interface Waypoint {
   lat: number;
   snapped: boolean;
 }
+
+export type Mode =
```
dabreegster, line 61: Then can we remove `lib/draw/types.ts` entirely and use this one instead?

tests/modes.spec.ts
```diff
@@ -1,19 +1,11 @@
...
-  await page.getByRole("region", { name: "Map" }).click({
-    position: {
-      x: 500,
-      y: 500,
-    },
-  });
-  page.on("dialog", (dialog) => dialog.accept());
-  await page.getByRole("button", { name: "Clear all" }).click();
+  
   // Make sure AttributeMode resets to nothing being selected
```
dabreegster, line 8: Comment out-of-date

dabreegster (commented):
Some code/style comments, but I think the behavior is decent. Should we get feedback from other people about what behavior to disable and when?

Pete-Y-CS:
> Some code/style comments, but I think the behavior is decent. Should we get feedback from other people about what behavior to disable and when?

Will address code stuff. Definitely think what to disable when wants some more thought. My preference if it's not too messy would be to disable only when you're actually editing something. So you can have the route tool open but if you haven't clicked an initial point then it doesn't disable these things. Draw point in this case wouldn't disable anything. Having a specific set of attributes open to edit might want to disable them too: basically I think save, load, and clear should be disabled if the state of the scheme is currently being edited

dabreegster:
> if the state of the scheme is currently being edited

Then maybe in the same way we currently have `start` and `stop` common functions, we have `isActive(): boolean`. And somehow plumb that into a store that's queryable from the sidebar. I think the examples you listed for determining activeness or not all make sense to me

src/lib/common/ConfirmationModal.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+    import Modal from "./Modal.svelte";
+    import { createEventDispatcher } from "svelte";
+    const dispatch = createEventDispatcher();
+
+    export let open: boolean = false;
+    export let displayEscapeButton: boolean = false;
+    export let canCloseByClickingBackground: boolean = false;
+    export let title: string = "Confirm";
+    export let message: string = "Message";
```
dabreegster, line 10:
```suggestion
    export let title: string;
    export let message;
```

this'll force the caller to set both props

src/lib/draw/AttributeMode.svelte
```diff
@@ -10,11 +10,12 @@
     openFromSidebar,
     formOpen,
   } from "../../stores";
+  import type { Writable } from "svelte/store";
 
   const thisMode = "edit-attribute";
 
   // TODO Just tell us if we're the current mode or not, actually...
-  export let mode: Mode;
+  export let currentMode: Writable<Mode>;
```
dabreegster, line 18: We don't need to plumb this as a prop. Stores are effectively app-global. We can import this from `stores.ts` like we do other stuff

src/lib/draw/SelectModeButton.svelte
```diff
@@ -0,0 +1,64 @@
+<script lang="ts">
```
dabreegster, line 1: Looks unused / an old copy of the other button component?

src/lib/common/ConfirmationModal.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+    import Modal from "./Modal.svelte";
+    import { createEventDispatcher } from "svelte";
+    const dispatch = createEventDispatcher();
```
dabreegster, line 4:
```suggestion

    const dispatch = createEventDispatcher();
```

Tiny nitpick, space between imports and other stuff, for consistency. I wish prettier would enforce stuff like this :\

src/lib/common/ConfirmationModal.svelte
```diff
@@ -0,0 +1,32 @@
...
+
+    export let open: boolean = false;
+    export let displayEscapeButton: boolean = false;
+    export let canCloseByClickingBackground: boolean = false;
+    export let title: string;
+    export let message;
```
dabreegster, line 10:
```suggestion
    export let message: string;
```

src/lib/draw/Toolbox.svelte
```diff
@@ -46,6 +47,8 @@
   let splitRouteMode: SplitRouteMode;
   let streetViewMode: StreetViewMode;
 
+  const unsubscribe = currentMode.subscribe(() => {});
```
dabreegster, line 50: Can remove this now

src/lib/draw/Toolbox.svelte
```diff
@@ -59,38 +62,38 @@
       "street-view": streetViewMode,
     };
 
-    if (mode == newMode) {
+    const mode = get(currentMode);
```
dabreegster, line 65: Can't we just do `const mode = $currentMode`? I think it should compile down to the same thing, and also magically insert the unsubscribe: https://svelte.dev/tutorial/auto-subscriptions

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -1,24 +1,30 @@
...
     gjScheme,
     formOpen,
     mapHover,
     sidebarHover,
     openFromSidebar,
+    currentMode,
   } from "../../stores";
   import type { Scheme, Schema } from "../../types";
+  import ConfirmationModal from "../common/ConfirmationModal.svelte";
+  import type { Mode } from "../../types";
```
dabreegster, line 14: Can do this in the import on line 12

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -1,24 +1,30 @@
...
 
   export let authorityName: string;
   export let schema: Schema;
+  let mode: Mode = "edit-attribute";
```
dabreegster, line 18: We can also use `$currentMode` here, not need the local var that's synced to it, and not need `unsubscribe`

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -1,24 +1,30 @@
...
+  const unsubscribe = currentMode.subscribe((value) => (mode = value));
 
   let baseFilename = authorityName;
   if (schema != "v1") {
     baseFilename += `_${schema}`;
   }
 
   let loaded = false;
+  let displayClearAllConfirmation: boolean = false;
```
dabreegster, line 27: nit: `:boolean` feels a bit overkill given type inference

src/types.ts
```diff
@@ -57,3 +57,12 @@ export interface Waypoint {
...
+  | "edit-attribute"
+  | "edit-geometry"
+  | "route"
+  | "point"
+  | "free-polygon"
+  | "snap-polygon"
+  | "split-route";
```
dabreegster, line 68: Aha, `npm run check` to the rescue!

```
StreetViewMode.svelte:103:6                   
Error: This comparison appears to be unintentional because the types 'Mode' and '"street-view"' have no overlap. (ts)

{#if $currentMode == thisMode}
```

We need `street-view` here; it was added during a rebase

Pete-Y-CS: Ahhh I was seeing this error and looking in the wrong types file: was very confused why I couldn't see the problem!

A great example of why I should've deleted the duplicate types file

src/lib/draw/Toolbox.svelte
```diff
@@ -59,15 +60,16 @@
...
+    const mode = $currentMode;
+    if ($currentMode == newMode) {
       console.log(`Mode is already ${mode}, not changing`);
       return;
     }
     console.log(`Stopping old mode ${mode}`);
     modes[mode].stop();
-    mode = newMode;
+    currentMode.set(newMode);
     console.log(`Starting new mode ${mode}`);
```
dabreegster, line 71: This needs to be `newMode` now

dabreegster:
Summarizing where we're at with this...

- The bug: Create a polygon. Edit geometry, click the polygon, then click clear all. Currently, we stay in edit mode, but finalizing the polygon crashes and leaves the page unusable.
- The fix in this PR: while we're outside of "edit attribute" mode, disable buttons that would destroy the current objects loaded (clear all & load geojson).
- The ideal fix: Same, but only disable if another tool is being "actively" used. For edit geometry mode, we're only active once we click something.

Implementation:
- This PR moves `mode` from a property in `Toolbox` that's plumbed everywhere to a global store, so we can check it from the sidebar. This unexpectedly uncovered a bug: click new point, create the point -- no form is opened. The reason is that the click event gets handled both by `point_tool.ts` and `AttributeMode.svelte`. The latter sees no object being clicked, so it does `formOpen.set(null)`, clobbering the previous change. We're unclear right now why using a store vs changing a prop has changed the ordering here. The clean fix is probably to mark the click event as handled in the point tool, and stop it from propagating, but some initial attempts to do this with `e.preventDefault()` and `e.originalEvent.stopPropagation()` aren't working.
- #210 adds some tests that catch this regression with the form not opening

I'd suggest these next steps happen as low-pri / not blocking for v2:
- Make a PR that _just_ moves the current mode into a store, with no behavior changes anywhere, using the new tests to verify that. That probably forces us to figure out the click ordering problem.
- Figure out how to expose a concept of "is the mode active and should block clearing all". Make the behavioral change to disable the buttons.
- Rethink how event handling between all the tools is happening generally. Every mode handles every single event right now, and immediately returns if the mode isn't active. It seems much cleaner to delegate the events to only the active mode.

And suggest a simple fix immediately for v2:
- In edit geom mode when we finalize something and try to update the feature, handle the feature vanishing with a popup warning the user they just did something silly

Pete-Y-CS:
Reopening with a new PR: starting the confirmation modal and so on afresh

PR #201, opened 2023-05-30
Store some user settings in local storage

dabreegster:
Not many for now, but this could be useful over time

### DIFF
--- a/src/stores.ts
+++ b/src/stores.ts
@@ -1,7 +1,7 @@
 import { writable, type Writable } from "svelte/store";
 import { emptyGeojson } from "./maplibre_helpers";
 import type { Map } from "maplibre-gl";
-import type { Scheme } from "./types";
+import type { Scheme, UserSettings } from "./types";
 import { type Remote } from "comlink";
 import { type RouteInfo } from "./worker";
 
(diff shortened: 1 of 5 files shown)


src/stores.ts
```diff
@@ -64,3 +71,25 @@ export function deleteIntervention(id: number) {
...
   openFromSidebar.set(null);
 }
+
+function loadUserSettings(): UserSettings {
+  let settings = {
+    streetViewImagery: "google",
+    avoidDoublingBack: false,
+  };
+
+  // Be paranoid when loading from local storage, and only copy over valid items
```
dabreegster, line 81: Maybe this is overkill. As we evolve ATIP, these settings will change over time, but compatibility won't be that hard to handle. Just use new fields if things change in some crazy way. Think we should simplify it?

src/stores.ts
```diff
@@ -64,3 +71,25 @@ export function deleteIntervention(id: number) {
...
+  try {
+    let x = JSON.parse(window.localStorage.getItem("userSettings") || "{}");
+    if (x.streetViewImagery == "google" || x.streetViewImagery == "bing") {
```
Pete-Y-CS, line 84: Is there a way to this a bit more naturally? Create a `StreetViewProvider` type or similar? and then something like
`if (type x is == "StreetViewProvider")`

Pete-Y-CS: Yeah looks good to me!

Pete-Y-CS (approved):

PR #202, opened 2023-05-30
By default, don't keep appending points to the end of a route

dabreegster:
See https://github.com/dabreegster/route_snapper/pull/30 for context. I haven't done UX testing with anyone, but I suspect this'll make the default experience much less weird. It's much easier to drag intermediate waypoints this way.

For people who want to change the default, combined with #201 would let them make the setting sticky.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@types/geojson": "^7946.0.10",
         "comlink": "^4.4.1",
         "maplibre-gl": "^2.4.0",
-        "route-snapper": "^0.1.13",
+        "route-snapper": "^0.1.14",
         "svelte": "^3.54.0"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


dabreegster:
Preview at https://acteng.github.io/atip/extend_route/scheme.html?authority=Stoke-on-Trent

Pete-Y-CS:
> Preview at https://acteng.github.io/atip/extend_route/scheme.html?authority=Stoke-on-Trent

I just had a quick play and think this is really good but have a suggestion: it defaults to adding points to end, except when you 'edit geometry'.
Also maybe add points to start might be nice

Pete-Y-CS:
Looking at the code it looks like maybe my second request isn't available unless we control the route snapper package

dabreegster:
>  it defaults to adding points to end, except when you 'edit geometry'.

Wait, really? The "Add points to end" checkbox should be off by default and this should not happen.

> Also maybe a feature to add points to start might be nice

That could be nice. Even visually distinguishing the "last" point seems useful! These changes would go in https://github.com/dabreegster/route_snapper/, I'd say they're out of scope for Friday, but would like to think about how to implement them

Pete-Y-CS:
>  Wait, really? The "Add points to end" checkbox should be off by default and this should not happen.
I don't mean it does this I meant to say it might be a more expected way for it to behave: when making a new route it defaults to adding new waypoints, when editing a previous route it defaults to only allowing you to change existing waypoints

dabreegster:
> when making a new route it defaults to adding new waypoints, when editing a previous route it defaults to only allowing you to change existing waypoints

Ah, I see. This makes sense and may be a gentler transition for users used to the old behavior. Changed it to do that.

dabreegster:
It will make #201 confusing though -- if the user manually ticks the box, do we then override this automatic behavior? Maybe we just don't let them change that setting then?

Pete-Y-CS:
Yeah not sure what's best for this. Maybe we could have a default for each type of editing geometry: existing or new route, on a per user basis. I could also just see not having user preference here.

Sparrow0hawk (changes requested):
I find the UX of this a little tricky, I find it too easy to delete a red waypoint when trying to grab it and drag it.  I found myself accidentally deleting red waypoints when trying to click and drag them and then having no way to undo it which was frustrating.

Is it feasible to:

- Change the interaction that triggers a waypoint deletion i.e. right mouse click?
- Make a click on a waypoint make it sticky to the cursor, so I don't have to drag I can just click and move it about (realise this would depend on changing waypoint deletion event)

PR #209, opened 2023-06-01
Hide the autofill button unless we have a linestring

dabreegster:
Oops! Quick fix for this:
Screenshot from 2023-06-01 12-48-29

### DIFF
--- a/src/lib/forms/FormV1.svelte
+++ b/src/lib/forms/FormV1.svelte
@@ -30,9 +30,9 @@
 
 <label>
   Name:
-  <button type="button" on:click={() => autoFillName()} disabled={!$routeInfo}
-    >Auto-fill</button
-  >
+  {#if length_meters}
+    <button type="button" on:click={() => autoFillName()} disabled={!$routeInfo}>Auto-fill</button>
+  {/if}
   <br />
   <input type="text" bind:value={name} style="width: 100%" />
 </label>
--- a/src/lib/forms/FormV1.svelte
+++ b/src/lib/forms/FormV1.svelte
@@ -16,6 +16,8 @@
     return (x / 1000.0).toFixed(1) + "km";
   }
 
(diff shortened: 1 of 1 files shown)


src/lib/forms/FormV1.svelte
```diff
@@ -31,7 +31,9 @@
 
 <label>
   Name:
-  <button type="button" on:click={() => autoFillName()}>Auto-fill</button>
+  {#if length_meters}
```
Pete-Y-CS, line 34:
```suggestion
  {#if length_meters} // the presence of length_meters tells us this is a route, which means we can auto name it '{start location} to {finish location}'
```

This might be the rare case where I'll ask for a comment! Normally I actively avoid comments, but here I don't think it's worth the cost of having something more readable like:
``` 
{#if canBeAutoNamed(intervention)}

//where
canBeNamed = (feature) => {
  isRoute(feature);
}
```
In the future we could imagine `canBeNamed `being more complicated because we have a way to name a region based on neighbourhoods that are in it, or naming points simply after the road/roads they are on, etc.

Anyway long winded of saying I think we might want a comment explaining why `length_meters` means the intervention can be given a name... typing this out I realise that entities in ATIP aren't always interventions right? Could just be a point of interest, a neighbourhood that is marked for some reason... anyway I'm rambling

Pete-Y-CS (changes requested):

src/lib/forms/FormV1.svelte
dabreegster: Very good point! It's a little bit too magic that `length_meters` is the proxy for this. Added some comments.

Auto-naming points based on the nearest road or intersection would be cool. Detecting neighbourhood/area names would be harder (from the OpenStreetMap perspective, have to start processing place-name area relations), but sounds useful!

Pete-Y-CS (approved):

Issue #164, opened 2023-05-13
Add handling for waiting for Map load to test suite


Sparrow0hawk:
As per discussion and link [here](https://github.com/acteng/atip/pull/163#discussion_r1193023720), we need a nice way to handle waiting for the map to load before running tests that interact with the map.

@Sparrow0hawk TODO rest of this PR

Sparrow0hawk:
Test

PR #210, opened 2023-06-01
Add more Playwright tests

dabreegster:
See code comments for details. Biggest issues is that the tests are flaking on me locally, by timing out. #164 has ideas for more robustly waiting for maplibre to be ready, which could solve some problems. But even with proper waiting, running them is slow, because we use a totally new browser context for every single test. Nice for isolation, but slow. Pete and I are talking about trying [hooks](https://playwright.dev/docs/api/class-test#test-before-all) or fixtures to clear all state in between tests.


tests/modes.spec.ts
```diff
@@ -1,19 +1,80 @@
-import { readFile } from "fs/promises";
 import { test, expect } from "@playwright/test";
 
-test("clearing all while a feature is open works", async ({ page }) => {
-  await page.goto("/scheme.html?authority=North Somerset");
-  await page.getByRole("button", { name: "New point" }).click();
+async function clickMap(page, x, y) {
```
dabreegster, line 3: Just a helper to reduce visual line noise below

tests/modes.spec.ts
```diff
@@ -1,19 +1,80 @@
...
-    position: {
-      x: 500,
-      y: 500,
-    },
+    position: { x, y },
   });
+}
+
+test("clearing all while a feature is open works", async ({ page }) => {
+  await page.goto("/scheme.html?authority=Adur");
```
dabreegster, line 10: Switching to an authority with one of the smallest route files to load, in the hopes that it helps test speed

tests/modes.spec.ts
```diff
@@ -1,19 +1,80 @@
...
+  await clickMap(page, 500, 500);
   page.on("dialog", (dialog) => dialog.accept());
   await page.getByRole("button", { name: "Clear all" }).click();
   // Make sure AttributeMode resets to nothing being selected
   await expect(
     page.getByText("Click an object to fill out its attributes")
   ).toBeVisible();
 });
+
+test("creating a new point opens a form", async ({ page }) => {
```
dabreegster, line 21: This should catch the problems that we uncovered in #196

tests/modes.spec.ts
```diff
@@ -1,19 +1,80 @@
...
+    route.fulfill({
+      status: 404,
+    })
+  );
+  await page.goto("/scheme.html?authority=Adur");
+
+  await page.getByRole("button", { name: "New route" }).click();
+  await expect(page.getByText("Failed to load")).toBeVisible();
+
+  // Other tools should still work
```
dabreegster, line 76: @robinlovelace-ate, actually the point tool still works when route tool fails. Editing geometry might be broken though. I can't tell from your screenshot what happened exactly, but I'll keep playing with it and add tests to make sure other tools still work

tests/modes.spec.ts
```diff
@@ -1,19 +1,119 @@
...
+  await page.getByRole("button", { name: "New point" }).click();
+  await clickMap(page, 500, 500);
+  await page.getByRole("button", { name: "1) Untitled point" }).isVisible();
+});
+
+test("editing geometry of a polygon works", async ({ page }) => {
```
dabreegster, line 93: For #218, except the test passes! Maybe it's flaky and indicates a race condition

tests/modes.spec.ts
```diff
@@ -1,25 +1,132 @@
 import { readFile } from "fs/promises";
 import { test, expect, type Page } from "@playwright/test";
-import { loadInitialPage } from "./shared.ts";
+import {
+  loadInitialPage,
+  clickMap,
+  clearExistingInterventions,
+} from "./shared.ts";
```
dabreegster, line 7: Fun new config problem: `npm run check` says `Error: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.`

I tried removing the `.ts` to be consistent with the rest of our code, but then `npm run test` fails. Apparently a different tsconfig is being applied for tests?

tests/modes.spec.ts
```diff
@@ -1,25 +1,132 @@
...
 
 let page: Page;
 
 test.beforeAll(async ({ browser }) => {
   page = await loadInitialPage(browser);
 });
+test.beforeEach(async () => {
+  await clearExistingInterventions(page);
```
dabreegster, line 15: I forgot this at first and was very confused why tests failed together, but not individually!

tests/modes.spec.ts
```diff
@@ -1,25 +1,132 @@
...
+  await page.getByLabel("Description:").click();
+
+  // Edit geometry, and click the polygon again
+  await page.getByRole("button", { name: "Edit geometry" }).click();
+  await clickMap(page, 312, 501);
+  // This button indicates the mode is working
+  await page.getByRole("button", { name: "Finish" }).click();
+});
+
+test("testing adding interventions, then deleting one, then adding another", async () => {
```
dabreegster, line 104: These two came from the deleted files. Moved them here for better grouping of mode/drawing tests, and to share the worker

tests/modes.spec.ts
robinlovelace-ate: Will test...

robinlovelace-ate: I'm pretty sure I could not add a point or anything to download as a GeoJSON in the accidental test with no network.

dabreegster: https://acteng.github.io/atip/fail_route_load/scheme.html?authority=Adur
Give it ~10 mins to deploy, then we can play with this. It's set up to grab the wrong route files and fail, so we can see everything that breaks. I added a test (now in `errors.spec.ts` that demonstrates making a new point still works, at least)

robinlovelace-ate: @dabreegster cannot see how to add anything here.. https://acteng.github.io/atip/fail_route_load/scheme.html?authority=Adur#5.74/52.469/-1.707

dabreegster: Ahh. "New point" and freehand polygon work fine. But as soon as you try to click "New route", everything gets stuck. Thanks, we can update the tests accordingly and fix this bug

dabreegster:
Oops, GH didn't add reviewers -- @Pete-Y-CS, how do these look to you? Rebased after the new hooks

tests/modes.spec.ts
Pete-Y-CS: I'm not sure, I do remember   having some similar problem and just doing what got the tests to run and forgetting about it which might not be the best policy. I don't mind permitting importing with extensions but would be ideal to understand why this needs it and see if we can avoid

Pete-Y-CS (approved):

dabreegster:
Thanks for the review! I'll merge this in now, so we can have some coverage for #220

PR #212, opened 2023-06-01
Share loading of page between multiple tests

Pete-Y-CS:
use worker-scoped fixture to setup page at start for every worker this makes some tests slower and fail because of it but it should front load loading of map editor for at least `savefiles.spec`
Needs somet thought about how to isolate only for tests that need it


tests/savefiles.spec.ts
```diff
@@ -1,66 +1,98 @@
...
 // Handle unusual GeoJSON inputs produced by other tools
 test("loading a file with null properties displays the length", async ({
   page,
 }) => {
   // Remove the property from the test data first
-  let json = JSON.parse(
-    await readFile("tests/data/route.json", { encoding: "utf8" })
-  );
+  const properties = json.features[0].properties;
   json.features[0].properties = null;
```
dabreegster, line 78: This mutates the shared `json`. I think we have to be paranoid and make a deep copy in here. `JSON.parse(JSON.stringify(json))` is the hack I've been using everywhere, but supposedly https://web.dev/structured-clone/ is the new way

Pete-Y-CS: ahh a classic

Pete-Y-CS:
@dabreegster @Sparrow0hawk  can we prioritise merging this since it makes the tests more reliable as far as I can tell? I'd like to rebase my state based bugs changes on top of this

dabreegster:
Reviewing now. I just did `npm run fmt` and pushed to this branch again

tests/add-delete-add.spec.ts
```diff
@@ -1,9 +1,18 @@
-import { test, expect } from "@playwright/test";
+import { test, expect, type Page } from "@playwright/test";
 
-test("testing adding interventions, then deleting one, then adding another", async ({
-  page,
-}) => {
-  await page.goto("/scheme.html?authority=Derby#16.84/52.906457/-1.504519");
+let page: Page;
+
+test.beforeAll(async ({ browser }) => {
```
dabreegster, line 5: Is there a way we can share this `beforeAll` code over multiple files? Not sharing workers between files, just deduping the code. If we have a `shared.ts` file and import from it, does that work? On the "fixtures" docs, they say

> Fixtures are reusable between test files - you can define them once and use in all your tests. That's how Playwright's built-in page fixture works.

Which makes me worry this is somehow not straightforward for hooks.

But this shouldn't block merging this PR, since it's an improvement in stability

tests/add-delete-add.spec.ts
```diff
@@ -1,9 +1,18 @@
...
+  await page.goto("/scheme.html?authority=Adur");
+  // wait for the map to load and interventions panel to appear
+  await page.getByText("Click an object to fill out its attributes").waitFor();
+  // wait for router snapper to load so we can use route tool
+  await page.getByText("Route tool loading...").waitFor({ state: "hidden" });
+});
+
+test("testing adding interventions, then deleting one, then adding another", async () => {
+  await page.goto("/scheme.html?authority=Adur");
   // wait for the map to load and interventions panel to appear
```
dabreegster, line 16: Some of these steps in the test are now redundant with `beforeAll`

tests/add-route.spec.ts
```diff
@@ -1,7 +1,17 @@
...
+  page = await browser.newPage();
+  await page.goto("/scheme.html?authority=Adur");
+  // wait for the map to load and interventions panel to appear
+  await page.getByText("Click an object to fill out its attributes").waitFor();
+  // wait for router snapper to load so we can use route tool
+  await page.getByText("Route tool loading...").waitFor({ state: "hidden" });
+});
+
+test("testing add a route and save it", async () => {
   // wait for the map to load and interventions panel to appear
```
dabreegster, line 15: Same -- we can delete some of these now

tests/savefiles.spec.ts
```diff
@@ -1,54 +1,87 @@
...
+  await page.getByText("Click an object to fill out its attributes").waitFor();
+  // wait for router snapper to load so we can use route tool
+  await page.getByText("Route tool loading...").waitFor({ state: "hidden" });
+
+  json = JSON.parse(
+    await readFile("tests/data/Adur.json", { encoding: "utf8" })
+  );
+});
+
+test.beforeEach(async () => {
```
dabreegster, line 20: I suspect this'll also be useful for other files, so if we can have a `common.ts` or similar and import, that feels ideal

dabreegster (commented):
Only blocking feedback is cleaning up some redundant checks in the tests. Ideally we can share some code, but no need to solve that now.

Other thing to keep in mind with common hooks is that some tests will be a little special and work across multiple pages -- starting on the homepage, loading a file, and winding up on scheme.html. So most tests will probably want to use the shared `beforeAll`, but not all

dabreegster:
I merged the file loader PR with a new test: https://github.com/acteng/atip/blob/45712dcead6238f111cbebc9337baf4e469fad1f/tests/savefiles.spec.ts#L78
This one is in `savefiles.spec.ts`, but it uses two pages. If it makes more sense to split that to a different file that doesn't use the hooks, feel free to do that

tests/add-delete-add.spec.ts
dabreegster: This looks fine to me, and is nice that it's explicit we're using `beforeAll` in each test file. Thanks!

dabreegster (approved):
Looks good to me, after the rebase. Awesome!

Pete-Y-CS:
I'm a little confused about the situation with the branch: when I look to rebase onto main there are some commits which I'm not sure why they're on this branch but not main:
image
Specifically the first 4 are a bit  unexpected to me

Issue #142, opened 2023-04-26
State changing bugs


dabreegster:
Ideally we get tests for these, but that's an upfront investment in figuring out how to mock some of the interactions.

- [ ] While in edit geometry mode, click "Clear all"
- [ ] While in edit geometry mode, load another geojson file
- [ ] While a form is open, clear all. Hover briefly breaks

I can't reproduce it, but "split route" and "edit geometry" on an existing scheme were warping to the object when clicked, as if "edit attributes" was partially active

Pete-Y-CS:
Hey, looking to investigate this while I wait for Docker solutions: I suppose our expected behaviour is to disable buttons/actions that don't make sense according to the state we're in?

So for example for
> While in edit geometry mode, click "Clear all"
Would we like to disable the clear all button, or have it clear what we're working on?

Likewise I guess for this:
> While in edit geometry mode, load another geojson file
My instinct would be disabling the feature.

Not sure what is meant by:
> While a form is open, clear all. Hover briefly breaks
&
> I can't reproduce it, but "split route" and "edit geometry" on an existing scheme were warping to the object when clicked, as if "edit attributes" was partially active

Cheers

dabreegster:
> While in edit geometry mode, click "Clear all"
> Would we like to disable the clear all button, or have it clear what we're working on?

I'm not sure what behavior we should go for. Right now, it breaks `GeometryMode.svelte` with `feature is undefined`, because suddenly we're editing geometry of a feature ID that's vanished. "Clear all" is meant to totally erase the entire loaded file. So either we disable the button in certain modes, or when we do clear everything, we reset the drawing mode to "Edit attributes". Which do you think would be more intuitive to the user?

> While a form is open, clear all. Hover briefly breaks

I think I actually fixed that one in https://github.com/acteng/atip/commit/4534c3a5f7113bbfff8ee68a68d4e3d2b81651a8 (and with `tests/modes.spec.ts`!). The outline effect from `HoverLayer.svelte` was sticking around after clearing, until clicking the map again.

> I can't reproduce it, but "split route" and "edit geometry" on an existing scheme were warping to the object when clicked, as if "edit attributes" was partially active

Sometimes when I activate "split route" and click a route, it would open the form for that route and warp to it -- as if "edit attributes" was active. I haven't been able to reliably repro this one. I _think_ I've only seen it when developing locally, and I suspect it's related to #174

dabreegster:
[...]

PR #213, opened 2023-06-01
Prevent the page from breaking when swapping files in the middle of e…

dabreegster:
…diting geometry

See https://github.com/acteng/atip/pull/196#issuecomment-1572313197 for context. I'm proposing this as a short-term fix for #142 that we can deploy in v2.

### DIFF
--- a/src/lib/draw/GeometryMode.svelte
+++ b/src/lib/draw/GeometryMode.svelte
@@ -66,6 +66,13 @@
     if (mode == thisMode) {
       gjScheme.update((gj) => {
         let feature = gj.features.find((f) => f.id == currentlyEditing)!;
+        // TODO Hack around https://github.com/acteng/atip/issues/142
+        if (!feature) {
+          window.alert(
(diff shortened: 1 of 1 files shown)


src/lib/draw/GeometryMode.svelte
```diff
@@ -66,6 +66,13 @@
     if (mode == thisMode) {
       gjScheme.update((gj) => {
         let feature = gj.features.find((f) => f.id == currentlyEditing)!;
```
dabreegster, line 68: Note that if the user loads a different file, and the arbitrary numeric feature ID happens to line up, then we'll silently replace geometry for the wrong thing. It might even change the type from a route to an area, or something silly like that. So, this fix is very much a hack, but it hopefully stops the most common error case from happening, and is better than a silent error showing up in console logs and the page being broken

Sparrow0hawk (approved):
Happy to merge, works as expected.

Pete-Y-CS (approved):

PR #216, opened 2023-06-02
Use numbered versions for data files

dabreegster:
I propose we change how we version the two sets of binary route data files. Previously I've been copying the same data between a dev (for the auto-deployed github version) and prod (for atip.uk), but there's an inherent awkwardness when updating the dev data format in sync with the code change. No reason for that complication; we can just put an arbitrary version number for the data files, and pin the code to that version.

I've used "v2" here for the data file, and don't really care how we increment this later (v2.1, v3, etc) -- as long as the strings are unique and we update the URL here, it'll be fine.

I haven't copied the files in S3 to this new URL directory structure yet; will do so once we agree on the naming of it. And will update abstreet-to-atip scripts too.

### DIFF
--- a/src/pages/App.svelte
+++ b/src/pages/App.svelte
@@ -27,27 +27,17 @@
   let showAbout = false;
   let showInstructions = false;
 
-  // TODO This is a hacky way of detecting environment. 'npm run dev' and
-  // Github should be false, and only Cloudfront should be true.
-  let prod = window.location.hostname.includes("atip.uk");
-
   const params = new URLSearchParams(window.location.search);
   // TODO Add validation and some kind of error page
   let authorityName: string = params.get("authority")!;
   let style: string = params.get("style") || "streets";
   let schema: Schema = (params.get("schema") as Schema) || "v1";
 
-  // TODO Slight hack. These files are stored in an S3 bucket, which only has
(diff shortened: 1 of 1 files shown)


src/pages/App.svelte
```diff
@@ -27,27 +27,17 @@
   let showAbout = false;
   let showInstructions = false;
 
-  // TODO This is a hacky way of detecting environment. 'npm run dev' and
```
dabreegster, line 30: It's also nice to get rid of this distinction. This code path isn't testable locally, so that's a bit scary anyway!

Sparrow0hawk (approved):
Happy with this, think moving to MAJOR.MINOR versioning is absolutely fine for the data.

dabreegster:
Copying the data over (minus the dry flag):
```
aws s3 sync --dry --content-encoding="gzip" s3://atip.uk/route-info/ s3://atip.uk/route-info/v2/
aws s3 sync --dry --content-encoding="gzip" s3://atip.uk/route-snappers/ s3://atip.uk/route-snappers/v2/
```

Pete-Y-CS (approved):
A rare situation where more red than green is good 😅

dabreegster:
Tests pass again after copying to the new directory, so merging! Thanks for reviews

PR #219, opened 2023-06-03
Load an ATIP file from the homepage

### LINKED ISSUE #217: Tool or page to open an ATIP file
Say I've got a .txt file locally (or even by URL) like https://github.com/acteng/atip/issues/215#issuecomment-1573302721. I want to open in ATIP, but I have to first open the editor page for the correct authority. I also potentially need to change the schema. Both of those things are clearly visible in the JSON file itself. Could we have a file uploader on the front page that just does all of this automatically?

The hard part implementation-wise is to pass in the GeoJSON data itself. If we have it in-memory in the browser on one page, can we "forward" it easily to another page? Are session cookies the right approach here?

It'd also be useful for #173. From a main page with potentially lots of schemes loaded, jump to the editor page with only one scheme loaded.

dabreegster:
Closes #217. Conveniently pick a file and jump to the right page with it loaded:

https://github.com/acteng/atip/assets/1664407/d3196682-0450-4d07-ae86-658ca8563f3c

### DIFF
--- a/src/lib/common/FileInput.svelte
+++ b/src/lib/common/FileInput.svelte
@@ -2,12 +2,19 @@
   export let label: string;
   export let onChange: (e: Event) => void;
 
+  // TODO Only one component per page is supported, because the ID is
+  // hardcoded. This is only necessary for the Playwright test.
   let fileInput: HTMLInputElement;
 </script>
 
 <!-- TODO Interactive elements inside a label are apparently invalid, but this works -->
 <label>
-  <input type="file" bind:this={fileInput} on:change={onChange} />
+  <input
+    type="file"
+    id="file-input"
+    bind:this={fileInput}
+    on:change={onChange}
+  />
   <button type="button" on:click={() => fileInput.click()}>{label}</button>
 </label>
 
--- a/src/lib/common/FileInput.svelte
+++ b/src/lib/common/FileInput.svelte
@@ -1,0 +1,26 @@
+<script lang="ts">
+  export let label: string;
+  export let onChange: (e: Event) => void;
+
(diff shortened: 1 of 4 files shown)


src/lib/common/FileInput.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+  export let label: string;
+  export let onChange: (e: Event) => void;
+
+  // TODO Only one component per page is supported, because the ID is
+  // hardcoded. This is only necessary for the Playwright test.
+  let fileInput: HTMLInputElement;
+</script>
+
+<!-- TODO Interactive elements inside a label are apparently invalid, but this works -->
```
dabreegster, line 10: This is an old hack, probably from Stackoverflow, to style the file input box reasonably

robinlovelace-ate (approved):
Big 👍 to this new functionality. Great work Dustin!

src/pages/ChooseArea.svelte
```diff
@@ -89,9 +92,50 @@
         window.location.href = `scheme.html?authority=${name}`;
       });
     });
+  });
+
+  function loadFile(e: Event) {
```
Pete-Y-CS, line 97: Can we share this in the `FileInput.svelte`? feels like a lot of this logic must be shared between the two locations

Pete-Y-CS (approved):

src/pages/ChooseArea.svelte
dabreegster: Done. It's unfortunate that `FileReader` uses callback APIs and not async/await

Pete-Y-CS (approved):

PR #220, opened 2023-06-05
Use current mode from a global store

Pete-Y-CS:


dabreegster (commented):
Code changes here LGTM. Let's get #212 and #210 in first to have trustworthy tests verify that just this refactor doesn't change any behavior. (with that "new point" bug)

Pete-Y-CS:
Alright, it is in a somewhat reviewable state @dabreegster @Sparrow0hawk ! I have discovered edit geometry isn't working as I'd hoped, yet. Relatively confident it's just some more plumbing to do but most tools are working as expected as far as I can tell so I think it's a good time for you guys to check how it's working. I've spoken to Alex in person about how we're doing it but if either of you have Qs throw em my way or ask for a call!

src/default_event_handler.ts
```diff
@@ -0,0 +1,38 @@
...
+      click: () => {},
+      dblclick: () => {},
+      mousemove: () => {},
+      mousedown: () => {},
+      mouseup: () => {},
+      mouseout: () => {},
+      dragstart: () => {},
+    };
+    this.documentHandlers = {
+      click: () => {},
```
dabreegster, line 15: Since we don't have any uses for this yet an it has confusing overlap with the map one, can we remove till we need it?

src/default_event_handler.ts
```diff
@@ -0,0 +1,38 @@
+import type { EventHandler } from "./types";
+
+export class DefaultEventHandler implements EventHandler {
+  constructor() {
+    this.mapHandlers = {
+      click: () => {},
```
dabreegster, line 6: Some of these take event params with a certain type, how hard is it to add those here? I haven't played with `npm run check` yet to see what it thinks

src/lib/draw/AttributeMode.svelte
```diff
@@ -60,44 +49,43 @@
...
+    });
+    if (results.length > 0) {
+      formOpen.set(results[0].id as number);
+    } else {
+      formOpen.set(null);
     }
-  }
+  };
+
+  eventHandler.mapHandlers.mousemove = onMouseMove;
```
dabreegster, line 82: I haven't tried yet, and it's not very important, but I wonder if we can't just do this directly and not need the intermediate `const`. From my understanding of closures in JS, it should be equivalent, with less boilerplate

src/lib/draw/GeometryMode.svelte
```diff
@@ -170,16 +164,16 @@
...
+    if (currentlyEditing == null) {
       mapHover.set(null);
     }
-  }
+  };
 
-  function onClick(e: MapMouseEvent) {
-    if (mode == thisMode && currentlyEditing == null) {
+  const onClick = (e: MapMouseEvent) => {
+    if (currentlyEditing == null) {
```
dabreegster, line 176: GeometryMode is the special one. If we're currently editing something, we need to forward events to the right place. So here for example, we need something like

```
if (currentlyEditing == null) {
  ...
} else {
  if (currentlyEditingControls == "point") {
    pointTool.onClick(e);
  } else if (currentlyEditingControls == "snap-polygon" || "route") {
    routeTool.onClick(e);
  }
  ...
}
```

and it'd be most bugproof and least boilerplatey to forward _all_ events to the appropriate place. Not sure yet how to do that nicely though.

src/lib/draw/Toolbox.svelte
```diff
@@ -46,6 +49,35 @@
...
+    "edit-geometry": new DefaultEventHandler(),
+    route: new DefaultEventHandler(),
+    point: new DefaultEventHandler(),
+    "free-polygon": new DefaultEventHandler(),
+    "snap-polygon": new DefaultEventHandler(),
+    "split-route": new DefaultEventHandler(),
+    "street-view": new DefaultEventHandler(),
+  };
+
+  MapEvents.forEach((eventName) => {
```
dabreegster, line 63: To let hot-module reloading work, we need to remove all these handlers in `onDestroy`. I _think_ to do that, we have to be careful about passing in exactly the same closure here when we do `map.off`. Not sure yet about the nicest way to do that

src/lib/draw/Toolbox.svelte
```diff
@@ -46,6 +49,35 @@
...
+    $map.on(eventName, (event) => {
+      console.log(
+        `map event happening ${$currentMode}, ${get(currentMode)}, ${eventName}`
+      );
+      eventHandlers[get(currentMode)].mapHandlers[eventName](event);
+    });
+  });
+
+  DocumentEvents.forEach((eventName) => {
+    $map.on(eventName, (event) => {
```
dabreegster, line 73: !!!
These need to be `document.addEventListener`, not map

src/lib/draw/Toolbox.svelte
```diff
@@ -79,18 +112,20 @@
...
       thisMode="edit-attribute"
       label="Edit attributes"
       icon={editAttributesIcon}
       {changeMode}
     />
-    <AttributeMode bind:this={attributeMode} {mode} {changeMode} />
+    <AttributeMode
+      bind:this={attributeMode}
+      {changeMode}
+      bind:eventHandler={eventHandlers["edit-attribute"]}
```
dabreegster, line 124: https://svelte.dev/tutorial/text-inputs
I think `bind:` is only if `AttributeMode` was overwriting the thing passed in and we want it to flow back up. It's not; it's just mutating it. The pointer to `eventHandlers["edit-attribute"]` shouldn't change, so I don't think we need it

src/lib/draw/point/point_tool.ts
```diff
@@ -19,7 +18,22 @@ export class PointTool {
   eventListenersFailure: (() => void)[];
   cursor: FeatureWithProps<Point> | null;
 
-  events: EventManager;
+  onMouseMove = (e: MapMouseEvent) => {
+    if (this.active) {
```
dabreegster, line 22: I think we can remove the `this.active` checks now? We're only forwarding events to the tools when they're in use

src/lib/draw/route/route_tool.ts
```diff
@@ -341,6 +328,18 @@ export class RouteTool {
     this.redraw();
   }
 
+  setHandlers = (eventHandler: EventHandler) => {
```
dabreegster, line 331: This is from an older commit and not the new way of doing things in this PR, right? The way point mode/tool works looks fine, this is different...

src/types.ts
```diff
@@ -70,3 +70,47 @@ export interface UserSettings {
...
+  | "edit-attribute"
+  | "edit-geometry"
+  | "route"
+  | "point"
+  | "free-polygon"
+  | "snap-polygon"
+  | "split-route"
+  | "street-view";
+
+export const MapEvents = [
```
dabreegster, line 84: All of this new stuff could maybe stay scoped to `lib/draw`. It's specific to everything managed by Toolbox. I don't think we'll want to do the same event management elsewhere in ATIP, but kind of unknown till we explore contextual layer stuff more

src/types.ts
```diff
@@ -70,3 +70,47 @@ export interface UserSettings {
...
+  "dragstart",
+];
+
+export type MapEvent = (typeof MapEvents)[number];
+
+export const DocumentEvents = ["click", "keyPress", "keyUp", "keyDown"];
+
+export type DocumentEvent = (typeof DocumentEvents)[number];
+
+export interface EventHandler {
```
dabreegster, line 100: Do we need the interface separate from the base class? I think we could remove this and just use the class everywhere, including for the type when passing stuff into each mode

src/lib/draw/Toolbox.svelte
Pete-Y-CS: Ahhhh good point, I did at some point think I needed to do map.off, didn't consider the closure and forgot to do it thought 😅

src/lib/draw/point/point_tool.ts
Pete-Y-CS: We shall see. Currently geometry mode is misbehaving weirdly so I don't wanna jump the gun on this. I've gotten rid of a lot of `currentMode` checks but I think geom mode means `this.active` checks probably wanna stay

src/lib/draw/route/route_tool.ts
Pete-Y-CS: there was a reason I did it this way, and I have since made it work this way for Point Mode, so no this isn't a leftover from another commit. It might be less nice to do it this way. I don't the remember the reason I did this, it may have simply been that I thought it was nicer, but in any case it can't work this way for GeomMode so we may need to rethink or just let GeomMode be an exception

src/types.ts
Pete-Y-CS: agreed, had this thought about the `eventHandler` types and `default_event_handler`

src/types.ts
Pete-Y-CS: Most likely

src/lib/draw/point/point_tool.ts
dabreegster: I might've been wrong here! We still need the currentMode checks in the Svelte components

dabreegster:
I pushed 3 commits, all showing up at https://github.com/acteng/atip/tree/current-mode-in-global-store. For some reason the PR view here is only showing 2 of them for me.

But anyway, I think with these fixes, this PR is working, including geometry mode. The problem was for `addEventListenerSuccess` and the others, all the custom event handlers to figure out when point/polygon/route tool has succeeded at something or failed. We need to keep the `currentMode == thisMode` guard in there. Otherwise from geometry mode, we finish with point tool by clicking, but `PointMode.svelte`'s callback incorrectly fires. I think I misled you on a comment about removing "am I active now?" checks, sorry!

dabreegster:
Thanks for checking. All 12 commits are now correctly showing up in GH; must've been some partial outage or delays yesterday.

src/lib/draw/AttributeMode.svelte
Pete-Y-CS: I wasn't able to see your think about preference between `function foo (bar){ ` and `const foo = (bar) =>{` but I don't really see one as preferable to the other. I guess consistency is preferable. I will try switch and see if it works

dabreegster: Arrow functions are fine, I just meant could we do `eventHandler.mapHandlers.mousemove = (e: MapMouseEvent) => { ... }` and skip the intermediate const

src/lib/draw/StreetViewMode.svelte
```diff
@@ -92,15 +80,16 @@
   }
 
   // The escape key isn't registered at all for keypress, so use keydown
+  // TODO This doesn't use eventHandler, so it might run even when the mode isn't active.
   function onKeyDown(e: KeyboardEvent) {
```
dabreegster, line 84: This isn't using the new document event handler construct at all; it's bound using `svelte:window` below. Do we want to change this? If we leave it like it is now, the mode check is important. Guessing we want to migrate this to the new eventHandler though!

dabreegster:
Some code review has been happening in DMs on Teams, so summarizing status here. We should fix a few things before merging:

- rebasing (there shouldn't be any conflicts, and it'll bring in new tests that catch the original bug found here) 
- using `document.addEventListener` for document events
- checking the key handler in StreetviewMode, since it doesn't use the new `EventHandler` right now
- unregistering map and document handlers in Toolbox when the component goes away (hard to test automatically, and less critical -- only affects local dev workflow)

There's some other cleanup that might be possible and useful, but it shouldn't block merging. After this is merged, we can go back to the original feature request in #196 and disable some sidebar actions when we're in drawing modes.

Let me know if you want help with any of this!

dabreegster (approved):
OK, I vote merge this (via rebasing to preserve commits). There are some smaller followup things we can do, but easier to split out into smaller next steps. I'll file a bug momentarily

Issue #203, opened 2023-05-31
Standardize code formatting


dabreegster:
Can we make VSCode use `prettier` under the hood, or make the prettier config match what VSCode does?

And can we auto-sort / group / clean up imports? Quick search shows https://github.com/trivago/prettier-plugin-sort-imports

dabreegster:
Using a plugin for imports now, so I'll close this issue. If anyone's using VS Code and it's not matching `npm run fmt`, we can figure out how and add some developer docs.

PR #222, opened 2023-06-05
Use a prettier plugin to auto-format imports. #203

dabreegster:


.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
```
dabreegster, line 1: There are more non-default options in https://www.npmjs.com/package/@ianvs/prettier-plugin-sort-imports#usage, but this seems to be the minimal to get the plugin to do something. I'm totally unopinionated on the actual style we use, so happy to configure anything anybody wants here. I just want automation, so I can stop thinking about it.

package.json
```diff
@@ -15,6 +15,7 @@
     "generate-schema-ts": "for x in v2 planning criticals; do npx ts-node --project tsconfig_tsnode.json --esm src/lib/forms/generate_ts.ts src/schemas/$x.json > src/schemas/$x.ts; done"
   },
   "devDependencies": {
+    "@ianvs/prettier-plugin-sort-imports": "^4.0.2",
```
dabreegster, line 18: This is https://github.com/ianvs/prettier-plugin-sort-imports. I chose it over https://github.com/trivago/prettier-plugin-sort-imports because it's more recently maintained, builds on the other repo, and seems actively used by people

dabreegster (commented):
`npm run fmt` is still all it takes to run this

Pete-Y-CS (approved):

dabreegster:
Rebased, now merging. No more worrying about this manually!

Issue #173, opened 2023-05-17
Create internal scheme overview page


dabreegster:
Internally, people are currently using an R-generated Leaflet viewer to look at all ATF4 schemes everywhere. There's been a request to be able to search by scheme name.

I actually didn't realize people are using the Leaflet viewer so much; I think we should move that functionality into ATIP proper to control it better. I propose a new page that has a map and a sidebar, but no drawing tools. It displays the consolidated GeoJSON file with simple styling, lists schemes (not individual interventions!) in the left sidebar, has a text search for scheme name, and makes it easy to click on a scheme and open ATIP edit mode with that file.

Biggest open question in my mind is how to give people the all-schemes GeoJSON file. Simplest answer is just to have a regular file loader on that page, send the big geojson file around using onedrive/sharepoint/whatever other internal thing.

Other open question is how to conveniently open ATIP with just that one scheme file after clicking.

robinlovelace-ate:
+1 to that. What would the input into this new page be? Currently we have

- A folder with sub-folders containing ATIPs
- A zip file
- GeoJSON files created by R/Python scripts that read-in the ATIP data

robinlovelace-ate:
Could there be a dedicated page for internal use with "load directory" that iterates through all .txt/.geojson files or am I missing something?

dabreegster:
> +1 to that. What would the input into this new page be? Currently we have

I don't think it's easy to recursively read a bunch of files from a picker in the browser. Either one of the consolidated files (`atips_all_raw.geojson` or something else) or a zip, I'd say. Probably the first is easier, but it maybe depends what we want to do besides just displaying everything in one map.

> Could there be a dedicated page for internal use with "load directory" that iterates through all .txt/.geojson files or am I missing something?

That's what I'm suggesting. It doesn't need to be internal or anything, just a new page where you manually load a file representing a bunch of ATIP files all smushed together. The different schemes are somehow grouped/identified -- `atip_analysis` currently does it by adding some properties to every feature. We could also nest `FeatureCollections`. This is allowed by the GeoJSON schema, but possibly the support for it in MapLibre or other standard tools is less good, so maybe we avoid that. Not sure yet.

robinlovelace-ate:
Sounds good. I think if the input can be the raw ATIPs (or other valid GeoJSON files without worrying much about attributes) that could be v. helpful.

PR #225, opened 2023-06-07
Improve scheme browse page

dabreegster:
#173 

This now takes `magicfoo.geojson` as input -- the previous input files will _not_ work. And the tooltips are formatted better.


Pete-Y-CS (approved):

dabreegster:
Thanks for the review! Merging, then I'll update the email with inspectors with the new input file

robinlovelace-ate:
Great, I'll do a live demo today that will benefit from this!

PR #229, opened 2023-06-08
Start a design doc for an ATIP backend + database

dabreegster:
This is more like a set of questions we need to start exploring, rather than a design. Happy to try other formats for this (like Google Docs with comments), but I think it could be helpful to document the architectural decisions we make as the project evolves.

There are _tons_ of questions raised, please comment!

### DIFF
--- a/docs/backend.md
+++ b/docs/backend.md
@@ -1,0 +1,138 @@
+# Backend design doc
+
+This is a design doc proposing some possible approaches for an ATIP backend. It doesn't represent any final decisions or describe anything actually built.
+
+As of June 2023, ATIP is a web app for editing GeoJSON data. Those GeoJSON files are stored in each user's browser local storage and manually saved/loaded as local files. There is no central place to submit and store files, or any concept around tracking changes to a file.
+
+Terminology:
+
+- a **scheme** is a single ATIP GeoJSON file, with a list of **interventions** or **features** (these are used interchangeably)
+- **ATE** is the central group evaluating and inspecting schemes
+- **LAs** (local authorities) are the groups submitting schemes for some area. (The exact geographic scope doesn't matter; these could be transport or highway authorities or something else)
+
+## Requirements
+
+The primary goal of this doc is to facilitate ATE and LAs collaborating on schemes. An example interaction might be:
+
+1.  User1 from an LA submits a scheme and supplementary PDF files
+2.  User2 from ATE inspections comments on the scheme
+3.  User1 responds to comments, then submits a 2nd version of the scheme
+4.  User2 reviews the changes (viewable as a diff and with the existing regular ATIP tool), then resolves the comment
+
+There are other tasks for which ATIP may later need a backend, like synchronizing data with OpenStreetMap. These are out-of-scope.
+
+### User stories
+
+- as a LA
+	- MVP: submit a scheme in various stages (for initial feedback or to request funding)
+	- MVP: submit new versions of an existing scheme
+	- MVP: respond to feedback
+	- MVP: see a timeline view of all interactions with ATE, including each version of a scheme
+	- out-of-scope: search or browse schemes by other LAs, to learn about best practices or other approaches to some common situation
+	- out-of-scope: share/publish scheme data publicly, as part of public consultation
+- as ATE inspections
+	- MVP: search or browse schemes across all LAs
+	- MVP: give or respond to feedback to a scheme, whether it's proposed or already built
+	- out-of-scope: assign particular inspectors to a scheme or track outstanding tasks
+- as ATE data analysts
+	- MVP: access all submitted schemes in bulk for offline analysis
+
+## Data & storage
+
+Not being prescriptive about a specific database yet, but imagine we're doing something relational. We need to decide roughly what rows, or keys in a blob store, will contain.
+
+### Data model for schemes
+
+A scheme is not submitted once and never again edited; it has a full lifecycle from initial proposal, iteration with feedback, being funded, being constructed, and being inspected once built. We have to decide what level of granularity to store schemes.
+
+The most coarse-grained approach is to store the entire GeoJSON blob per scheme and version. In order to understand how a scheme changes between versions, we would need to calculate and visualize a diff from the raw list of interventions.
+
+Per scheme, we could instead store individual edit commands, allowing for more granular undo/redo. A command might be something like `delete feature 5` or `modify feature 7 with geometry X and properties Y`. The user could undo/redo through this detailed command list, and also explicitly "save" a named version in this command history, and submit that for review.
+
+Some open questions about schemes:
+
+- How "big" is it? For proposals, the meaningful unit could be the set of interventions for which funding is requested.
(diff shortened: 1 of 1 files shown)


docs/backend.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Syncing
+
+Suppose somebody is editing an ATIP scheme. How often should the frontend save the changes to the cloud? Should this only happen when the user explicitly saves and submits a version of the scheme? Should we log every command/edit and stream those to a database with something like websockets?
+
+How should concurrent edits work? We do **not** want to take on the immense technical and design complexity of an app with real-time collaboration, like Google Docs or Felt or Figma, without very good reason. Can we ensure only one person from an LA is submitting a new version of some scheme at a time? Should we surface something in the UI to show the user currently editing a scheme? Can we come up with a robust UI and algorithm to diff two schemes and guide the user through resolving any conflicts, and make the user submitting a scheme second deal with any conflicts?
```
Sparrow0hawk, line 134: I think we'd aim to prevent concurrent edits. 

How we handle it i'm unsure, it feels like something that might require a more complex backend, to actually handle what files have been requested/loaded so that you could lock editing of a given file.

docs/backend.md
```diff
@@ -0,0 +1,138 @@
...
+
+Similarly, whatever backend we develop will inevitably be coupled to ATE use cases, scheme metadata, etc. Making it generic to allow another org to someday deploy it is not high priority. All the same, we should future-proof this from the start by keeping all the backend code open source from the start, using Docker and Terraform and similar to provision cloud resources based on config that somebody could easily adjust. There will be exceptions (like GDS auth integration), but we should try to keep things reasonably modular.
+
+### Syncing
+
+Suppose somebody is editing an ATIP scheme. How often should the frontend save the changes to the cloud? Should this only happen when the user explicitly saves and submits a version of the scheme? Should we log every command/edit and stream those to a database with something like websockets?
```
Sparrow0hawk, line 132: I need to think about this one. 

It feels like only updating the database would be when the user explicitly saves/submits but it might be nice to log commands/edits so that progress is saved without submitting.

Sparrow0hawk (commented):
Thanks for this @dabreegster have smatters thoughts throughout.

docs/backend.md
```diff
@@ -0,0 +1,138 @@
...
+
+## API
+
+The data model and storage are likely the difficult parts of this design; the API and server implementing it should be much simpler. They can just be a thin **stateless** wrapper, performing authentication and then interacting with the database. Thus, they could be serverless, since the start-up time should be trivial.
+
+We should check if GDS prefers any particular languages or frameworks. Some options that some/all of the team is familiar with:
+
+- Go
+- Python (Flask?) -- is the GIL a problem?
+- Rust (overkill, Dustin not in favor)
```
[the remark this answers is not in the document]
dabreegster: I guess it's not a strong / well-informed opinion yet. I haven't looked through the ecosystem lately, but https://www.arewewebyet.org/ seems reassuring. Async can be one of the trickier parts of Rust, and compile times are potentially a bit annoying for large projects with many dependencies to link -- but probably not going to be us. Rust could be fine. Maybe once we know what the auth layer will look like, we can check around for client libraries in languages we're considering

docs/backend.md
```diff
@@ -0,0 +1,138 @@
...
+- get one scheme + version
+- get the latest version of a scheme
+- update one scheme
+- get all schemes combined (possibly filtering the results)
+
+### Svelte / UI changes
+
+We need to integrate whatever GDS auth layer. Talking to the backend may be trivial (`fetch` calls), or maybe we use something like websockets for saving commands while people edit.
+
+We should be very careful to keep the existing ATIP pages (browsing authorities, editing a scheme, and browsing a bunch of combined schemes) able to be used **without** any of the backend and database this doc describes. The current file-based approach is very simple to deploy. There may be non-ATE users who don't want any sort of backend. So, we need to find a way to make the frontend changes agnostic to a backend (the current one being "local storage").
```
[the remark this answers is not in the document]
dabreegster: Exactly, I think simple config should be fine. Right now there's a handful of places doing things with `window.localStorage`; we'd refactor some kind of interface to do those things, implement it twice, and use a bit of config to instantiate the right one. Complexity will grow as we make components for login, sharing, undo/redo, etc, but I don't think any of it will be too hard. Just flagging that as soon as we start working on the backend, we should do this from the start, and not be caught in a difficult refactor later

docs/backend.md
dabreegster: Even if it's something simple like the same person switching laptops and expecting their intermediate work to be visible, auto-saving everything could be useful and/or expected by users

docs/backend.md
```diff
@@ -0,0 +1,138 @@
...
+
+This is a design doc proposing some possible approaches for an ATIP backend. It doesn't represent any final decisions or describe anything actually built.
+
+As of June 2023, ATIP is a web app for editing GeoJSON data. Those GeoJSON files are stored in each user's browser local storage and manually saved/loaded as local files. There is no central place to submit and store files, or any concept around tracking changes to a file.
+
+Terminology:
+
+- a **scheme** is a single ATIP GeoJSON file, with a list of **interventions** or **features** (these are used interchangeably)
+- **ATE** is the central group evaluating and inspecting schemes
+- **LAs** (local authorities) are the groups submitting schemes for some area. (The exact geographic scope doesn't matter; these could be transport or highway authorities or something else)
```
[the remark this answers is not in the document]
robinlovelace-ate: We should be more specific here to avoid confusion. Will suggest changes to show what I mean..

robinlovelace-ate, line 11:
```suggestion
- A **scheme** is one or more infrastructure interventions represented as a single ATIP GeoJSON file.
- An **intervention** is a single point (e.g. crossing), line (e.g. route) or area (e.g. area-wide traffic management intervention) that is part of a scheme.
- **Authorities** are the groups submitting schemes for some area. 
```

robinlovelace-ate: I just used the `suggestion` syntax for the first time 🎉 

In terms of the suggestion: I think we should keep docs in here as non England specific as possible. Short, generalisable, simple and easy to maintain is my thinking. We can add ATE specific content for the version that goes on our website but that's not happening for a while. Local authorities is a term that has multiple meanings (including local authority districts and our transport authorities) so we should avoid, I've suggested just authorities here. And 👎  to using acronyms, let's keep it as readable as possible.

robinlovelace-ate (commented):
Lots of good ideas in here, have added some specific comments and suggestions about terminology. I think a lot of this stuff is 'we can cross that bridge later' but great to have ideas down already.

dabreegster:
I updated to use UK-agnostic terminology everywhere. In lieu of "ATE", I've defined "Central" -- more than happy to use another less awkward term here

dabreegster:
Cleaning up old PRs+branches, so going to merge this for posterity. It links to this PR with all the comments, and states at the top that it's just a brainstorm / set of questions to think about

PR #230, opened 2023-06-09
State changing bugs: disable things when editing 142

Pete-Y-CS:
…files and clearing all

This is a new version of:
https://github.com/acteng/atip/pull/196

The old work for that, for now, is kept at the branch: `state-changing-bugs-142-legacy` so I can refer to it if there's stuff I want to reuse.

### DIFF
--- a/src/lib/common/FileInput.svelte
+++ b/src/lib/common/FileInput.svelte
@@ -1,11 +1,11 @@
 <script lang="ts">
-  import { isAToolInUse } from "../../stores";
-
   export let label: string;
   // This must be unique in the DOM
   export let uniqueId: string;
   // Called with the file contents as text
   export let loadFile: (text: string) => void;
+
+  export let disabled = false;
 
   let fileInput: HTMLInputElement;
 
@@ -26,12 +26,10 @@
     id={uniqueId}
     bind:this={fileInput}
     on:change={onChange}
-    disabled={$isAToolInUse}
+    {disabled}
   />
-  <button
-    type="button"
-    on:click={() => fileInput.click()}
-    disabled={$isAToolInUse}>{label}</button
+  <button type="button" on:click={() => fileInput.click()} {disabled}
+    >{label}</button
   >
 </label>
 
--- a/src/lib/common/FileInput.svelte
+++ b/src/lib/common/FileInput.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+  import { isAToolInUse } from "../../stores";
+
   export let label: string;
   // This must be unique in the DOM
   export let uniqueId: string;
@@ -19,8 +21,18 @@
 
 <!-- TODO Interactive elements inside a label are apparently invalid, but this works -->
 <label>
-  <input type="file" id={uniqueId} bind:this={fileInput} on:change={onChange} />
-  <button type="button" on:click={() => fileInput.click()}>{label}</button>
+  <input
+    type="file"
+    id={uniqueId}
+    bind:this={fileInput}
+    on:change={onChange}
+    disabled={$isAToolInUse}
+  />
+  <button
+    type="button"
+    on:click={() => fileInput.click()}
+    disabled={$isAToolInUse}>{label}</button
+  >
 </label>
 
 <style>
--- a/src/lib/draw/point/point_tool.ts
+++ b/src/lib/draw/point/point_tool.ts
@@ -7,6 +7,7 @@ import {
   overwriteSource,
   type FeatureWithProps,
 } from "../../../maplibre_helpers";
+import { isAToolInUse } from "../../../stores";
 import type { EventHandler } from "../event_handler";
 
 const source = "edit-point-mode";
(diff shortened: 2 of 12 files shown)


Pete-Y-CS:
Tests passing, I'm removing this from WIP!

tests/settings.spec.ts
```diff
@@ -1,12 +1,28 @@
 import { expect, test } from "@playwright/test";
+import { checkPageLoaded, loadInitialPageFromBrowser } from "./shared.ts";
 
-test("changed settings are preserved across refresh", async ({ page }) => {
+test("changed settings are preserved across refresh", async ({ browser }) => {
```
Pete-Y-CS, line 4: This test wasn't passing, according to Alex it was unreliable on `main` too. So I've made some incidental changes to make it more reliable I think. Certainly the changes to shared I think are valuable, maybe it's overkill to be ready for either avoid doubling back or not defaulting to true

src/lib/common/FileInput.svelte
```diff
@@ -19,8 +21,18 @@
...
 <!-- TODO Interactive elements inside a label are apparently invalid, but this works -->
 <label>
-  <input type="file" id={uniqueId} bind:this={fileInput} on:change={onChange} />
-  <button type="button" on:click={() => fileInput.click()}>{label}</button>
+  <input
+    type="file"
+    id={uniqueId}
+    bind:this={fileInput}
+    on:change={onChange}
+    disabled={$isAToolInUse}
```
dabreegster, line 29: Hmm, this component is also used in `BrowseSchemes`, where this store is meaningless. Can we pass in a new `export let disabled = false` prop here, and from `EntireScheme`, set it to this store? I think changes to the store should propagate through properly

src/stores.ts
```diff
@@ -38,6 +38,7 @@ userSettings.subscribe((value) =>
   window.localStorage.setItem("userSettings", JSON.stringify(value))
 );
 export const currentMode: Writable<Mode> = writable("edit-attribute");
+export const isAToolInUse: Writable<boolean> = writable(false);
```
dabreegster, line 41: Could we add a comment explaining what this means?

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -168,9 +181,22 @@
...
+    disabled={$gjScheme.features.length == 0 || $isAToolInUse}>Clear all</button
   >
+  <ConfirmationModal
+    bind:open={displayClearAllConfirmation}
+    title={"Would you like to clear your work?"}
+    message={"This will delete all your drawn interventions."}
+    on:cancelAction={cancelClearAll}
+    on:confirmAction={clearAll}
+  />
+  {#if $isAToolInUse}
```
dabreegster, line 194: Screenshot from 2023-06-12 10-21-03
The new text appearing feels a little distracting to me. What if we put tooltips on the disabled buttons instead, so if the user tries to interact with those functions and can't, they understand? Something like `title={$isAToolInUse ? "edit attribs first" : null}` seems to work. (I dunno if setting to `null` is the cleanest approach. Or how hard plumbing to `FileInput` will be)

src/lib/draw/point/point_tool.ts
```diff
@@ -81,17 +82,22 @@ export class PointTool {
...
-    this.active = true;
+    this.setActivity(true);
     // TODO Figure out where the cursor is and immediately draw? To be useful,
     // the user has to move their mouse anyway; it doesn't matter much
   }
 
   stop() {
     this.cursor = null;
-    this.active = false;
+    this.setActivity(false);
```
dabreegster, line 92: I think this all should be fine, and the tests will tell us if not. If for some reason we called `stop()` on a tool when it wasn't active in the first place, this will mess up the store. But we shouldn't do that right now, so it's OK

tests/settings.spec.ts
dabreegster: (Worth adding a comment in this test about why we have to be prepared for either initial checkbox state?)

src/lib/sidebar/EntireScheme.svelte
Pete-Y-CS: Should be easy to plumb to `FileInput`, should be simply adding a `disabledTooltip` prop. That said idk how many people would hover over a disabled button to work out why it's disabled. Maybe this is a question for a  UX type person?

tests/savefiles.spec.ts
```diff
@@ -1,12 +1,15 @@
...
+  clearExistingInterventions,
+  loadInitialPageFromBrowser,
+} from "./shared.ts";
 
 let page: Page;
 let json;
 
 test.beforeAll(async ({ browser }) => {
-  page = await loadInitialPage(browser);
+  page = await loadInitialPageFromBrowser(browser);
```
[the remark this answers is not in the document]
Pete-Y-CS: It's partly that, it's partly because I've split the function into different parts so we can reuse different steps as appropriate

src/lib/sidebar/EntireScheme.svelte
dabreegster: Indeed. I don't have a strong UX opinion, so let's keep it like this until we have more guidance or hear feedback. Thanks!

dabreegster (approved):
Nicely done! Feel free to merge

dabreegster:
Bump -- any reason to not merge this?

Pete-Y-CS:
Not to my knowledge

robinlovelace-ate:
Final code janitor type job if you get a chance @Pete-Y-CS: Delete the merged branch :shower:

Pete-Y-CS:
I suppose we can set repo-level defaults for this kinda stuff? (Preferring rebase and deleting branch)

robinlovelace-ate:
Yeah good point and links to @Sparrow0hawk's ideas about repo defaults. I was looking at bots like 'bors' and all sorts but seems it's just a per repo setting: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches

robinlovelace-ate:
Can we configure all acteng repos to do that? Will save many clicks if so!

Sparrow0hawk:
I've enabled it on this repo and will look at configuring across all acteng repos

PR #232, opened 2023-06-12
Update v2 schema to match what actually happens

Pete-Y-CS:
Tiny wee pull request so that Typescript is happy when we run `npm run check`

### DIFF
--- a/src/schemas/v2.ts
+++ b/src/schemas/v2.ts
@@ -1,6 +1,9 @@
 // This file is auto-generated; do not manually edit
 
-export type Intervention = Route | Crossing;
+export type Intervention = {
+  Route?: Route;
+  Crossing?: Crossing;
+};
 
 export interface Crossing {
   name?: string;


src/schemas/v2.ts
```diff
@@ -1,6 +1,9 @@
 // This file is auto-generated; do not manually edit
```
dabreegster, line 1: We need to update the tool that produces this file. https://github.com/acteng/atip/tree/ts_forms is an attempted start, but it raises some questions I haven't been able to answer yet:

- How do we want to model enums like this? The style below means it's possible for both a Route and Crossing to be defined, or for neither -- both of those feel wrong to me. The branch I linked just does "discriminated unions" to match the actual JSON currently produced: `type Intervention = { Route: Route } | { Crossing: Crossing };`. https://mariusschulz.com/blog/tagged-union-types-in-typescript is one article I found talking about these options.
- The bit where I got stuck: should all fields everywhere be optional, including enums? When we're working with finalized data, it would kind of be nice to assume everything is properly filled out. But in reality, the UI doesn't enforce that right now -- you can create an object, not fill out any of its form, and save. Or you can start clicking through the form, but not fill everything out. Maybe the reality is that every field _is_ optional and maybe missing. If so, we can adjust `generate_ts.ts` accordingly

Pete-Y-CS: Maybe I'm too comment-averse because I didn't choose to ignore this comment, I had no idea it was there 😅

Pete-Y-CS:
I have found the underlying issue with our JsonSchema `v2.json`: we are using `OneOf` differently in JsonSchema compared to what happens in ATIP.

In ATIP when using v2 schema we have an intervention being an object with exactly one of the fields Route or Crossing filled in e.g.
```JSON
{
  Route: {
    // Route stuff blahblah blah
  },
} 
```
But `OneOf` in JSON schema means we match against one of two types, so whilst an intervention in practice is an object with a `Route` or `Crossing` object as a member, what we've defined in JSONSchema resolves to either Route or Crossing directly
```JSON
{
  // Route stuff Blahbalbhab
}
```

As a side note it looks unusual to me to have capitalised members of a json object so I think we should, if want to keep the current actual format rename so it's:
```JSON
{
  route: {
    // route stuff blahblahblah
  }
}
```

Pete-Y-CS:
Thinking about it I feel like our JsonSchema is correct and we simply want an object which meets either the route _or_ the crossing criteria

dabreegster:
Echoing my question here: https://github.com/acteng/atip/pull/232#discussion_r1226983345. Do we want to allow someone to save an object without completely filling out the form? If no, then we need to pretty dramatically change the UI. If yes, then we can adjust the generated TS types to let every field effectively be optional, and let every oneOf case have nothing at all filled out. This decision will impact downstream data analysis of submitted data.

My instinct is to let everything be optional, and to think carefully about letting people fill in detail gradually. Sometimes we just need to know something will be a route intended for walking, wheeling, and cycling, and don't care about anything else yet. At a later stage of a scheme, we want all detail filled out.

> looks unusual to me to have capitalised members of a json object

Oh definitely, we should update the TS code generation and the dynamic form to produce more typical looking JSON. Backing up a little, this hand-rolled form generation started in #145 as an alternative to broken [JSON schema](https://json-schema.org/) form implementations. We needed a quick prototyping to iterate on the schema, and communicate it as a web form.

I've since found projects like https://zod.dev/. This is maybe something we should consider using?

dabreegster:
> json schema understands one of

Just so it's clear, we are not using any standard JSON schema library. https://github.com/acteng/atip/blob/main/src/lib/forms/generate_ts.ts reads a file like https://github.com/acteng/atip/blob/main/src/schemas/v2.json and produces `v2.ts`. https://github.com/acteng/atip/blob/main/src/lib/forms/AutogenerateForm.svelte reads the same `v2.json` and displays a dynamic form.

That means we are in complete control over what `oneOf` means. I picked this name to match https://protobuf.dev/programming-guides/proto3/#oneof. The way I wrote it originally was meant as "Route XOR Crossing. Not both. Not empty." But I think we may want to allow every field in the forms to be optional, and so for oneOf cases, that means we can represent it in TS as something like `{ Route: Route } | { Crossing: Crossing } | {}`. But before we go down that route, we have to decide if all fields being optional is OK from a UI perspective for people filling out forms and for inspections/analysis perspective when doing anything with the data.

Pete-Y-CS:
> Just so it's clear, we are not using any standard JSON schema library.

Ahh okay my bad. I'll have a rethink and look at `generate_ts`

dabreegster:
Closing the PR and cleaning up the branch, to simplify GH build stuff. Still TBD how to solve the underlying problem here.

PR #233, opened 2023-06-12
Move schema design doc into docs

dabreegster:
I added https://github.com/acteng/schema/blob/main/article.md in early April, but didn't get much feedback at the time. This is a slightly updated copy, moved here for visibility and to re-open some questions, since we're looking at this actively again. Please comment/discuss!

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -1,0 +1,183 @@
+# Schema design doc
+
+This document has some evaluation criteria and open questions regarding the
+data schema that ATE is designing, as of spring 2023.  ATIP has already been
+used to collect data about planned active travel schemes for ATF4, but aside
+from mapping the geometry of interventions, we imposed only the simplest
+categorization -- is it a route, area, crossing, or something else? Going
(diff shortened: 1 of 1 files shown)


robinlovelace-ate (commented):
Looks good but not 100% sure what the intention is for this and who's supposed to read/comment on it. Will be in a better place to review it when I know!

dabreegster:
Intended audience is everyone listed as a reviewer in the PR. The intention is to solicit comments on the _process_ of figuring out the schema, rather than specifics of the schema, like we've been doing before. How much should we understand OSM's or OS's schema for active travel infra? How do we make a schema that can be evolved over time? How do we "finalize" it before sending it off to everyone, and make sure we're going to collect the right data to answer all the questions we'll have?

dabreegster:
I'm cleaning up old branches+PRs, so I'm going to merge it in for posterity. The top of the doc states clearly it's just a brainstorm / notes, not representing any kind of decision

PR #234, opened 2023-06-13
Update Vite to 4.04

Pete-Y-CS:


dabreegster:
It looks like there's a much newer vite after the 4.0 series (https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md). Should we upgrade to the latest? Only thing to watch out for is compatibility with https://github.com/nshen/vite-plugin-wasm-pack; we're pinned to a specific older version there, because the latest is broken

Sparrow0hawk:
Should probably give it a test with a newer version if possible.

This came about because when running `npm ci` I got a vulnerability warning relating to this - https://security.snyk.io/package/npm/vite/4.0.0

Pete-Y-CS:
I just gave it a test on my newly working WSL install and happily it seems to work with latest stable (4.1.1). Could be ready to merge?

dabreegster (approved):
No noticeable behavioral differences, LGTM!

dabreegster:
Feel free to merge whenever

PR #237, opened 2023-06-14
Switch to homepage without opening new tab

Pete-Y-CS:


src/pages/App.svelte
```diff
@@ -95,9 +95,7 @@
 <Layout>
   <div slot="sidebar">
     <div>
-      <button type="button" on:click={() => window.open("index.html")}>
-        Home</button
-      >
+      <button type="button" on:click={() => window.location.replace("/index.html")}> Home</button>
```
dabreegster, line 98:
```suggestion
      <button type="button" on:click={() => window.location.href = "index.html"}> Home</button>
```

The absolute path breaks on Github pages; try it from https://acteng.github.io/atip/home-without-new-tab/scheme.html?authority=Adur

dabreegster (approved):
LGTM after fixing the path. I don't remember a strong reason we made it open in a new tab originally. I wouldn't expect many users to click this and need to change areas right now, but maybe as inspectors use the new browse scheme page, this will change a bit. We could also consider linking to that.

PR #240, opened 2023-06-16
Change to use `.js` in imports according to ts convention

Pete-Y-CS:
Begone red squiggly lines in my editor (tests still run)


dabreegster (approved):
Related to #227 a bit, maybe. I don't understand why tsconfig works differently for the main app and tests. `npm run check` works fine before and after this change, so fine to make VS code happy too

dabreegster:
Feel free to merge whenever

PR #241, opened 2023-06-16
Fix final TS errors and enable npm run check

dabreegster:
This is an alternative to #232 to fix some errors involving incomplete forms. All TS errors are now gone, and so `npm run check` is added to Github CI, with the goal being that we never check in anything that doesn't type-check again!

(As a further step, enabling strict mode would be great, but there are more problems to sort out first)

### DIFF
--- a/src/lib/sidebar/InterventionList.svelte
+++ b/src/lib/sidebar/InterventionList.svelte
@@ -14,13 +14,18 @@
       return feature.properties.planning?.name || "Untitled polygon";
     }
     if (schema == "v2") {
-      return (
-        feature.properties.v2?.Route?.name ||
-        feature.properties.v2?.Crossing?.name ||
-        "Untitled intervention"
-      );
+      if (feature.properties.v2) {
+        if ("Route" in feature.properties.v2) {
+          return feature.properties.v2!.Route.name || "Untitled route";
+        }
+        if ("Crossing" in feature.properties.v2) {
+          return feature.properties.v2!.Crossing.name || "Untitled crossing";
+        }
+      }
+      return "Untitled intervention";
     }
 
+    // v1
     if (feature.properties.name) {
       return feature.properties.name;
     }
--- a/src/pages/App.svelte
+++ b/src/pages/App.svelte
@@ -95,7 +95,12 @@
 <Layout>
   <div slot="sidebar">
     <div>
-      <button type="button" on:click={() => window.location.href = "index.html"}> Home</button>
+      <button
+        type="button"
+        on:click={() => (window.location.href = "index.html")}
+      >
+        Home</button
+      >
       <button type="button" on:click={toggleAbout}>About</button>
       <button type="button" on:click={toggleInstructions}>Instructions</button>
     </div>
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -31,7 +31,10 @@ jobs:
       run: npx playwright install --with-deps
 
     - name: Run Playwright tests
-      run: npm test
+      run: npm run test
+
+    - name: Run type-checker
+      run: npm run check
 
     - uses: actions/upload-artifact@v3
       with:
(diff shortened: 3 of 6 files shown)


src/pages/App.svelte
```diff
@@ -95,7 +95,12 @@
 <Layout>
   <div slot="sidebar">
     <div>
-      <button type="button" on:click={() => window.location.href = "index.html"}> Home</button>
+      <button
```
dabreegster, line 98: Just `npm run fmt` diffs missed from another PR

dabreegster:
Gentle ping on reviews, please

src/lib/sidebar/InterventionList.svelte
```diff
@@ -14,13 +14,18 @@
...
     }
     if (schema == "v2") {
-      return (
-        feature.properties.v2?.Route?.name ||
-        feature.properties.v2?.Crossing?.name ||
-        "Untitled intervention"
-      );
+      if (feature.properties.v2) {
+        if ("Route" in feature.properties.v2) {
+          return feature.properties.v2!.Route.name || "Untitled route";
```
Pete-Y-CS, line 19: I'm still a bit confused. Maybe I'm missing something but I think we're using something that looks exactly like jsonschema (in `v2.json`) and choosing to have it behave differently and I don't know why we're doing that. Why don't we just have `OneOf` work such that `v2` is either a `Route` or a `Crossing` rather than it containing either `Route` or `Crossing` as a member?

Pete-Y-CS (commented):
I think we should have our `generate_ts` and form generation work in line with JsonSchema, I think it would be less confusing to newcomers and I don't really understand the downside

dabreegster:
My goal with this PR is to iterate on what we have already, get to 0 TS errors, and start enforcing checks, to let us more confidently develop going forwards. If we want to change fundamentally how the schema generation works, I would vote for doing that in a followup.

#145 has context about why I didn't choose https://json-schema.org/ initially. We did prototype with it in https://github.com/acteng/schema/, but there were problems:

- I couldn't get any of the libraries to display forms in Svelte to work. Things may be better now, or we may need more control over the appearance of the form anyway, to display things like help text about certain fields. See #99.
- The language to write schemas is more powerful and expressive -- see https://github.com/acteng/schema/blob/main/option_atf4/all.json. If we wind up writing our own Svelte form adapter for it, we either need to support everything it handles (even if we don't use it) or pick a subset of the language to handle. Either way is an investment into something where the existing tooling isn't fully doing what we need now. I'm not convinced JSON schema is the right thing for us to use yet long-term, and we haven't evaluated alternatives like https://zod.dev/ yet. Hence starting with something custom, but very simple and fully in our control.

src/lib/sidebar/InterventionList.svelte
dabreegster: Any way we express the `OneOf` needs to work with https://www.typescriptlang.org/docs/handbook/2/narrowing.html -- in the code, we need to be able to distinguish whether something is a route, a crossing, or undefined/missing/not filled out yet.

The current approach is `type Intervention = { Route: Route } | { Crossing: Crossing } | {}`. This explicitly nests the `OneOf` case as another dictionary, so the JSON looks like `v2: { Route: { name: "foo", ... } }`. Type narrowing uses the `in` operator.

Alternatively, we could remove the nesting and use a `kind` or `type` field. This'd be something like `type Intervention = Route | Crossing | {}`, where the `Route` interface would include a required field `kind: "Route"`, and the JSON would be something like `v2: { kind: "Route", name: "foo", ... }`. Our [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) in this case would check `properties.v2.kind == "Route"` to do the narrowing. I think this might be the approach you're in favor of?

For reference, https://transform.tools/json-schema-to-typescript is a tool to see one interpretation of expressing the JSON schema in TS. (I don't think there's necessarily one canonical implementation or anything like that)

dabreegster:
Going to switch this to draft. I have a half-finished next step, but may not look at it again soon. It'd be great to start enforcing TS check, but this last error really is something that needs more thought

dabreegster:
Closing for now, since the immediate need of TS errors is solved elsewhere. Will revisit how to represent the union/enum types later

Issue #228, opened 2023-06-08
Cleanups after event handler refactor


dabreegster:
#220 made some big changes! Here's some smaller followups we can do in any order:

- [ ] Make sure callbacks get correctly destroyed when the Svelte component reloads. Now that's centralized in Toolbox, but it's probably not correct as it is now. #174 is related. It'd be awesome if we could figure out how to automatically test this, instead of manually inserting logs, `npm run dev`, editing components, watching logs, etc.
- [x] Deal with `KeyHandler.svelte` and generally audit all places handling the escape key; I think currently they all run and stomp on each other. We can start by adding a Playwright test of this, pressing escape from modes where we shouldn't just wind up back in attribute mode.
- [ ] Add some comments / dev docs about how event handling works, why we wound up doing it this way
- [ ] Continue to think if there's a cleaner way of forwarding events in `GeometryMode`. Maybe we can store the currently active tool in local state, and automatically pass _all_ events to that. And make each tool effectively embed its own `EventHandler` and delegate to it, or something like that.

PR #244, opened 2023-06-23
Fix escape key

dabreegster:
Part of #228. #220 introduced some bugs with handling keys. This PR adds tests and fixes the problems.


src/lib/draw/event_handler.ts
```diff
@@ -10,7 +10,7 @@ export const MapEvents = [
   "dragstart",
 ];
 
-export const DocumentEvents = ["keyPress", "keyUp", "keyDown"];
+export const DocumentEvents = ["keypress", "keyup", "keydown"];
```
dabreegster, line 13: The bug is that the DOM events aren't camel-cased. And sadly there's no warning or anything if you do `document.addEventListener("typo")`

tests/modes.spec.ts
```diff
@@ -109,10 +107,10 @@ test("editing geometry of a polygon works", async () => {
   await page.getByRole("button", { name: "Edit geometry" }).click();
   await clickMap(page, 312, 501);
   // This button indicates the mode is working
-  await page.getByRole("button", { name: "Finish" }).click();
+  await expect(page.getByRole("button", { name: "Finish" })).toBeVisible();
```
dabreegster, line 110: We just want to see the button. Actually clicking it changes state, and is confusing

tests/modes.spec.ts
```diff
@@ -141,3 +139,105 @@ test("testing add a route and save it", async () => {
...
+  await page.getByRole("button", { name: "New polygon (snapped)" }).click();
+  await clickMap(page, 500, 500);
+  await clickMap(page, 400, 500);
+  await clickMap(page, 400, 600);
+  await page.getByRole("button", { name: "Finish" }).click();
+
+  await page.getByRole("button", { name: "Edit geometry" }).click();
+  await clickMap(page, 430, 515);
+  // This button indicates the mode is working
+  await expect(page.getByRole("button", { name: "Cancel" })).toBeVisible();
```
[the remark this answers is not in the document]
dabreegster: I've fixed the flaky test by reseting the map viewport between tests. It was very subtle -- the test worked fine on its own or with some other tests in the same file, but one particular test here changes the map view a bit, so the order of runs matters!

Pete-Y-CS (approved):

Issue #231, opened 2023-06-12
Tie down schema


robinlovelace-ate:
A big part of v3 is switching to a schema that asks ATIP users to fill in specific bits of information such as path width depending on the type of scheme that is being added. According to the ATF guidance there are 11 (I think?) types of scheme: https://www.gov.uk/government/publications/how-to-complete-the-active-travel-fund-4-proforma/guidance-note-for-local-authorities-to-support-completion-of-the-active-travel-fund-4-proforma#monitoring-and-evaluation

Things that we need to tie down:

- [x] #242
- [ ] Feedback on the user interface and user experience from the inspectorate
- [ ] Identification of data needs not captured by the schema currently
- [ ] Identification of functionality in the schema that is not needed

We don't need to have a dropdown to select from one of those but we **do** need to be able to categorise schemes/interventions into one or more of those categories (are there more) from ATIP data. This issue is about the contents of the schema, currently

https://github.com/acteng/atip/blob/2d5893c4e341df4754c2e167966c47b122425c80/src/schemas/v2.json#L1-L153

@KitAllwinter was looking at this. We should iterate based on his feedback and then widen it out with an updated schema.

There's a wider issue about UI, but that's for another issue. Quickfire question @dabreegster is it easy to turn on the v2 schema briefly for testing? Currently only seeing V1, Critical Issues, Housing Developments issues.

dabreegster:
> we do need to be able to categorise schemes/interventions into one or more of those categories

To check my understanding: we still likely want to capture more detailed information like the current v2 draft does, but whatever we collect, we need to map into things like high/medium/low complexity for one of those 11 categories?

> is it easy to turn on the v2 schema briefly for testing

Just append `&schema=v2` to the end of a URL to get there immediately. Enable https://github.com/acteng/atip/blob/2d5893c4e341df4754c2e167966c47b122425c80/src/pages/App.svelte#L109 to get it back in the list

robinlovelace-ate:
Heads-up @acteng/atip-core I've just updated the original post with more clarity on what I think the ask is here. I do think this is a priority and, if we set up a meeting/workshop and focus on clear questions, I don't think it is a huge ask.

dabreegster:
Closing old issues. We have a very simple v2 schema today, and there's a bigger discussion internally around the kinds of detail we need for infrastructure.

PR #245, opened 2023-06-25
Color adjustments from Kelin

dabreegster:
Kelin (UA programme manager with an architecture background) kindly joined a UX session, and we generated lots of ideas to improve ATIP. Here are a few simple ones.

First, change the colors for routes and areas, and add an outline to areas. The outline emphasizes it as a clickable object more, and matches how Felt looks. Before:
Screenshot from 2023-06-25 15-02-11
After:
Screenshot from 2023-06-25 15-02-17

And second, change the color scale used for showing speed limits. It's based on the first big map in https://knightlab.northwestern.edu/2016/07/18/three-tools-to-help-you-make-colorblind-friendly-graphics/, but we adjusted the middle color, since it's too close to the basemap background color. We adjusted it by using the "mix" tool in `gpick`, mixing it with the category for 40 mph. Before:
Screenshot from 2023-06-25 15-02-36
After:
Screenshot from 2023-06-25 15-04-22

Also note that the new color for routes (grey) is sort of a "continuation" of that color scale, a darker version of the "< 20" speed category. This is meant to suggest new route infrastructure proposed will be even better than an naturally slow-traffic street.

We can take any or all of the ideas here. Any opinions?

### DIFF
--- a/src/colors.ts
+++ b/src/colors.ts
@@ -1,7 +1,7 @@
 export const colors = {
   // https://colorbrewer2.org/#type=qualitative&scheme=Set1&n=4
-  area: "#e41a1c",
-  route: "#377eb8",
+  area: "#D94324",
+  route: "#555F5E",
   crossing: "#4daf4a",
   other: "#984ea3",
 
--- a/src/lib/draw/InterventionLayer.svelte
+++ b/src/lib/draw/InterventionLayer.svelte
@@ -126,7 +126,15 @@
     filter: ["all", isPolygon, hideWhileEditing] as FilterSpecification,
     color:
       schema == "planning" ? colorByReferenceType : colorByInterventionType,
+    opacity: 0.2,
+  });
+  overwriteLineLayer($map, {
+    id: "interventions-polygon-outlines",
+    source,
+    filter: ["all", isPolygon, hideWhileEditing] as FilterSpecification,
+    color:
+      schema == "planning" ? colorByReferenceType : colorByInterventionType,
     opacity: 0.5,
-    // TODO Outline too?
+    width: 0.7 * lineWidth,
   });
 </script>
--- a/src/lib/layers/SpeedLimits.svelte
+++ b/src/lib/layers/SpeedLimits.svelte
@@ -22,7 +22,7 @@
   // Show along a route if specified, or show all otherwise
   export let id: number | undefined;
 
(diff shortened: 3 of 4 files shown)


Pete-Y-CS:
I think the first changes look great, and the reasoning sounds sound for all changes.

To me the second changes... well it's quite noisy in either case. I think it's hard to follow the map with speed limits but I think that might be a tricky problem to solve. I think the changes make it look better but still difficult, so I kinda like this as what seems like a step in the right direction

robinlovelace-ate:
These are clear improvements I think, *except* for the speed limits. I think dark red for the fastest road category is good. Overall thumbs up and will add my review accordingly.

robinlovelace-ate (approved):
Big 👍 up to this with only minor question being if the fastest roads should be a darker red.

Slightly off-topic question based on the idea of modularity and config files vs business logic: can some of these configs be usefully split out as, for example, `speed-limit-cols.json`?

As it happens I recently did a similar thing on another project and it helped make the code more legible and accessible I think.

dabreegster:
Thank you for the reviews / feedback! I reverted the changes to the speed limit layer for now, since that needs more thought. Just changing base intervention colors and adding area outlines.

> can some of these configs be usefully split out as, for example, speed-limit-cols.json?

We partly do: https://github.com/acteng/atip/blob/main/src/colors.ts. For some projects, I've tried [centralizing most color choices](https://github.com/a-b-street/abstreet/blob/main/map_gui/src/colors.rs). A downside is loss of context and requiring sufficiently explanatory names for everything. At the moment, I think the codebase is small enough to work as it is (and easy to live-edit with Vite's hot reloading). If/when we properly handle multiple basemaps, then we could revisit how we organize color code. That's a much harder problem -- all color choices depend on the maptiler style, and don't look good against satellite.

robinlovelace-ate:
Looking good for Manchester where I plan to demo ATIP v2-latest (OK with that as a version name for now Dustin, with thinking still being that v3 will have the updated schemas after #231 ?) to transport planners.

image

@EmmaVinterATE shared a link to guidance on colourschemes for data visualisation, could you share the link to that if you get a chance Emma?

EmmaVinterATE:
No problem, here is the link: https://analysisfunction.civilservice.gov.uk/policy-store/data-visualisation-colours-in-charts/ . Hope this helps!

PR #246, opened 2023-06-25
Show clickable objects on the map using cursor, not an outline. Reser…

dabreegster:
…ve the outline for the object currently opened in the sidebar.

The UX session with Kelin demonstrated confusion about possible mouse actions at different points. It's not obvious you can click a waypoint in the polygon or route tool to delete it, but also click and drag to move it. Other tools change the cursor icon to show this. This PR is a very gentle first step in that direction, showing hover state on the map by changing the cursor. The thick black outline is reserved for showing the object opened in the sidebar.

Before/after screenshots are tough; try before at https://acteng.github.io/atip/scheme.html?authority=Adur, after at https://acteng.github.io/atip/cursors/scheme.html?authority=Adur

I'm thinking through the changes to the cursor we want to make for all of the drawing tools, aiming to draw inspiration from terra-draw, mapbox-gl-draw, etc.

### DIFF
--- a/src/lib/draw/HoverLayer.svelte
+++ b/src/lib/draw/HoverLayer.svelte
@@ -18,6 +18,11 @@
     sidebarHover,
   } from "../../stores";
 
+  // Show clickable objects on the map using the cursor
+  $: {
+    $map.getCanvas().style.cursor = $mapHover ? "pointer" : "grab";
+  }
+
   let source = "hover";
 
   // Use a layer that only ever has zero or one features for hovering. I think


src/lib/draw/HoverLayer.svelte
```diff
@@ -18,6 +18,13 @@
     sidebarHover,
   } from "../../stores";
 
+  // Show clickable objects on the map using the cursor
+  $: {
+    $map.getCanvas().style.cursor = $mapHover ? "pointer" : "grab";
```
dabreegster, line 23: Once other drawing tools also modify the cursor, we'll do this differently -- maybe only running this when we're in attribute or geometry mode.

dabreegster:
We could keep the outline too. It may be a bit confusing, because we're conflating two things: is your cursor over an object, and is an object currently focused/selected (by clicking on it in the map or in the sidebar)? While an object is focused, it's possible to click something else, but we don't currently draw the outline for the second object.

felt.com and geojson.io both use only a cursor to show an object is selectable, and add an outline or tooltip only to the current focused object. Google MyMaps is closer to what we do now -- both a hover/outline effect and cursor for the selected and hovered object.

Pete-Y-CS:
one thing I'm thinking about is schemes can often have things overlaid, right? You could have a route through a low traffic neighbourhood. You could have two routes going along the same stretch of road (I guess you'd split that up but maybe there are times). In any case I figure it might be nice to show for sure what you're gonna select if you're for example hovering over a route through an area

dabreegster:
Oh, overlapping objects is a really good point! Totally right, that's the reason we want to visually change the object to show what's being selected.

I'll update this PR to only change cursors after standup

dabreegster:
I updated to keep existing outlines, but change the cursor.

Later on, I'll look into using cursors in all the draw tools.

Pete-Y-CS (approved):

PR #249, opened 2023-06-25
Fix bug where we zoom to an object at the wrong time

dabreegster:
See the test for a description of the bug. We zoom to an object when opening it from the sidebar. But while the form is open, if you move or zoom the map and then click anything in the open form, we zoom again. This PR adds a test and fixes the problem.

### DIFF
--- a/tests/modes.spec.ts
+++ b/tests/modes.spec.ts
@@ -258,8 +258,7 @@ test("the viewport changes only once when opening a form", async () => {
   await page.getByText("Crossing", { exact: true }).click();
   // If we're incorrectly changing the viewport here, wait for the effect to happen
   await page.waitForTimeout(700);
-  // TODO Expect the test to fails here. Remove 'not' after fixing.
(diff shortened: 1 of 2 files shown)


tests/modes.spec.ts
```diff
@@ -141,3 +141,36 @@ test("testing add a route and save it", async () => {
...
+  await page.getByRole("button", { name: "New point" }).click();
+  await clickMap(page, 500, 500);
+  expect(new URL(page.url()).hash).toEqual(defaultViewport);
+
+  // Close the form
+  await page.getByRole("button", { name: "Save" }).click();
+
+  // Open the form from the sidebar. The viewport should change
+  await page.getByRole("button", { name: "1) Untitled point" }).click();
+  // The zoom effect is set to happen over 500ms, so wait a bit
```
dabreegster, line 158: I'd love to make Playwright's auto-await magic work for all the maplibre interactions, but not sure how. Manual waits are the workaround.

Pete-Y-CS (approved):

PR #251, opened 2023-06-27
Update most NPM dependencies

### LINKED ISSUE #188: Upgrade to maplibre-gl 3.0
3.0 was just released (literally so recently that they're still fiddling with their release scripts to get the NPM package uploaded)! We should understand [the changelog](https://github.com/maplibre/maplibre-gl-js/blob/main/CHANGELOG.md) carefully and update it. I don't know if any of the changes will have any effect, or if this will be a very routine update.

dabreegster:
Closes #188. I upgraded to MapLibre 3 a few weeks ago in https://github.com/dabreegster/osm2svelte and haven't seen any issues. There are new API calls like https://maplibre.org/maplibre-gl-js-docs/api/sources/#geojsonsource#updatedata that we might take advantage of in the future.

Also upgrades to the new [Svelte 4](https://svelte.dev/blog/svelte-4). No big new changes, but possibly a few places where we can get [better TS support](https://svelte.dev/docs/v4-migration-guide).

No behavioral changes I can notice from manually playing around or running tests.

### DIFF
--- a/src/lib/common/Modal.svelte
+++ b/src/lib/common/Modal.svelte
@@ -6,6 +6,7 @@
 </script>
 
 <!-- svelte-ignore a11y-click-events-have-key-events -->
+<!-- svelte-ignore a11y-no-static-element-interactions -->
 <div
   class="background"
   on:click={() => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,22 +18,22 @@
         "@turf/nearest-point-on-line": "^6.5.0",
         "@types/geojson": "^7946.0.10",
         "comlink": "^4.4.1",
-        "maplibre-gl": "^2.4.0",
+        "maplibre-gl": "^3.1.0",
         "route-snapper": "^0.1.14",
-        "svelte": "^3.54.0"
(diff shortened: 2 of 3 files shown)


src/lib/common/Modal.svelte
```diff
@@ -6,6 +6,7 @@
 </script>
 
 <!-- svelte-ignore a11y-click-events-have-key-events -->
+<!-- svelte-ignore a11y-no-static-element-interactions -->
```
dabreegster, line 9: We use the big background div as a clickable thing to close the modal. Using a div for clicking goes against a bunch of a11y guidelines, and there's a new warning here to silence.

But actually, we could probably write this component in a more modern way using [<dialog>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog). (Tip thanks to https://www.youtube.com/watch?v=ywtkJkxJsdg)

Pete-Y-CS:
Tests pass and we're mostly just changing some version numbers, lgtm!

Pete-Y-CS (approved):

Issue #247, opened 2023-06-25
Rethink autosaving while editing geometry


dabreegster:
With how ATIP works today, it's easy for the user to suddenly lose their work:

- if you draw a new polygon, but then click a different tool without saving, you completely lose the new polygon
- if you edit an existing route or polygon, but then click something in the sidebar or another tool, you lose the edits

We asked this question in https://github.com/acteng/atip/issues/142#issuecomment-1552820804, but never really answered it. I see a few options:

1) Silently throw away work when changing states. We mostly do this today.
2) Prompt the user to save or discard changes when they try to switch to a different mode.
3) Disable the ability to click buttons or things in the sidebar when we're drawing or editing geometry. #230 started going in this direction for the very simple case of saving/loading files.
4) Auto-save all geometry changes as the user makes them. When switching to another tool, changes are never lost.

After the UX session with Kelin, I'm more convinced that option 4 is the behavior that best matches expectations and how other tools usually work. This would mean when creating a new polygon or route, as soon as the geometry is valid (at least 2 or 3 points), we add a new object to the sidebar. Option 4 would make most sense when paired with a universal undo/redo function in ATIP, something we're headed towards anyway. That way if you click an object and accidentally adjust geometry, you can just Ctrl+Z.

Pete-Y-CS:
I think autosave is a great idea. I have a slight concern that undo might be a prerequisite, or that we'll have to think very carefully about how we use autosave. 

It feels like it could be annoying to try changing a snapped route, finding it didn't do what you'd like, and not having a way to go back because it autosaved.

As for the specific options suggested, I quite like option 2, but not against 4 so long as we are sure we're gonna resolve undo

Pete-Y-CS:
Caveat to above prefrences: 2 and 4 both seem fine to me but speaking to a UX person/users would be ideal

dabreegster:
> It feels like it could be annoying to try changing a snapped route, finding it didn't do what you'd like, and not having a way to go back because it autosaved.

A "blunt" form of undo would be very easy. When you first start editing, we remember the original geometry, and if the user presses the existing "Cancel" button, we revert to that. The more granular undo would let you control each new waypoint, drag, etc while inside the tool, and outside a specific tool, reverse entirely new features or entire edits to features.

Pete-Y-CS:
Maybe we have the blunt undo for now and granular undo later on, sounds like a great solution

Issue #44, opened 2022-11-27
Undo/redo


Robinlovelace:
Split out from #5

dabreegster:
https://news.ycombinator.com/item?id=34312546
This is a comically simple idea -- encode the entire state of the app into the URL every time there's a change.

PR #253, opened 2023-06-27
Don't lose geometry edits when changing modes

dabreegster:
See #247 for context. This PR does a few things:

- If you're drawing something new and switch to another mode (by clicking something in the sidebar, or clicking another draw tool), the new object is added anyway, but the form isn't opened. The user doesn't unexpectedly lose work.
- If you're in edit geometry mode and switch to another mode, the unsaved changes do get saved.
- The existing "Cancel" buttons continue to work as expected. For new objects, they totally throw away whatever's drawn. For editing, they revert to the previous state of the object. This lets the user undo any accidental changes.

These changes are for polygons and linestrings, but not points. The point tool is trivial; the user just has to click the map to place or move something. If they go click another button somewhere, there's no reasonable "last edited state" to save.

Manual testing seems rock solid, but I'm working on new Playwright tests now.

### DIFF
--- a/src/lib/draw/GeometryMode.svelte
+++ b/src/lib/draw/GeometryMode.svelte
@@ -124,6 +124,21 @@
     });
   }
 
+  // Auto-save
+  for (let tool of [polygonTool]) {
+    tool.addEventListenerUpdated((feature) => {
+      if ($currentMode == thisMode) {
+        gjScheme.update((gj) => {
+          let updateFeature = gj.features.find(
+            (f) => f.id == currentlyEditing
+          )!;
+          updateFeature.geometry = feature.geometry;
+          return gj;
+        });
+      }
+    });
+  }
+
   // Handle failures
   for (let tool of [pointTool, polygonTool, routeTool]) {
     tool.addEventListenerFailure(() => {
--- a/src/lib/draw/GeometryMode.svelte
+++ b/src/lib/draw/GeometryMode.svelte
@@ -94,6 +94,18 @@
           let updateFeature = gj.features.find(
             (f) => f.id == currentlyEditing
           )!;
+          if (!updateFeature) {
+            stopEditing();
+            // We could've been editing anything; just handle all possibilities
+            pointTool.stop();
+            polygonTool.stop();
+            routeTool.stop();
+            window.alert(
+              "You loaded another file or cleared everything while editing. Your changes were lost."
+            );
+            return gj;
+          }
+
           updateFeature.geometry = feature.geometry;
           if (feature.properties.length_meters) {
             updateFeature.properties.length_meters =
--- a/src/lib/draw/GeometryMode.svelte
+++ b/src/lib/draw/GeometryMode.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import type { LineString, Polygon } from "geojson";
+  import type { Geometry, LineString, Polygon } from "geojson";
   import { MapMouseEvent } from "maplibre-gl";
   import { currentMode, gjScheme, map, mapHover } from "../../stores";
   import type { Feature, FeatureUnion } from "../../types";
@@ -27,6 +27,8 @@
     | "snap-polygon"
     | "route"
     | null = null;
+  // Retain a copy of the original geometry before starting to edit
+  let uneditedGeometry: Geometry | null = null;
 
   export function start() {}
   export function stop() {
@@ -43,8 +45,7 @@
       });
     }
 
-    currentlyEditing = null;
-    currentlyEditingControls = null;
+    stopEditing();
     mapHover.set(null);
   }
 
@@ -70,8 +71,7 @@
       });
 
       // Stay in this mode
-      currentlyEditing = null;
-      currentlyEditingControls = null;
+      stopEditing();
     }
   });
 
@@ -94,8 +94,7 @@
       });
 
       // Stay in this mode
-      currentlyEditing = null;
-      currentlyEditingControls = null;
+      stopEditing();
     }
   });
 
@@ -118,8 +117,7 @@
         });
 
         // Stay in this mode
-        currentlyEditing = null;
-        currentlyEditingControls = null;
+        stopEditing();
       }
     });
   }
@@ -143,7 +141,7 @@
   for (let tool of [pointTool, polygonTool, routeTool]) {
     tool.addEventListenerFailure(() => {
       if ($currentMode == thisMode) {
-        // Don't modify the thing we were just editing
+        // Revert to the unedited geometry
         gjScheme.update((gj) => {
           let feature = gj.features.find((f) => f.id == currentlyEditing)!;
           if (!feature) {
@@ -152,13 +150,13 @@
             );
             return gj;
           }
+          feature.geometry = uneditedGeometry;
           delete feature.properties.hide_while_editing;
           return gj;
         });
 
         // Stay in this mode
-        currentlyEditing = null;
-        currentlyEditingControls = null;
+        stopEditing();
       }
     });
   }
@@ -254,6 +252,7 @@
     let feature = maybeFeature!;
 
     currentlyEditing = id;
+    uneditedGeometry = JSON.parse(JSON.stringify(feature.geometry));
 
     if (feature.geometry.type == "LineString") {
       routeTool.editExistingRoute(feature as Feature<LineString>);
@@ -272,6 +271,12 @@
       pointTool.start();
       currentlyEditingControls = "point";
     }
+  }
+
+  function stopEditing() {
+    currentlyEditing = null;
+    currentlyEditingControls = null;
+    uneditedGeometry = null;
   }
 </script>
 
--- a/src/lib/draw/GeometryMode.svelte
+++ b/src/lib/draw/GeometryMode.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import type { Geometry, LineString, Polygon } from "geojson";
+  import type { LineString, Polygon } from "geojson";
   import { MapMouseEvent } from "maplibre-gl";
   import { currentMode, gjScheme, map, mapHover } from "../../stores";
   import type { Feature, FeatureUnion } from "../../types";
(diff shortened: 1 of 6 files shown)


src/lib/draw/GeometryMode.svelte
```diff
@@ -38,88 +41,58 @@
...
 
-  for (let tool of [pointTool, polygonTool]) {
-    tool.addEventListenerSuccess((feature) => {
+    stopEditing();
+    mapHover.set(null);
+  }
+
+  // Handle successful edits
+  for (let tool of [pointTool, polygonTool, routeTool]) {
+    tool.addEventListenerSuccess((editedFeature) => {
```
dabreegster, line 69: It may be easier to review individual commits in order. I also did some refactoring here to merge the `routeTool` event handlers for areas vs routes into one. The added type-safety wasn't useful at all; it just meant we duplicated a bunch of handler code in here.

src/lib/draw/GeometryMode.svelte
```diff
@@ -38,88 +41,58 @@
...
+        stopEditing();
+      }
+    });
+  }
+
+  // Auto-save for everything except pointTool
+  for (let tool of [polygonTool, routeTool]) {
+    tool.addEventListenerUpdated((feature) => {
+      if ($currentMode == thisMode) {
+        // Just remember the update; don't apply it yet
```
dabreegster, line 94: We have a choice here. We could also save updates immediately as they happen, and to handle "Cancel", keep a copy of the feature before starting to edit. I've chosen not to do that, because saving when we switch mode is a bit more performant and is simpler. When we implement undo in #44, we can make the decision here to retain individual edits within a tool, or to compress them all like this. It's just a matter of doing something else in this callback.

src/lib/draw/polygon/polygon_tool.ts
```diff
@@ -82,6 +84,8 @@ export class PolygonTool {
   finish() {
     let polygon = this.polygonFeature();
     if (polygon) {
+      // TODO RouteTool passes a copy to each callback for paranoia. Should we
```
dabreegster, line 87: Unrelated to this PR, just something I noticed while touching this code again. There were bugs long ago with split route mode, and as a response, I made defensive copies to callbacks. Might be good for us to do the same everywhere for consistency, since we haven't got anything like Rust's borrow checker to prevent us from accidentally modifying something passed to a callback.

src/lib/draw/route/route_tool.ts
```diff
@@ -94,6 +98,8 @@ export class RouteTool {
       this.inner.onMouseMove(e.lngLat.lng, e.lngLat.lat, circleRadiusMeters)
     ) {
       this.redraw();
+      // TODO We'll call this too frequently
```
dabreegster, line 101: This one is a bit spammy. Every time you mouseover a waypoint, it'll act like an edit, even if nothing has changed yet. I'll fix this upstream in route-snapper and update here later.

src/lib/draw/route/route_tool.ts
```diff
@@ -302,18 +309,9 @@ export class RouteTool {
...
-      let f = JSON.parse(rawJSON) as Feature;
-      if (f.geometry.type == "LineString") {
-        for (let cb of this.eventListenersSuccessRoute) {
-          cb(JSON.parse(rawJSON) as FeatureWithProps<LineString>);
-        }
-      } else {
-        for (let cb of this.eventListenersSuccessArea) {
-          cb(JSON.parse(rawJSON) as FeatureWithProps<Polygon>);
-        }
+      // Pass copies to each callback
```
dabreegster, line 312: Here's where I got rid of the two different callbacks for route vs area

dabreegster (commented):
And a demo, where I create/edit stuff, but then change to another mode before finishing:

https://github.com/acteng/atip/assets/1664407/c7fd4cd2-0e92-4d49-84d3-1b8859553173

tests/modes.spec.ts
```diff
@@ -54,6 +54,31 @@ test("creating a new freehand polygon opens a form", async () => {
...
+  await page.getByRole("button", { name: "New polygon (freehand)" }).click();
+  await clickMap(page, 500, 500);
+  await clickMap(page, 400, 500);
+  await clickMap(page, 400, 600);
+
+  await page.getByRole("button", { name: "Cancel" }).click();
+  await expect(page.getByText("0 objects")).toBeVisible();
+});
+
+// TODO Repeat switching modes and canceling for other draw tools
```
dabreegster, line 80: We have really repetitive tests... it's tedious to enumerate all of them manually. Will think about future refactors to make this less painful.

dabreegster (commented):
Some tests added. This PR is ready!

dabreegster:
Adding Robin as a reviewer for UX / "is this nicer?" check: https://acteng.github.io/atip/autosave/scheme.html?authority=Adur

robinlovelace-ate (approved):
I had noticed that geometries vanished when changing modes. Based on the need and a view of the video, very impressive work, lots of TypeScript code I won't try to understand, general 👍

robinlovelace-ate:
> Adding Robin as a reviewer for UX / "is this nicer?" check: https://acteng.github.io/atip/autosave/scheme.html?authority=Adur

Will try now...

robinlovelace-ate:
Tested and seems to work great to me. Great work and many thanks for detailed comments on the code, will help onboard others to the codebase and make code review easier (although I'm not in a position to review the code)!

src/lib/draw/GeometryMode.svelte
```diff
@@ -124,6 +124,21 @@
     });
   }
 
+  // Auto-save
+  for (let tool of [polygonTool]) {
+    tool.addEventListenerUpdated((feature) => {
+      if ($currentMode == thisMode) {
+        gjScheme.update((gj) => {
+          let updateFeature = gj.features.find(
```
Pete-Y-CS: I think a name like `featureToBeUpdated` would be slightly cleary

src/lib/draw/GeometryMode.svelte
```diff
@@ -122,26 +122,45 @@
...
+  polygonTool.addEventListenerUpdated((feature) => {
+    if ($currentMode == thisMode) {
+      gjScheme.update((gj) => {
+        let updateFeature = gj.features.find((f) => f.id == currentlyEditing)!;
+        updateFeature.geometry = feature.geometry;
+        return gj;
+      });
+    }
+  });
+  // TODO Maybe use one callback (here and elsewhere); the type-safety is not really helpful
```
Pete-Y-CS: I like this idea. Seems like we're gonna be doing the exact same thing in 3+ places which is code smelly

src/lib/draw/GeometryMode.svelte
```diff
@@ -123,38 +123,26 @@
...
-        updateFeature.properties.waypoints = feature.properties.waypoints;
-        return gj;
-      });
-    }
-  });
+  for (let tool of [polygonTool, routeTool]) {
+    tool.addEventListenerUpdated((feature) => {
+      if ($currentMode == thisMode) {
+        gjScheme.update((gj) => {
+          let updateFeature = gj.features.find(
```
Pete-Y-CS: As before, `featureToBeUpdated` seems a little clearer to me

src/lib/draw/route/RouteMode.svelte
```diff
@@ -36,6 +40,17 @@
   }
   export function stop() {
     routeTool?.stop();
+
+    // If we leave this mode without saving, still create a new feature
```
Pete-Y-CS, line 44: wonder if, once again we could refactor to share some identical code

Pete-Y-CS (changes requested):
Requesting  small rename and a refactor similar to one done earlier in the PR

src/lib/draw/GeometryMode.svelte
dabreegster: This is from an intermediate commit; this file has been changed since then. I've made the equivalent change in the latest version though

src/lib/draw/GeometryMode.svelte
dabreegster: A later commit in this PR does this

src/lib/draw/common.ts
```diff
@@ -0,0 +1,70 @@
...
+  tool.addEventListenerFailure(() => {
+    if (get(currentMode) == thisMode) {
+      unsavedFeature.value = null;
+      changeMode("edit-attribute");
+    }
+  });
+}
+
+// If a user leaves a mode after drawing something valid but not saving, save it anyway.
+export function handleUnsavedFeature(
```
dabreegster, line 56: I refactored this function, so now the `stop()` in 3 modes is very simple

src/lib/draw/common.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      unsavedFeature.value = null;
+
+      changeMode("edit-attribute");
+      formOpen.set(feature.id as number);
+    }
+  });
+
+  tool.addEventListenerUpdated((feature) => {
+    if (get(currentMode) == thisMode) {
```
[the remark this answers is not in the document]
Pete-Y-CS: Hmm, maybe our event management needs further finessing! good find though!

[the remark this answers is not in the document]
dabreegster: Yeah, the event listening could also plumb through success/update/failure for the tools. #174 is a (very old!!) attempt to centralize handling and clean up the callbacks upon component destruction there. Not immediate urgency, but part of #228

tests/modes.spec.ts
```diff
@@ -261,6 +290,21 @@ test("the viewport changes only once when opening a form", async () => {
...
+  await clickMap(page, 400, 500);
+  await clickMap(page, 400, 600);
+
+  await page.getByRole("button", { name: "New route" }).click();
+  await expect(
+    page.getByRole("button", { name: "1) Untitled area" })
+  ).toBeVisible();
+
+  await page.getByRole("button", { name: "New polygon (snapped)" }).click();
+  await expect(page.getByText("1 objects")).toBeVisible();
```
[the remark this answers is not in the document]
Pete-Y-CS: More resilient, too, in case there is a delay in creation of the area/route object

Pete-Y-CS (approved):

tests/modes.spec.ts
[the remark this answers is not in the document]
dabreegster: >  in case there is a delay in creation of the area/route object

Good point. Playwright auto-await is smart about polling for the existence of locators, but not sure how it's possible for it to confirm non-existence of something other than just sleeping

Issue #254, opened 2023-06-27
Handling schema changes mid-edit


robinlovelace-ate:
I know the schema dropdown is experimental but still think it's worth exploring:

- [ ] Worth notifying users that their edits will re-appear when they go back to the schema in which edits were made if they change schema after adding objects?
- [ ] Can objects edited with one schema be imported into a session that is using a different schema (I guess I should try, writing in haste sorry)?

dabreegster:
Re-thinking the schema dropdown, maybe it better belongs as a big modal that lists the other schemas, describes their purpose, and makes it clear the entire page/app is about to change substantially. Or maybe we don't even have a dropdown, and direct people for each use case using a URL we manually give them right now.

> Worth notifying users that their edits will re-appear

It could be. Similar to how "clear all" today has a confirmation modal, we could use this as a place to let people know what's going on.

> Can objects edited with one schema be imported into a session that is using a different schema

No, by design, because the attributes per object are different. No known use case for doing this yet.

robinlovelace-ate:
> Or maybe we don't even have a dropdown, and direct people for each use case using a URL we manually give them right now.

That one sounds better to me at this stage: each use case is so domain specific it deserves its own URL and I can rarely/ever imagine users needing to switch mid edits. So big +1 to encoding schema in URL not dropdown from me.

robinlovelace-ate:
That would fix the issue for me. I think there may still need to be documentation and defensive programming that catches situations in which people accidentally try to upload edits made in one schema to another but that's probably capture in another issue. Another thing I like about the URL-based schema selection: simpler UX and hides away that dropdown that can could cause confusion.

robinlovelace-ate:
Yes there should be a heading, e.g. "Scheme design mode", "Development planning mode", "Critical issue mode", "Experimental mode" where we can test solutions to #231

Issue #242, opened 2023-06-19
Allowing users to select intervention types from the ATF4 guidance


robinlovelace-ate:
- [ ] New segregated cycling facility
- [ ] New junction treatment
- [ ] New permanent footway
- [ ] New shared use (walking and cycling) facilities
- [ ] Improvements to make an existing walking/cycle route safer
- [ ] Area-wide traffic management (including by TROs - both permanent and experimental)
- [ ] Bus priority measures that also enable active travel (for example, bus gates)
- [ ] Provision of secure cycle parking facilities
- [ ] New road crossings
- [ ] Restriction or reduction of car parking availability
- [ ] School streets

<details>

- New segregated cycling facility
    - Sub-category: High complexity
        - What does this scheme sub-category look like? Urban, high density, complex junctions, side roads
        - How will this be measured? Miles of cycle route, number of junctions

    - Sub-category: Medium complexity
        - What does this scheme sub-category look like? Suburban, medium density, fewer junctions/turning movements
        - How will this be measured? Miles of cycle route, number of junctions

    - Sub-category: Low complexity
        - What does this scheme sub-category look like? Out of town location, low density, few/no junctions
        - How will this be measured? Miles of cycle route, number of junctions

- New junction treatment
    - Sub-category: High complexity
        - What does this scheme sub-category look like? Separation in time and space for all active travel movements, protected junctions.
        - How will this be measured? Junctions treated

    - Sub-category: Medium complexity
        - What does this scheme sub-category look like? Protection of key movements for walking and cycling across a junction.
        - How will this be measured? Junctions treated

    - Sub-category: Low complexity
        - What does this scheme sub-category look like? Minor advantages to enable defensive positioning.
        - How will this be measured? Junctions treated

- New permanent footway
    - Sub-category: High complexity
        - What does this scheme sub-category look like? Large-scale town centre pedestrianisation including area-wide traffic and car parking removal
        - How will this be measured? Miles of footway

    - Sub-category: Medium complexity
        - What does this scheme sub-category look like? Conversion of carriageway to footway on a medium to large scale
        - How will this be measured? Miles of footway

    - Sub-category: Low complexity
        - What does this scheme sub-category look like? Addressing severance in existing walking routes
        - How will this be measured? Miles of footway

- New shared use (walking and cycling) facilities
    - Sub-category: Medium complexity
        - What does this scheme sub-category look like? Provision of a traffic-free rural or suburban route linking settlements as an alternative to hostile road conditions.
[...]

PR #256, opened 2023-06-28
Replace the schema dropdown with a title for the current mode. #254

dabreegster:
Screenshot from 2023-06-28 10-07-43

See issue for details. We don't want most users to change the schema right now; if someone's using something other than v1, we'll send them a URL with instructions. But all the same, if someone has used ATIP in multiple modes, we want to communicate somewhere which schema they're working with.

Changes to text styling or wording are welcome!

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -80,3 +80,12 @@ export type Mode =
   | "snap-polygon"
   | "split-route"
   | "street-view";
+
+export function schemaTitle(schema: Schema): string {
+  return {
+    v1: "Schema design",
+    v2: "Experimental schema design",
+    planning: "Development planning",
+    criticals: "Critical issues",
+  }[schema];
+}
--- a/src/types.ts
+++ b/src/types.ts
@@ -83,9 +83,9 @@ export type Mode =
 
 export function schemaTitle(schema: Schema): string {
   return {
-    v1: "Schema design",
-    v2: "Experimental schema design",
-    planning: "Development planning",
-    criticals: "Critical issues",
+    v1: "Scheme Design",
+    v2: "Experimental Schema Design",
+    planning: "Development Planning",
+    criticals: "Critical Issues",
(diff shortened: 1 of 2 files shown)


robinlovelace-ate:
Big 👍 to this. Links with #231 and the more specific #242 issues.

src/types.ts
```diff
@@ -80,3 +80,12 @@ export type Mode =
   | "snap-polygon"
   | "split-route"
   | "street-view";
+
+export function schemaTitle(schema: Schema): string {
+  return {
+    v1: "Schema design",
```
robinlovelace-ate, line 86:
```suggestion
    v1: "Basic design mode",
```

src/types.ts
```diff
@@ -80,3 +80,12 @@ export type Mode =
...
+    v2: "Experimental schema design",
```
robinlovelace-ate, line 87:
```suggestion
    v2: "Experimental design mode",
```

robinlovelace-ate (commented):
I think 'schema' is quite a big word to use here. Suggestion to use 'mode' instead: ATIP, like Vim, is a modal editor... Thoughts?

src/types.ts
Pete-Y-CS: Isn't the confusion that we should be calling it Scheme design? they're designing a scheme not a schema.

I'd say 
``` 
v1: Scheme Design
v2: Experimental Scheme Design
```
might be what we're looking for. 

Otherwise this change looks good

dabreegster: Ahh, what an awful typo -- I meant scheme, but we have two words in active use too similar >_<   Thank you, fixed. Also capitalized as suggested.

Screenshot from 2023-07-07 09-54-47

Pete-Y-CS (approved):

robinlovelace-ate (approved):

PR #258, opened 2023-06-29
WIP: add Gov uk frontend

Pete-Y-CS:


src/lib/forms/AutogenerateForm.svelte
```diff
@@ -51,7 +51,7 @@
 </script>
 
 {#if spec.description}
-  <p>{spec.description}</p>
+  <p class="govuk-body">{spec.description}</p>
```
dabreegster, line 54: I bet we could make every `<p>` do this centrally by using [global](https://svelte.dev/docs/svelte-components#style). Only in some places we may want a different paragraph style

dabreegster (commented):
I just took a quick skim -- looks pretty good, like a straightforward integration!

Pete-Y-CS:
Gonna address the comments Dustion made this afternoon but might be worth @Robinlovelace having a quick go for how this looks at https://acteng.github.io/atip/gov-uk-frontend/scheme.html?authority=Adur#12.83/50.84604/-0.29406

Robinlovelace:
> Gonna address the comments Dustin made this afternoon but might be worth @Robinlovelace having a quick go for how this looks at https://acteng.github.io/atip/gov-uk-frontend/scheme.html?authority=Adur#12.83/50.84604/-0.29406

Woo will take a look, thanks for pinging my non work GitHub handle (that I'm actually checking on non ATE day ; )

Robinlovelace:
First impressions: looks great to me!

image

src/lib/forms/AutogenerateForm.svelte
Pete-Y-CS: I don't think we can do this without copying the stylings manually, which I don't think would be good practice. I think we can use global to apply styling to e.g. each `<p>`, but we I don't think we can use it to apply classes to each `<p>`

Pete-Y-CS: I have reduced number of `class="govuk-body"`, though by simply applying it something higher in the DOM hierarchy. I assume we basically want everythign to have the gov uk font so it seems easiest and valid

dabreegster: I think we should find a way to avoid having to write `class` for all/most elements in our code. Two reasons:

1) It's easy to miss one spot, or forget to add it to something new
2) It adds visual noise and violates DRY

Ideally there'd be a way to tell CSS "select all p elements and apply the govuk-body class to them." I haven't found a way to do that yet...

A backup plan is to define our own `<P>`, `<Button>`, etc Svelte components, like https://github.com/acteng/atip/blob/main/src/lib/common/ExternalLink.svelte. It'd be annoying to have to import those in a bunch of places, but is better IMO than writing `class` everywhere.

There's a Slack for asking questions, right? Or we could search / ask on https://github.com/alphagov/govuk-frontend about common practices here.

dabreegster:
This PR adds `src/style` with things that look auto-generated. I'd expect those to be in `node_modules` generally. Did you have to deviate from https://frontend.design-system.service.gov.uk/get-started/ to get things working? How do we update/maintain this file?

src/lib/forms/AutogenerateForm.svelte
Pete-Y-CS: Hmm interesting. Sounds like a nice solution for button for sure. I think writing `class="govuk-body"` or similar at the top level of most components might not be too bad... But again, maybe we just want everything inside a `GovUkBody.svelte` component.

Pete-Y-CS:
> This PR adds `src/style` with things that look auto-generated. I'd expect those to be in `node_modules` generally. Did you have to deviate from https://frontend.design-system.service.gov.uk/get-started/ to get things working? How do we update/maintain this file?

I haven't had to deviate too far I don't think. It's true that I've added a CSS file which is compiled to the repo, which is probably not ideal. This is compiled from the `.sass` file which imports the govuk frontend stuff. I think the plan should be to change our github build process to include sass compilation.

dabreegster:
> I think the plan should be to change our github build process to include sass compilation

We could add an NPM script like https://github.com/ADD-William-WaltersDavis/planning_tool/blob/4f7abb0f194952072df3ccba08eff9a5e9cbe660/package.json#L12C43-L12C58, add it to the developer docs as one-time setup, and also run it in the Github workflows. We could go even further and add the generated file to `.gitignore` if we want to be sure we don't accidentally edit it manually, but that might be overkill; rerunning the script should produce no change in git

Pete-Y-CS:
> > I think the plan should be to change our github build process to include sass compilation
> 
> We could add an NPM script like https://github.com/ADD-William-WaltersDavis/planning_tool/blob/4f7abb0f194952072df3ccba08eff9a5e9cbe660/package.json#L12C43-L12C58, add it to the developer docs as one-time setup, and also run it in the Github workflows. We could go even further and add the generated file to `.gitignore` if we want to be sure we don't accidentally edit it manually, but that might be overkill; rerunning the script should produce no change in git

So I think I've updated the .gitignore, the github builds, all that, but I think the way our build script doesn't work nicely with adding a new dependency on a branch, but not on all branches. Let's chat when you're back from the conf

dabreegster:
Looking into the build issues now...

This is shaping up quickly! In terms of merging, it looks like we need a rebase against `main`, and we should still decide if we want to put `class=` everywhere, wrap in small Svelte components, or do something else. Thoughts?

.github/workflows/web.yml
```diff
@@ -35,7 +35,7 @@ jobs:
               base=/atip/$branch
             fi
 
-            if npm run wasm-release && npm ci && npm run build --if-present -- --base=$base; then
+            if npm run wasm-release && npm run compile-styles && npm ci && npm run build --if-present -- --base=$base; then
```
dabreegster, line 38: As noted, this'll break on every single branch except this new one. How about we run this separately above and allow it to fail? After merging this PR and rebasing all others, we can make it part of the mandatory step.

But also, the script uses `sass`, which'll only be installed after doing `npm ci`. So we should invert the order in the `&&`s, and in the temporary version

```suggestion
            npm run ci; npm run compile-styles
            if npm run wasm-release && npm ci && npm run build --if-present -- --base=$base; then
```

dabreegster:
I just rebased; there were lots of merge conflicts. Good in the future for long-running PRs to rebase more incrementally!

Going to introduce small Svelte components to tackle the `class=` problem. I found https://github.com/govuk-react/govuk-react and https://github.com/PaulioRandall/svelte-gds-design-system as more inspiration / precedent

dabreegster:
Alright, I started a paragraph component wrapping `govuk-body` and a button component. Along the way, finding more things to work on:

- [ ] When you activate a tool, the "clear all" button breaks with the reminder paragraph appearing
- [ ] Newlines broke in the v2 form; radio buttons are bunched together
- [ ] SelectToolButton CSS can get much simpler, and we can maybe stop dynamically changing the image color
- [ ] We should consistently use the govuk styles everywhere (ie, missing in BrowseSchemes page right now) and have a way of enforcing that
- [ ] There are new browser console errors downloading some font; maybe static assets aren't fully working
- [ ] The heading sizes in Instructions broke; we may need to switch to the govuk classes for all of these too

I wouldn't rush to merge this and leave the codebase in a half-migrated state. We can keep iterating here; I'll get back to some of these tasks later.

Pete-Y-CS:
One small comment: we should add the sass cache to the gitignore I think

Pete-Y-CS:
One small comment: we should add the sass cache to the gitignore I think

dabreegster:
OK to close this one and clean up the branch now? Thanks a ton for blazing the trail and figuring out this integration

PR #260, opened 2023-07-05
Use externally defined progress bar and make the route loading more obvious.

Pete-Y-CS:


robinlovelace-ate:
Clearly a step forward. @chris-ate approved. 

Can we make the UI reactive, e.g. making the question

```
Average cycleway width (m)
```

If it's a route-related intervention?

Also, there is a many-to-many mapping between intervention types and geometry types I think. Could we incorporate that also at some stage?

We can worry about all this later, big thumbs up to merging this so it's available, with a different url extension right?

Sorry for all the follow-on questions and great work on this!

robinlovelace-ate (approved):

package-lock.json
dabreegster, line 1: Could we revert the changes here please? Guessing this was a rebase problem. Without changes to package.json, I wouldn't expect the lock file to change generally

src/lib/draw/route/RouteMode.svelte
```diff
@@ -61,11 +68,58 @@
       changeMode
     );
   });
+
+  async function fetchWithProgress(url, setProgress: (number) => void) {
```
dabreegster, line 72:
```suggestion
  // setProgress will generally be called with a number between 0 and 100, but it may exceed this for gzipped files.
  // The server must send back a Content-Length header.
  async function fetchWithProgress(url: string, setProgress: (number) => void) {
```

src/lib/draw/route/RouteMode.svelte
```diff
@@ -61,11 +68,58 @@
...
+    progress = 100;
+    routeToolReady = true;
+  }
 </script>
 
-{#if !routeTool}
-  <!-- TODO the text should be fixed, and the progress bar float -->
-  <div bind:this={progress}>Route tool loading...</div>
+{#if !routeToolReady && !failedToLoadRouteTool && !downloadComplete}
+  <label for="route-loading">Route tool loading</label>
```
dabreegster, line 110: Elsewhere we've been doing `<label>Route tool loading <progress value={progress} />`, the second form mentioned by https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label, to avoid making up an ID. Not too opinionated about it here

src/lib/draw/route/RouteMode.svelte
```diff
@@ -61,11 +68,58 @@
...
+  <progress id="route-loading" value={progress} />
+{:else if downloadComplete && !routeToolReady && !failedToLoadRouteTool}
+  <label for="route-unpacking">Route data unpacking</label>
+  <progress id="route-unpacking" />
```
dabreegster, line 114: I really like the default animation of bouncing back and forth!

So the strategy is:
1) Show a 0-100 progress bar while we're downloading, up to the Content-Length of the .gzip file
2) We'll exceed 100%, because the browser is unpacking and passing along more than that Content-Length (and we have no way of knowing the uncompressed size, unless we store and send back a second header somehow)
3) As soon as we hit 100%, we switch to the bouncey animation and say unpacking. We might still be loading from the network for some of this, but it's fine.

dabreegster (approved):
Looks great to me, after fixing the 2 requested changes!

A future idea is to extract the fetch logic and all the error / loading / done states into some kind of dedicated Svelte component. It could be agnostic to what's being loaded, except for understanding this Content-Length + gzip + extra unpack complication. Will probably need a bunch of callback plumbing, so not really important now

PR #261, opened 2023-07-06
Use cursors in most tools

dabreegster:
A followup to #246. Change the cursor for most of our tools to indicate the current interactions possible on the map. Showing a few off, but easiest to just try at https://acteng.github.io/atip/curses_and_cursors/scheme.html?authority=Adur:

https://github.com/acteng/atip/assets/1664407/346cce17-8ab2-495f-9f8d-9b10a51d95d7

In general I tried to mimic design choices made by geojson.io (mapbox-gl-draw), Felt, terra-draw.

### DIFF
--- a/src/lib/draw/HoverLayer.svelte
+++ b/src/lib/draw/HoverLayer.svelte
@@ -13,6 +13,7 @@
   import {
     formOpen,
     gjScheme,
+    isAToolInUse,
     map,
     mapHover,
     sidebarHover,
@@ -20,7 +21,10 @@
 
   // Show clickable objects on the map using the cursor
   $: {
-    $map.getCanvas().style.cursor = $mapHover ? "pointer" : "grab";
+    // Don't override what an active tool has set
+    if (!$isAToolInUse) {
+      $map.getCanvas().style.cursor = $mapHover ? "pointer" : "grab";
+    }
   }
 
   let source = "hover";
--- a/src/lib/draw/HoverLayer.svelte
+++ b/src/lib/draw/HoverLayer.svelte
@@ -23,7 +23,7 @@
   $: {
     // Don't override what an active tool has set
     if (!$isAToolInUse) {
-      $map.getCanvas().style.cursor = $mapHover ? "pointer" : "grab";
+      $map.getCanvas().style.cursor = $mapHover ? "pointer" : "inherit";
     }
   }
 
--- a/src/lib/draw/StreetViewMode.svelte
+++ b/src/lib/draw/StreetViewMode.svelte
@@ -1,11 +1,5 @@
 <script lang="ts">
-  import type { Feature, Point } from "geojson";
-  import type { GeoJSONSource, MapMouseEvent } from "maplibre-gl";
-  import {
-    emptyGeojson,
-    overwriteCircleLayer,
-    overwriteSource,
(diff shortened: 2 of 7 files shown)


src/lib/draw/HoverLayer.svelte
```diff
@@ -13,14 +13,18 @@
...
     mapHover,
     sidebarHover,
   } from "../../stores";
 
   // Show clickable objects on the map using the cursor
   $: {
-    $map.getCanvas().style.cursor = $mapHover ? "pointer" : "grab";
+    // Don't override what an active tool has set
+    if (!$isAToolInUse) {
+      $map.getCanvas().style.cursor = $mapHover ? "pointer" : "inherit";
```
dabreegster, line 26: `inherit` instead of `grab` to let maplibre set grab or grabbing as appropriate. Basically saying "we don't care"

src/lib/draw/StreetViewMode.svelte
```diff
@@ -15,38 +9,18 @@
...
 
-  const circleRadiusPixels = 10;
-
   export let changeMode: (m: Mode) => void;
 
-  export function start() {}
-  export function stop() {
-    cursor = null;
+  export function start() {
+    $map.getCanvas().style.cursor = "zoom-in";
```
dabreegster, line 15: Maybe this is a strange icon to use for street view? Open to suggestions. Built-in options at https://developer.mozilla.org/en-US/docs/Web/CSS/cursor, and we can use any custom icon.

src/lib/draw/StreetViewMode.svelte
```diff
@@ -15,38 +9,18 @@
...
   }
-
-  let cursor: Feature<Point> | null = null;
-
-  // Rendering
```
dabreegster, line 29: Manually drawing points that followed the cursor was always a bit laggy and verbose code; nice to get rid of this

src/lib/draw/route/SplitRouteMode.svelte
```diff
@@ -25,40 +26,43 @@
...
+  }
   export function stop() {
-    cursor = null;
+    $map.getCanvas().style.cursor = "inherit";
+    snappedCursor = null;
     snappedIndex = null;
   }
 
-  let cursor: Feature<Point> | null = null;
+  let snappedCursor: Feature<Point> | null = null;
```
dabreegster, line 39: The behavior in this tool is hopefully intuitive. We always show the scissors icon where the user's cursor is. When they're close enough to snap to a route and make the action become valid, we draw a solid circle to show where the split will happen.

dabreegster (commented):
If these changes look good, I'll carry on with the final piece -- doing the same thing in the route-snapper.

src/lib/draw/StreetViewMode.svelte
Pete-Y-CS: I can't see a better one listed

Pete-Y-CS (approved):
Looks like this takes more lines than it adds and doesn't break anything, easy approve!

src/lib/draw/route/SplitRouteMode.svelte
Pete-Y-CS: This is super nice. Seems really intuitive to me and quite elegant

PR #262, opened 2023-07-10
Add a schema based on ATF4 guidance

dabreegster:
A minimal take on #242. The UX of selecting one of 11 options is not nice, and if we wanted to visualize these categorically on the map, we'd need a discrete palette with 11 choices. Purpose of this is just to raise questions about how we should better collect this type of data.
Screenshot from 2023-07-10 10-46-14
https://acteng.github.io/atip/atf4_schema/scheme.html?authority=Adur

### DIFF
--- a/src/schemas/atf4.ts
+++ b/src/schemas/atf4.ts
@@ -1,0 +1,20 @@
+// This file is auto-generated; do not manually edit
+
+export interface ATF4Intervention {
+  name?: string;
+  description?: string;
+  type?: type;
+}
+
+export type type =
+  | "New segregated cycling facility"
+  | "New junction treatment"
+  | "New permanent footway"
+  | "New shared use (walking and cycling) facilities"
+  | "Improvements to make an existing walking/cycle route safer"
+  | "Area-wide traffic management (including by TROs - both permanent and experimental)"
+  | "Bus priority measures that also enable active travel (for example, bus gates)"
+  | "Provision of secure cycle parking facilities"
+  | "New road crossings"
+  | "Restriction or reduction of car parking availability"
+  | "School streets";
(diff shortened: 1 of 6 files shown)


robinlovelace-ate:
See comments from #260 and apologies for confusion, commented on wrong PR I think (I thought it was a LOT of changes for adding new schema)!

robinlovelace-ate (approved):
`atf4` is a good name for this. Could we have `atf4-basic` and `atf4-detailed-1` for inpections people to test?

Clearly a step forward and I recall @chris-ate approved during demo yesterday. 

Can we make the UI reactive, e.g. making the question

```
Average cycleway width (m)
```

If it's a route-related intervention?

Also, there is a many-to-many mapping between intervention types and geometry types I think. Could we incorporate that also at some stage?

We can worry about all this later, big thumbs up to merging this so it's available, with a different url extension right?

Sorry for all the follow-on questions and great work on this!

dabreegster:
+ Pete, ready for review. I was going to add some questions from the table above -- I think each one just maps onto the types, in order (eg "Area-wide traffic management" would ask "Size of area improved for walking/cycling") but many of these are just derived from the geometry, or don't clearly map onto one geographic object drawn ("New junction treatement" and "Junctions treated" -- draw one object per junction?)

The intent is just to share for feedback and iterate

src/schemas/atf4.ts
Pete-Y-CS, line 1: Maybe we should be .gitignoring these?

dabreegster: They are auto-generated, so you're right. Done!

Pete-Y-CS: Might need to add something to the dev guide if we're removing some compiled elements? a new build step for new devs?

Pete-Y-CS (approved):

Issue #252, opened 2023-06-27
Reduce coordinate precision


dabreegster:
Our output GeoJSON is unnecessarily large today with too many decimal places. GJ spec recommends fewer. We can get much smaller file sizes as a result. This should be a relatively straightforward fix to the draw tools in this repo, and another to route-snapper. We could also consider adding this to backfill scripts, though it doesn't seem too important.

https://github.com/perrygeo/geojson-precision has more context/examples. I don't think we need to depend on this package or any dependency to do the trimming, just listing it as reference

dabreegster:
https://wiki.openstreetmap.org/wiki/Precision_of_coordinates for more reference

robinlovelace-ate:
This should make ATIP files even smaller and snappier to share. Great work!

PR #263, opened 2023-07-10
Trim coordinate precision for drawing points and freehand polygons

dabreegster:
For #252. Trim to 6 decimal places recommended by the GJ spec.

https://github.com/dabreegster/route_snapper/pull/38 did this for the route and snapped polygon tool, but it's not included in ATIP yet (I'll make another few changes and then release a new NPM version). This PR does it for the point, freehand polygon tool, and when splitting routes.

I thought about also doing this transformation in the backfill routine (when we load a GJ file from somewhere), but I don't think it's as important. If/when we have a story about storing official submitted data in one central place, we could do the transformation there.

Tested by drawing/splitting stuff, exporting, and manually checking the file

### DIFF
--- a/src/maplibre_helpers.ts
+++ b/src/maplibre_helpers.ts
@@ -184,14 +184,14 @@ export function pointFeature(pt: Position): FeatureWithProps<Point> {
     properties: {},
     geometry: {
       type: "Point",
-      coordinates: trimPrecision(pt),
+      coordinates: setPrecision(pt),
     },
   };
 }
 
 // Per https://datatracker.ietf.org/doc/html/rfc7946#section-11.2, 6 decimal
 // places (10cm) is plenty of precision
-export function trimPrecision(pt: Position): Position {
+export function setPrecision(pt: Position): Position {
   return [Math.round(pt[0] * 10e6) / 10e6, Math.round(pt[1] * 10e6) / 10e6];
 }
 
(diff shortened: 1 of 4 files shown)


src/maplibre_helpers.ts
```diff
@@ -171,6 +178,23 @@ export function emptyGeojson(): FeatureCollection {
...
+    geometry: {
+      type: "Point",
+      coordinates: trimPrecision(pt),
+    },
+  };
+}
+
+// Per https://datatracker.ietf.org/doc/html/rfc7946#section-11.2, 6 decimal
+// places (10cm) is plenty of precision
+export function trimPrecision(pt: Position): Position {
```
Pete-Y-CS, line 194: Can we call this `setPrecision`, `trimPrecision` is a bit confusing because we're either trimming the numbers to a precision, or (which seems naturally) we're setting the precision of the numbers

Pete-Y-CS (approved):

Pete-Y-CS (changes requested):
previous review was meant to request changes:
requesting name change to `setPrecision`

src/maplibre_helpers.ts
dabreegster: Done! Yeah, `trimPrecision` sounds like it ought to take a second parameter for the specified level

PR #266, opened 2023-07-11
Add some dropdown filters to the scheme browsing page

dabreegster:
Inspired by yesterday, this is a small improvement to the scheme browse page. Beside free-text filtering, also filter by two particular fields in the input data:

https://github.com/acteng/atip/assets/1664407/a0c16b66-a3dd-4a41-9479-e53eeb69fc3f

My goal is to flesh out a few more improvements to this page ahead of next week's meetings, so we can explore more possible directions to take this.

To try it out, https://acteng.github.io/atip/browse_dropdown_filters/browse.html and load magicfoo.geojson (from emails/Sharepoint). Motivation to make progress on auth soon :)

### DIFF
--- a/src/pages/BrowseSchemes.svelte
+++ b/src/pages/BrowseSchemes.svelte
@@ -15,6 +15,7 @@
   import InterventionLayer from "../lib/draw/InterventionLayer.svelte";
   import FormElement from "../lib/govuk/FormElement.svelte";
   import SecondaryButton from "../lib/govuk/SecondaryButton.svelte";
+  import Select from "../lib/govuk/Select.svelte";
   import Legend from "../lib/Legend.svelte";
   import Map from "../lib/Map.svelte";
   import ZoomOutMap from "../lib/ZoomOutMap.svelte";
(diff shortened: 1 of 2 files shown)


src/pages/BrowseSchemes.svelte
```diff
@@ -39,24 +39,43 @@
   // TODO Extract into a component
   let counts = { area: 0, route: 0, crossing: 0, other: 0 };
 
+  // Dropdown filters
```
dabreegster, line 42: Reminder that TS checks are disabled for this file, and everything in here is a bit of a hack (not split into distinct components for one). I have ideas about improving the code, but would like a little confidence we'll end up using/developing this before investing the effort

robinlovelace-ate (approved):
Great new feature. You're pumping out functionality and this bodes well for the future of ATIP :rocket: 

Additional thought: this will make it easy to filter based on other variables if we allow uploads of additional scheme level data from pulse.

src/pages/BrowseSchemes.svelte
Pete-Y-CS: In that case maybe this can remain an experimental branch rather than merging it in?

dabreegster: Hmm, do you mean moving the current browse page out of `main`? I would prefer keeping it in main and continuing to iterate on it (also in the govuk CSS PR). I think by early next week, we'll have a better idea how important this page is. If it's unlikely we'll use it, we can remove it entirely, and if more features will get added in the short term, we can improve the code. But before then, I'm trying to clean the page up before some meetings, and make it look like govuk style

Pete-Y-CS: Yeah I suppose since there's already plenty of this on main we can just price in a refactor down the line. I did mean just not merging these changes in before Monday but I guess given we want to show the gov uk stylings as well it's worth

dabreegster: Ah, this PR is intentionally here for a Monday demo, to make possible ideas around an inspector view more tangible :)  So if you have other ideas for this page that would be helpful to prototype quickly or even just sketch up / discuss, I'd say they're worth it

dabreegster:
I rebased against the govuk change and switched to https://design-system.service.gov.uk/components/select/. Because the 3 filter controls take up more space now (the govuk style seems to be to have a label and text input field on separate lines), I grouped the filter controls into a collapsible card. PTAL @Pete-Y-CS

src/pages/BrowseSchemes.svelte
```diff
@@ -129,6 +149,21 @@
...
+    let set2 = new Set();
+    for (let x of schemes) {
+      if (x.authority_or_region) {
+        set1.add(x.authority_or_region);
+      }
+      if (x.funding_programme) {
+        set2.add(x.funding_programme);
+      }
+    }
+    authorities = Array.from(set1.entries());
```
dabreegster, line 163: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/entries
Option value and display text may differ, so `Select` takes `[string, string][]`, and `entries` conveniently does this for us

Pete-Y-CS (approved):

PR #267, opened 2023-07-11
Get govuk installed, following

dabreegster:
https://frontend.design-system.service.gov.uk/get-started. Very lightly change the browse scheme page typography using Sass, to preview the next steps for integration.

This is an alternative to #258. Because switching is going to take a few steps, I propose we do it gradually. The build setup is subtle enough to deserve its own step. After this PR, I propose we update the two simple pages (scheme browsing and choosing an area) to completely use govuk styles, figuring out some of the patterns for using Sass vs Svelte components. Then tackle the much more complex main page.

Preview: https://acteng.github.io/atip/govuk_frontend_v2/browse.html
The text style changes slightly:
Screenshot from 2023-07-11 18-46-56
I have also temporarily copied in https://design-system.service.gov.uk/components/accordion to test the JS integration e2e.

Great thanks for help in https://github.com/alphagov/govuk-frontend/issues/3940 figuring out how to use Sass to keep our code simple!

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "fmt": "npx prettier --write *.html tests/* src/*",
     "test": "npx playwright test",
     "check": "svelte-check --tsconfig ./tsconfig.json",
-    "generate-schema-ts": "for x in v2 planning criticals atf4; do npx ts-node --project tsconfig_tsnode.json --esm src/lib/forms/generate_ts.ts src/schemas/$x.json > src/schemas/$x.ts; done"
+    "generate-schema-ts": "for x in v2 planning criticals atf4; do npx ts-node --project tsconfig_tsnode.json --esm src/lib/forms/generate_ts.ts src/schemas/$x.json > src/schemas/$x.ts; done",
+    "setup-govuk": "sass src/style/main.sass src/style/main.css; rm -rf public/assets; mkdir -p public/assets; cp -R node_modules/govuk-frontend/govuk/assets/ public/"
   },
   "devDependencies": {
     "@ianvs/prettier-plugin-sort-imports": "^4.0.2",
@@ -23,6 +24,7 @@
     "playwright": "^1.35.1",
     "prettier": "2.8.8",
     "prettier-plugin-svelte": "^2.10.1",
+    "sass": "^1.63.6",
     "svelte-check": "^3.4.4",
     "ts-node": "^10.9.1",
     "typescript": "^5.1.3",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -23,6 +24,7 @@
     "playwright": "^1.35.1",
     "prettier": "2.8.8",
     "prettier-plugin-svelte": "^2.10.1",
+    "sass": "^1.63.6",
```
dabreegster, line 27: Note I made this a dev dependency; we don't need to ship it to the user

src/pages/BrowseSchemes.svelte
```diff
@@ -17,6 +19,11 @@
   import { bbox } from "../maplibre_helpers";
   import { gjScheme, map } from "../stores";
 
+  onMount(() => {
+    // For govuk components. Must happen here.
```
dabreegster, line 23: Following https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#import-javascript-using-a-bundler, I originally tried calling it in `browse.html` outside of the Svelte stuff entirely, but the accordion example had no JS. No console errors or indication that something went wrong. If we call this in the component script outside of `onMount`, it also doesn't seem to work. So it must rely on part of the DOM being set up?

src/pages/BrowseSchemes.svelte
```diff
@@ -163,7 +170,7 @@
 </script>
 
 <Layout>
-  <div slot="sidebar">
+  <div slot="sidebar" class="sidebar">
```
dabreegster, line 173: The idea is that we'll opt into a bunch of style rules at once in a bigger section like this, then automatically change all `<p>` and `<h1>` beneath. Specifics TBD -- we'll figure out an HTML element <-> govuk class mapping in the next PRs.

dabreegster: Also, another thing to keep in mind: https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk#if-your-service-isnt-on-govuk
We're not on gov.uk yet, but the point of bringing in this design system is to move towards that point, so using the style and fonts and such seems like the correct step

dabreegster:
I copied in the accordion HTML temporarily to verify JS working:

https://github.com/acteng/atip/assets/1664407/e1f1c8ae-c555-4bd0-8b29-4a76a94268e0

robinlovelace-ate:
> I copied in the accordion HTML temporarily to verify JS working:
> 
>  screencast.mp4

This looks great!

Worth posting in the ATIP channel and tagging people who were asking for it?

Pete-Y-CS (approved):

Issue #223, opened 2023-06-05
Cache github CI build


dabreegster:
https://github.com/acteng/atip/actions shows the time for both actions is dominated by `npm run wasm-release`. We can use https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows to bring this down to ~seconds, I'll bet. Just need to puzzle through how to set the cache key... ideally there's a real example on github somewhere we can refer to

dabreegster:
wasm-opt is still slow, but this nothing else actionable right now

PR #268, opened 2023-07-12
Cache the WASM build in Github Actions. #223

dabreegster:
The deploy web workflow in https://github.com/acteng/atip/actions takes around 15 minutes lately, dependent on the number of open branches we've got. This PR uses [GH caches](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) to at least save the initial step of building the Rust route-info crate.

Much of the time is unfortunately spent in a `wasm-opt` step, which doesn't look like it's good at figuring out no source has changed and short-circuiting. This new cache step can't hurt, though.

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -23,6 +23,12 @@ jobs:
       - name: Install wasm-pack
         uses: jetli/wasm-pack-action@v0.4.0
 
+      - name: Cache WASM build
+        uses: actions/cache@v3
+        with:
+          path: route_info/target
+          key: doesnt-matter-share-everything
+
       - name: Build all branches
         run: |
           mkdir deployme
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -24,6 +24,12 @@ jobs:
     - name: Install dependencies
       run: npm ci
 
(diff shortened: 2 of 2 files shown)


dabreegster:
If desired, I can try and get approximate time savings by pushing some temporary commits in another branch to get the cache to be used

.github/workflows/web.yml
```diff
@@ -23,6 +23,12 @@ jobs:
       - name: Install wasm-pack
         uses: jetli/wasm-pack-action@v0.4.0
 
+      - name: Cache WASM build
+        uses: actions/cache@v3
+        with:
+          path: route_info/target
+          key: doesnt-matter-share-everything
```
dabreegster, line 30: The key can include things like the OS where this is running. We only build on one platform, so irrelevant. We could also hash the lockfile, but that doesn't really make sense either. If a commit changes or adds a dependency to the Rust crate, we can still reuse most of the cache. The rust build tools know how to rebuild only new things.

https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

Pete-Y-CS (approved):

PR #270, opened 2023-07-12
Convert the scheme browse and area choose page to govuk style

dabreegster:
This is the next step to use the govuk design system, on our two simpler pages. I've tried my best to interpret https://design-system.service.gov.uk/styles/ and https://design-system.service.gov.uk/components/, but I (really hope!) we go through some kind of design review later, because there are plenty of unclear decisions made.

Compare: [old homepage](https://acteng.github.io/atip/index.html) with [new homepage](https://acteng.github.io/atip/govuk_step2/index.html), and [old browse](https://acteng.github.io/atip/browse.html) with [new browse](https://acteng.github.io/atip/govuk_step2/browse.html)

https://github.com/acteng/atip/assets/1664407/181d3c35-6d86-4e11-93ef-0b540c08663f

In the step after this one, I'll convert our main edit page... or at least the sidebar. I haven't found much guidance yet for design patterns on maps, such as tooltips, changing cursors, hover effects, etc -- lots of the questions we've been asking anyway! I'll search/ask around to see if there's any guidance here yet.

# Code style choices

Copying the sample HTML directly from govuk docs would severely pollute our codebase, and make it tough to check that we've converted all elements on every page to using the right CSS classes. So there are 3 ways we use govuk components:

1) New Svelte components that wrap something, like a group of radio buttons. The user doesn't care that govuk things are happening internally.
2) Automatically adding CSS classes to standard HTML elements, as long as they're under a `<div class="govuk-prose">`
3) In "one-off" situations, directly use govuk HTML+CSS code. When the usage becomes more complex or we need the same pattern somewhere else, we can apply one of the two strategies above, but I feel it'd be premature refactoring at this point.

### DIFF
--- a/src/lib/common/FileInput.svelte
+++ b/src/lib/common/FileInput.svelte
@@ -1,5 +1,9 @@
 <script lang="ts">
+  import FormElement from "../govuk/FormElement.svelte";
+
   export let label: string;
+  // This must be unique in the page
+  export let id: string;
   // Called with the file contents as text
   export let loadFile: (text: string) => void;
 
@@ -17,15 +21,13 @@
   }
 </script>
 
-<div class="govuk-form-group">
-  <label class="govuk-label" for="file-upload">{label}</label>
+<FormElement {label} {id}>
   <input
     bind:this={fileInput}
     on:change={onChange}
     {disabled}
     class="govuk-file-upload"
-    id="file-upload"
-    name="file-upload"
+    {id}
     type="file"
   />
-</div>
+</FormElement>
--- a/src/lib/common/FileInput.svelte
+++ b/src/lib/common/FileInput.svelte
@@ -1,7 +1,5 @@
 <script lang="ts">
   export let label: string;
-  // This must be unique in the DOM
-  export let uniqueId: string;
   // Called with the file contents as text
   export let loadFile: (text: string) => void;
 
@@ -19,30 +17,15 @@
   }
 </script>
 
-<!-- TODO Interactive elements inside a label are apparently invalid, but this works -->
-<label>
+<div class="govuk-form-group">
+  <label class="govuk-label" for="file-upload">{label}</label>
   <input
-    type="file"
-    id={uniqueId}
     bind:this={fileInput}
     on:change={onChange}
     {disabled}
+    class="govuk-file-upload"
+    id="file-upload"
+    name="file-upload"
+    type="file"
   />
-  <button type="button" on:click={() => fileInput.click()} {disabled}
-    >{label}</button
-  >
-</label>
-
-<style>
-  input[type="file"] {
-    cursor: pointer;
-
-    /* Make the input type=file effectively invisible, but still let browser accessibility stuff work */
-    width: 0.1px;
-    height: 0.1px;
-    opacity: 0;
-    overflow: hidden;
-    position: absolute;
-    z-index: -1;
-  }
-</style>
+</div>
--- a/src/lib/govuk/DefaultButton.svelte
+++ b/src/lib/govuk/DefaultButton.svelte
@@ -1,0 +1,9 @@
+<button
+  type="button"
+  class="govuk-button"
+  data-module="govuk-button"
+  {...$$props}
+  on:click
+>
+  <slot />
+</button>
--- a/src/lib/govuk/FormElement.svelte
+++ b/src/lib/govuk/FormElement.svelte
@@ -1,0 +1,12 @@
+<script lang="ts">
+  // The label to show
+  export let label: string;
+  // A unique (per page) ID of the form element being labelled. Something in
+  // this component's slot must have this ID.
+  export let id: string;
+</script>
+
+<div class="govuk-form-group">
+  <label class="govuk-label" for={id}>{label}</label>
+  <slot />
+</div>
--- a/src/lib/govuk/Radio.svelte
+++ b/src/lib/govuk/Radio.svelte
@@ -1,0 +1,34 @@
+<script lang="ts">
+  // A label for the entire group of radio buttons
+  export let legend: string;
+  // A unique (per page) ID of the radio group
+  export let id: string;
+  // A list of [value, label] representing the choices
+  export let choices: [string, string][];
+
+  // The current value
+  export let value: string;
+</script>
+
+<div class="govuk-form-group">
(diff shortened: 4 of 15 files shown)


src/lib/govuk/FormElement.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+  // The label to show
+  export let label: string;
+  // A unique (per page) ID of the form element being labelled. Something in
```
dabreegster, line 4: I wish we didn't have to make this up at all. Previously we've been doing `<label>xyz <input ... /></label>`, but this breaks the govuk styling by eliminating a line-break. So, this Svelte component slightly helps.

src/lib/govuk/Radio.svelte
```diff
@@ -0,0 +1,34 @@
+<script lang="ts">
```
dabreegster, line 1: This component is very necessary. If you check the HTML sample at https://design-system.service.gov.uk/components/radios/, it's so much code to copy. Wrapping in a Svelte component is quite similar to their Nunchucks system

src/lib/sidebar/About.svelte
```diff
@@ -6,67 +6,71 @@
 </script>
 
 <Modal title="About the Active Travel Infrastructure Platform" bind:open>
-  <p>
-    ATIP v2 is an
-    <ExternalLink href="https://github.com/acteng/atip"
-      >open source project</ExternalLink
-    > supported by Active Travel England and developed by:
-  </p>
+  <div class="govuk-prose">
```
dabreegster, line 9: Here's an example of auto-applying CSS classes to everything below. All the diff below is just introducing one level of nesting in the DOM

src/pages/BrowseSchemes.svelte
```diff
@@ -202,21 +206,17 @@
...
-              <li>
-                <button type="button" on:click={() => editScheme(scheme)}
-                  >Edit scheme</button
-                >
-              </li>
-            </ul>
+            <p>Authority or region: {scheme.authority_or_region}</p>
+            <p>Capital scheme ID: {scheme.capital_scheme_id}</p>
+            <p>Funding programme: {scheme.funding_programme}</p>
+            <div class="govuk-button-group">
```
dabreegster, line 212: Made two of the buttons side-by-side. I'm sure the contents of these expandable cards will keep evolving, so not changing much else yet

src/pages/ChooseArea.svelte
```diff
@@ -142,40 +151,52 @@
...
-  <h1>Welcome to ATIP v2</h1>
-  <button type="button" on:click={() => (showAbout = !showAbout)}>About</button>
-
-  <p>Select Transport Authority or Local Authority District:</p>
-  <div>
-    <input
-      data-testid="transport-authority"
-      list="authorities-list"
-      bind:value={inputValue}
+<div class="govuk-grid-row">
```
dabreegster, line 154: Using https://design-system.service.gov.uk/styles/layout/ to split the page into two halves

src/pages/ChooseArea.svelte
```diff
@@ -142,40 +151,52 @@
...
+    <SecondaryButton on:click={() => (showAbout = !showAbout)}
+      >About</SecondaryButton
+    >
+
+    <FormElement
+      label="Select Transport Authority or Local Authority District"
+      id="inputValue"
+    >
+      <input
+        class="govuk-input govuk-input--width-20"
```
dabreegster, line 166: Here's a case where it's simplest to just directly use govuk classes, rather than refactor a component yet. I couldn't find any guidance on text input fields with custom autocomplete like this.

src/pages/ChooseArea.svelte
```diff
@@ -142,40 +151,52 @@
...
+      <datalist id="authorities-list" bind:this={dataList} />
+    </FormElement>
+    <DefaultButton on:click={start} disabled={!validEntry}>Start</DefaultButton>
+
+    <hr />
+
+    <Radio
+      legend="Or pick from the map"
+      id="showBoundaries"
+      choices={[
```
dabreegster, line 181: This used to be a select/dropdown, but https://design-system.service.gov.uk/components/select/ recommends against, and I agree here

robinlovelace-ate:
Based on the video bit :+1: to this, looks really good. Great work Dustin.

robinlovelace-ate:
It gets away from the very vanilla feel, it may benefit from some post merge tweaks but to my eyes this is already a big improvement, even without the benefits of conformity from GDS perspective.

image

robinlovelace-ate (approved):

robinlovelace-ate:
Interested to hear what @Pete-Y-CS, @Sparrow0hawk or others think also..

src/lib/govuk/DefaultButton.svelte
```diff
@@ -0,0 +1,9 @@
+<button
+  type="button"
+  class="govuk-button"
```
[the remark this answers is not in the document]
Pete-Y-CS: Don't think this needs to hold up this merge but could be nice

src/lib/common/FileInput.svelte
```diff
@@ -19,30 +17,15 @@
   }
 </script>
 
-<!-- TODO Interactive elements inside a label are apparently invalid, but this works -->
-<label>
+<div class="govuk-form-group">
+  <label class="govuk-label" for="file-upload">{label}</label>
```
Pete-Y-CS, line 21: Shouldn't we be using the form element defined below here?

src/lib/govuk/DefaultButton.svelte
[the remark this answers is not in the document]
dabreegster: I thought about that, but not sure yet. https://design-system.service.gov.uk/components/button/ recommends avoiding multiple "default buttons" per page, so making it extremely clear what type of button is used seems nice. We could definitely do `<Button kind="default|secondary">` too to share a bit of code, but maybe it makes the places used a little harder to read.

src/lib/common/FileInput.svelte
dabreegster: Ah good point, thanks! Fixed. I also brought back the unique ID, because `<label>` will break otherwise. We don't have any page with multiple file inputs, but we'll need to when we do

Pete-Y-CS (approved):

PR #271, opened 2023-07-13
Use a different word for intervention per schema. #160

dabreegster:
See #160.

I think I got the tests, but slow internet at the moment, so can't run easily

### DIFF
--- a/src/lib/draw/AttributeMode.svelte
+++ b/src/lib/draw/AttributeMode.svelte
@@ -10,11 +10,12 @@
     mapHover,
     openFromSidebar,
   } from "../../stores";
-  import type { Mode } from "../../types";
+  import { schemaSingularNoun, type Mode, type Schema } from "../../types";
   import type { EventHandler } from "./event_handler";
 
   const thisMode = "edit-attribute";
 
+  export let schema: Schema;
   export let changeMode: (m: Mode) => void;
   export let eventHandler: EventHandler;
 
@@ -83,8 +84,8 @@
 
 {#if $currentMode == thisMode}
   {#if $formOpen}
-    <p>Edit attributes to the left, or click another object</p>
+    <p>Edit attributes to the left, or click {schemaSingularNoun(schema)}</p>
   {:else}
-    <p>Click an object to fill out its attributes</p>
+    <p>Click {schemaSingularNoun(schema)} to fill out its attributes</p>
   {/if}
 {/if}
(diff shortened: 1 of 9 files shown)


src/lib/draw/AttributeMode.svelte
```diff
@@ -83,8 +84,8 @@
 
 {#if $currentMode == thisMode}
   {#if $formOpen}
-    <p>Edit attributes to the left, or click another object</p>
```
dabreegster, line 86: I ditched "another", which maybe makes things less clear. Could have another variation of the singular noun function that omits the article, but was trying to avoid that :)

src/lib/draw/GeometryMode.svelte
```diff
@@ -269,6 +275,6 @@
   {:else if currentlyEditingControls == "route"}
     <RouteControls {routeTool} extendRoute={false} />
   {:else}
-    <p>Click an object to edit its geometry</p>
```
robinlovelace-ate, line 272: This is definitely an improvement and allows us to be more specific. Nice!

src/types.ts
```diff
@@ -92,3 +92,25 @@ export function schemaTitle(schema: Schema): string {
     atf4: "ATF4 Scheme",
   }[schema];
 }
+
+// The singular form of the object (with the article) managed by this schema
+export function schemaSingularNoun(schema: Schema): string {
```
robinlovelace-ate, line 97: So nice! I wonder if we could read this in from a .json config file to separate configuration from code as we did for the schemea.json files?

src/types.ts
```diff
@@ -92,3 +92,25 @@ export function schemaTitle(schema: Schema): string {
...
+    v1: "an intervention",
+    v2: "an intervention",
+    planning: "a development",
+    criticals: "a critical issue",
+    atf4: "an intervention",
+  }[schema];
+}
+
+// The plural form of the object managed by this schema
+export function schemaPluralNoun(schema: Schema): string {
```
robinlovelace-ate, line 108: Gota have the plural versions also! They look right to me.

robinlovelace-ate (approved):

src/types.ts
dabreegster: We could put these and the title in the same .json defining the schema, yes. Maybe it does fit nicely there. OK if we think about that and defer a refactor? Still unsure how much we want to invest in this approach generally, especially now that we're running against govuk style best practices with the expanding form so much.

robinlovelace-ate: > OK if we think about that and defer a refactor?

:+1: 

Yes this is a nice to have from my perspective.

Pete-Y-CS (approved):

PR #272, opened 2023-07-13
Handling wrong authority names

Pete-Y-CS:
https://github.com/acteng/atip/issues/131 work to resolve this issue


src/lib/common/ErrorMessage.svelte
dabreegster, line 1: Hmm, I guess we've now confused `lib/govuk` and `lib/common`. This is fine here; we can merge these directories later as cleanup

src/lib/common/data_getter.ts
```diff
@@ -0,0 +1,18 @@
+import type { FeatureCollection } from "geojson";
+import authoritiesUrl from "../../../assets/authorities.geojson?url";
+
+export async function getAuthoritiesData() {
+  const resp = await fetch(authoritiesUrl);
+  const body = await resp.text();
+  const json: FeatureCollection = JSON.parse(body);
+  return Promise.resolve(json);
```
dabreegster, line 8: Just `return json;` should be fine. "If the return value of an async function is not explicitly a promise, it will be implicitly wrapped in a promise." from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

src/lib/common/data_getter.ts
dabreegster, line 1: Thoughts on naming this `authorities.ts`?

src/lib/common/data_getter.ts
```diff
@@ -0,0 +1,18 @@
+import type { FeatureCollection } from "geojson";
+import authoritiesUrl from "../../../assets/authorities.geojson?url";
+
+export async function getAuthoritiesData() {
```
dabreegster, line 4: Can we add the return types? `: FeatureCollection`
(Hopefully we can get to enabling strict in tsconfig "soon" to catch things like this)

And how about naming it `getAuthoritiesGeojson` or `getAuthoritiesGj`? "Data" is a bit vague. The return type is helpful though (if your IDE shows it)

src/lib/common/data_getter.ts
```diff
@@ -0,0 +1,18 @@
...
+  const resp = await fetch(authoritiesUrl);
+  const body = await resp.text();
+  const json: FeatureCollection = JSON.parse(body);
+  return Promise.resolve(json);
+}
+
+export async function getAuthoritiesNameSet() {
```
dabreegster, line 11: See comment in `App.svelte`; I don't think we need this method at all

src/pages/App.svelte
```diff
@@ -82,6 +83,14 @@
     );
     return geojson;
   }
+
+  async function checkAuthorityValid(authorityName: string): Promise<void> {
+    let authortiesNameSet = await getAuthoritiesNameSet();
```
dabreegster, line 88: This will cause us to fetch and parse the authorities file twice. Inside `onMount`, we already do `boundaryGeojson = await loadAuthorityBoundary();`. Inside that function, we filter the GJ to look for the authority name. If the result is empty, we have an error -- so we don't even need the name set at all, actually

src/pages/ChooseArea.svelte
```diff
@@ -157,6 +158,9 @@
     <SecondaryButton on:click={() => (showAbout = !showAbout)}
       >About</SecondaryButton
     >
+    {#if errorMessage}
+      <ErrorMessage bind:errorMessage />
```
dabreegster, line 162: Just `{errorMessage}` is fine. `bind` is for two-way binding. Since `ErrorMessage` doesn't mutate this, we don't need two-way -- just pass in a regular prop
https://svelte.dev/tutorial/component-bindings

dabreegster (changes requested):
Could we also add a Playywright test for this please?

src/lib/common/data_getter.ts
Pete-Y-CS: Well I figured we might use it down the line for other stuff, as part of my drive to move more logic out of components 💪🏻

src/lib/common/ErrorMessage.svelte
Pete-Y-CS: Yeah I was thinking I could still try to resolve the other cases we talked about on this PR

src/lib/common/data_getter.ts
Pete-Y-CS: Should add type, and agree GeoJson is more specific and thererfore better!

src/pages/ChooseArea.svelte
Pete-Y-CS: Yeah I think I tried that at first but got red squiggles so did this for now, but maybe I got it slightly wrong somehow

Pete-Y-CS: Just tried it again, weirdly the red squiggle decides that error is a boolean. Even weirder, that becomes true.

I'll post screenshot of it with bind, and then without:
image


without:
image

Pete-Y-CS: Can probably resolve with black magic reactive statements

src/lib/common/data_getter.ts
dabreegster: Sounds good!

src/pages/ChooseArea.svelte
dabreegster: Even with the braces? `<Component errorMessage>` means "set errorMessage=true" and `Component {errorMessage}>` means "set errorMessage prop equal to the errorMessage variable"

dabreegster: https://svelte.dev/repl/72808c5cdce14fc6862e1661b68d8809?version=3.22.2

Pete-Y-CS: Sorry saw this during standup and decided not to respond at that moment. This resolves it yes cheers!

Pete-Y-CS:
I think I've resolved the other cases described in the issue. Just gotta write some tests!

src/lib/common/data_getter.ts
```diff
@@ -0,0 +1,8 @@
+import type { FeatureCollection } from "geojson";
+import authoritiesUrl from "../../../assets/authorities.geojson?url";
+
+export async function getAuthoritiesData(): Promise<FeatureCollection> {
```
dabreegster, line 4: Optional: don't need to say `Promise`, async implies it
"Async functions always return a promise. If the return value of an async function is not explicitly a promise, it will be implicitly wrapped in a promise. "
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -163,6 +165,9 @@
 <br />
 
 <div>
+  {#if errorMessage}
+    <ErrorMessage {errorMessage} />
```
dabreegster, line 169: Note the styling won't appear till after #274, since this page isn't wired up yet

src/pages/App.svelte
```diff
@@ -26,7 +26,6 @@
   let showInstructions = false;
 
   const params = new URLSearchParams(window.location.search);
-  // TODO Add validation and some kind of error page
```
dabreegster, line 29: I personally would leave this... we could later work on a bit of paranoia for other URL query params like `style` and `schema`. They probably warrant less intrusive handling, like just quietly switching to a default

src/pages/BrowseSchemes.svelte
```diff
@@ -108,7 +110,7 @@
 
       $map?.fitBounds(bbox(gj), { padding: 20, animate: false });
```
dabreegster, line 111: We need to blank out the error here, in case the user loads a bad file, then fixes by loading a good one

src/pages/ChooseArea.svelte
```diff
@@ -5,17 +5,20 @@
...
-  import authoritiesUrl from "../../assets/authorities.geojson?url";
+  import { getAuthoritiesData } from "../lib/common/data_getter";
   import FileInput from "../lib/common/FileInput.svelte";
   import About from "../lib/sidebar/About.svelte";
   import { bbox } from "../maplibre_helpers";
   import type { Schema } from "../types";
 
   let showAbout = false;
+  const params = new URLSearchParams(window.location.search);
+  let errorMessage: string = params.get("error")!;
```
dabreegster, line 21: The `!` insists that this isn't null, but that's not true here. `params.get("error") || ""` would work, since we're using the empty string to mean none

dabreegster (commented):
Looking good!
Screenshot from 2023-07-14 12-08-40

Should be good to go after a test and a few more small fixes

src/lib/common/data_getter.ts
Pete-Y-CS: I get red squiggles unless I put Promise, maybe an issue with my VSCode ts extension

tests/homepage.spec.ts
```diff
@@ -19,3 +19,14 @@ test("choosing a local authority and clicking start changes the url", async ({
     /.*scheme.html\?authority=West%20Yorkshire%20Combined%20Authority/
   );
 });
+
+test("Uploading a valid atip geojson redirects to the appropriate authority scheme page", async ({
```
dabreegster, line 23: Thanks for adding in the happy-path test case too!

dabreegster (approved):
Beautiful! Merge away

PR #275, opened 2023-07-14
Display user-friendly field names for auto-generated forms

### LINKED ISSUE #273: Capital letters for attribute addition panel headings
As illustrated below we still have lowercase headings.

This is fine for prototyping but if this is a quick fix I would advocate that as a low hanging fruit to improve feel and how polished it feels.

image

dabreegster:
Closes #273 
https://acteng.github.io/atip/capitalize_auto_form/scheme.html?authority=Adur&schema=v2

Screenshot from 2023-07-14 14-52-43

An incremental improvement. It'd also be nice to apply these to radio buttons, so we don't see `SchoolCrossing`. But we need to be careful with https://github.com/sindresorhus/humanize-string, because it mangles the dash in "20 - 30 mph"

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@types/geojson": "^7946.0.10",
         "comlink": "^4.4.1",
         "govuk-frontend": "^4.6.0",
+        "humanize-string": "^3.0.0",
         "maplibre-gl": "^3.1.0",
         "route-snapper": "^0.1.14",
         "svelte": "^4.0.0"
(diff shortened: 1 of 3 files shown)


Pete-Y-CS:
Looks good, only question is if it mangles one of our strings, maybe we want to manually write something since it's probably not much effort and it can be good to avoid adding too many package

Pete-Y-CS (approved):

dabreegster:
Yeah, agreed, thanks -- let's treat this as a tiny step forward and revisit next week

Pete-Y-CS:
> Manually writing might be better long-term. It'd make it harder to initially write schemas, if the human-visible label is required. Imagine adding something to everything in `schemas/*.json`... maybe it can be optional, and by default the actual value we stick in JSON can match the label. Thoughts?

I was actually thinking writing our own string tidier, but adding an optional human-friendly-label to the schema might not be a bad idea

PR #276, opened 2023-07-14
Make prettier handle closing tags better

dabreegster:
The code is much easier to read this way, personally.

See https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting. This can theoretically affect display. I'm combing through the changes in code / on-screen and don't spot any differences, so I think we're fine and this is worth it.

### DIFF
--- a/src/lib/draw/StreetViewMode.svelte
+++ b/src/lib/draw/StreetViewMode.svelte
@@ -58,9 +58,13 @@
   <CollapsibleCard label="Help">
     <ul>
       <li>
-        <b>Click</b> on the map to open a new tab with a 3rd-party imagery provider
+        <b>Click</b>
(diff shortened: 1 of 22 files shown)


src/lib/draw/StreetViewMode.svelte
```diff
@@ -58,9 +58,13 @@
   <CollapsibleCard label="Help">
     <ul>
       <li>
-        <b>Click</b> on the map to open a new tab with a 3rd-party imagery provider
+        <b>Click</b>
+         on the map to open a new tab with a 3rd-party imagery provider
```
dabreegster, line 62: Hmm, this is a little worse IMO, but worth the improvement for buttons

PR #277, opened 2023-07-14
Consolidate 3 layer-like controls

dabreegster:
Layer controls are often a map overlay, like in https://www.planning.data.gov.uk/map. We use that today for switching basemaps and showing a legend, but the speed limit layer is smushed into the sidebar for no good reason. Group everything and make the entire overlay hideable.

https://github.com/acteng/atip/assets/1664407/12fe6689-530e-4c73-ad09-7f575a24dfa7

### DIFF
--- a/src/lib/Legend.svelte
+++ b/src/lib/Legend.svelte
@@ -6,54 +6,42 @@
   export let schema: Schema;
 </script>
 
-<div>
-  <CollapsibleCard label="Objects" open>
-    <ul>
-      {#if schema == "planning"}
-        <li>
-          <ColorLegend color={colors.preapp} />
-          Preapp
-        </li>
-        <li>
-          <ColorLegend color={colors.outline} />
-          Outline
-        </li>
-        <li>
-          <ColorLegend color={colors["reserved matters"]} />
-          Reserved matters
-        </li>
-        <li>
-          <ColorLegend color={colors["local plan"]} />
-          Local plan
-        </li>
-      {:else}
-        <li>
-          <ColorLegend color={colors.area} />
-          Areas
-        </li>
-        <li>
-          <ColorLegend color={colors.route} />
-          Routes
-        </li>
-        <li>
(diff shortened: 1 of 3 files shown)


Pete-Y-CS:
I wonder if the name layers isn't quite right? Maybe it should be called map controls?

robinlovelace-ate:
Just saw this. Definite improvement I think, and makes important speed data more visible. Layers sounds find to me, maybe "Map layers" as opposed to data layers would be better?

src/lib/Legend.svelte
```diff
@@ -2,12 +2,25 @@
   import { colors } from "../colors";
   import type { Schema } from "../types";
   import CollapsibleCard from "./common/CollapsibleCard.svelte";
+  import Select from "./govuk/Select.svelte";
```
robinlovelace-ate, line 5: 👍  to gov.uk style legend

robinlovelace-ate (approved):

dabreegster:
Moving to draft, I'll pick this up after #282

dabreegster:
Please take another look; this PR now does two things:

- remove the legend from the browse page entirely; it's unhelpful there currently, and we're going to change to let people control the coloring by different variables
- On the editing page, group layer controls together. I decided to hide controls by default to make the UI simpler, and de-emphasize the 4-color legend.

https://github.com/acteng/atip/assets/1664407/d2bae816-f100-4f8d-a0c4-f3bbbae8190a

dabreegster:
Demo at https://acteng.github.io/atip/combine_layer_controls/scheme.html?authority=West of England Combined Authority

robinlovelace-ate:
Finally taking a look...

robinlovelace-ate:
Instant 👍 to integrated legend. This is a minor but IMO important change to the UI with UX benefits so definitely worth merging to keep momentum going. Thanks Dustin for the work and patience!

PR #278, opened 2023-07-16
Refactor the scheme browse page

dabreegster:
#173. This splits the monolithic Svelte component with TS checks disabled into a few logical components, and fixes all TS errors. Not much behavioral change yet, other than simplifying the map tooltips to only show the meaningful fields.

### DIFF
--- a/src/lib/browse/Filters.svelte
+++ b/src/lib/browse/Filters.svelte
@@ -1,0 +1,127 @@
+<script lang="ts">
+  import { onMount } from "svelte";
+  import { gjScheme } from "../../stores";
+  import CollapsibleCard from "../common/CollapsibleCard.svelte";
+  import FormElement from "../govuk/FormElement.svelte";
+  import SecondaryButton from "../govuk/SecondaryButton.svelte";
+  import Select from "../govuk/Select.svelte";
+  import { type Scheme } from "./data";
+
+  // These are immutable; re-create this component if they change
+  export let schemes: Map<string, Scheme>;
+
+  // by scheme_reference
+  export let showSchemes: Set<string> = new Set();
+
+  let filterText = "";
+
+  // Dropdown filters
+  let authorities: [string, string][] = [];
+  let filterAuthority = "";
+  let fundingProgrammes: [string, string][] = [];
+  let filterFundingProgramme = "";
+
+  // Stats about filtered schemes
+  let counts = { area: 0, route: 0, crossing: 0, other: 0 };
+
+  onMount(() => {
+    let set1 = new Set();
+    let set2 = new Set();
+    for (let x of schemes.values()) {
+      if (x.authority_or_region) {
+        set1.add(x.authority_or_region);
+      }
+      if (x.funding_programme) {
+        set2.add(x.funding_programme);
+      }
+    }
+    authorities = Array.from(set1.entries());
+    authorities.sort();
+    fundingProgrammes = Array.from(set2.entries());
+    fundingProgrammes.sort();
+  });
+
+  // When any filters change, update showSchemes
+  function filtersUpdated(
+    filterText: string,
+    filterAuthority: string,
+    filterFundingProgramme: string
+  ) {
+    showSchemes.clear();
+    let filterNormalized = filterText.toLowerCase();
+    for (let feature of $gjScheme.features) {
+      // TODO This is a very blunt free-form text search of any property in the
+      // feature only, not the scheme.
+      if (
+        filterNormalized &&
+        !JSON.stringify(feature.properties)
+          .toLowerCase()
+          .includes(filterNormalized)
+      ) {
+        continue;
+      }
+      if (filterAuthority || filterFundingProgramme) {
+        let scheme = schemes.get(feature.properties.scheme_reference);
+        if (filterAuthority && scheme.authority_or_region != filterAuthority) {
+          continue;
+        }
+        if (
+          filterFundingProgramme &&
+          scheme.funding_programme != filterFundingProgramme
+        ) {
+          continue;
+        }
+      }
+      showSchemes.add(feature.properties.scheme_reference);
+    }
+    // Make Svelte see the update
+    showSchemes = showSchemes;
+
+    // Hide things on the map, and recalculate stats
+    counts = { area: 0, route: 0, crossing: 0, other: 0 };
+    gjScheme.update((gj) => {
+      for (let feature of gj.features) {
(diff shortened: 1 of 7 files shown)


src/lib/browse/Filters.svelte
```diff
@@ -0,0 +1,127 @@
...
+  // When any filters change, update showSchemes
+  function filtersUpdated(
+    filterText: string,
+    filterAuthority: string,
+    filterFundingProgramme: string
+  ) {
+    showSchemes.clear();
+    let filterNormalized = filterText.toLowerCase();
+    for (let feature of $gjScheme.features) {
+      // TODO This is a very blunt free-form text search of any property in the
```
dabreegster, line 53: I also rewrote the filter logic, trying to be more performant. If you only change a dropdown, don't stringify all the properties unnecessarily

src/pages/BrowseSchemes.svelte
```diff
@@ -1,26 +1,21 @@
...
   import Layout from "../lib/common/Layout.svelte";
   import MapTooltips from "../lib/common/MapTooltips.svelte";
   import InterventionLayer from "../lib/draw/InterventionLayer.svelte";
   import ErrorMessage from "../lib/govuk/ErrorMessage.svelte";
-  import FormElement from "../lib/govuk/FormElement.svelte";
   import SecondaryButton from "../lib/govuk/SecondaryButton.svelte";
-  import Select from "../lib/govuk/Select.svelte";
   import Legend from "../lib/Legend.svelte";
-  import Map from "../lib/Map.svelte";
+  import MapLibreMap from "../lib/Map.svelte";
```
dabreegster, line 16: Awkward name because we need to use the built-in JS Map in this file

src/types.ts
```diff
@@ -49,6 +49,8 @@ export interface InterventionProps {
   v2?: Intervention;
   criticals?: CriticalIssue;
   atf4?: ATF4Intervention;
+  // An extra field present in input for the browse schemes page only
```
dabreegster, line 52: Still using `gjScheme`, and thus editing this type. The only reason is so we can reuse `InterventionLayers.svelte`. As requirements for this page evolve, maybe we revisit this too, and render the features differently on the map anyway. (Coloring by funding status, scheme, etc?)

src/lib/browse/Filters.svelte
```diff
@@ -0,0 +1,127 @@
...
+  import FormElement from "../govuk/FormElement.svelte";
+  import SecondaryButton from "../govuk/SecondaryButton.svelte";
+  import Select from "../govuk/Select.svelte";
+  import { type Scheme } from "./data";
+
+  // These are immutable; re-create this component if they change
+  export let schemes: Map<string, Scheme>;
+
+  // by scheme_reference
+  export let showSchemes: Set<string> = new Set();
```
Pete-Y-CS, line 14: Maybe `schemesToBeShown` would be a little more accurate

src/lib/browse/Filters.svelte
```diff
@@ -0,0 +1,127 @@
...
+
+  // When any filters change, update showSchemes
+  function filtersUpdated(
+    filterText: string,
+    filterAuthority: string,
+    filterFundingProgramme: string
+  ) {
+    showSchemes.clear();
+    let filterNormalized = filterText.toLowerCase();
+    for (let feature of $gjScheme.features) {
```
Pete-Y-CS, line 52: Can we use an actual `Array.filter`?
Something like: `schemesToBeShown = schemes.filter(getPredicate(filterText, filterAuthority, filterFundingProgramme))`

If we want to make it more performant we store previous filters somewhere and work out how our filters have changed and if we are simply getting more specific (e.g. filter authority stays the same but we add text or funding programme where we didn't have that before) in which case we can filter on the existing schemes being shown.

src/pages/BrowseSchemes.svelte
```diff
@@ -135,78 +48,19 @@
...
-    fundingProgrammes = Array.from(set2.entries());
-    fundingProgrammes.sort();
-  }
-
   function tooltip(props: { [name: string]: any }): string {
     // TODO Move into a Svelte component, so we don't have to awkwardly build up HTML like this
-    var html = `<table>`;
-    for (let [key, value] of Object.entries(props)) {
-      html += `<tr><td>${key}</td><td>${value}</td></tr>`;
+    var html = `<div class="govuk-prose" style="max-width: 30vw;">`;
```
Pete-Y-CS, line 53: This feels kinda weird. Looks like we're doing what Svelte is supposed to do natively in ts?

Pete-Y-CS (changes requested):

src/lib/browse/Filters.svelte
dabreegster: Done

src/lib/browse/Filters.svelte
dabreegster: Changed to use filter and map. I love the idea of filtering on prior results when the only change is appending to filterText -- that's a comm one! I haven't yet profiled to figure out where the (not so terrible) lag in filtering comes from yet -- maybe recalculating the set is fine, and updating maplibre is the slow part.

src/pages/BrowseSchemes.svelte
dabreegster: You're right. At the moment, it's a bit convoluted to make maplibre popups work with Svelte, because it needs this `setHTML` call. A nicer approach using `svelte-maplibre` looks [like this](https://github.com/dabreegster/geodiffr/blob/76c4f596a0659c1c9a192314063e117b842371ac/src/App.svelte#L142), where the popup component is nested logically under the layer, which is also expressed as a Svelte component. We're not using that library yet here (and it'll take some work to migrate to it, and I'm not sure it's the right move yet).

Alternatively, I've done a simpler hack in https://github.com/a-b-street/osm2streets/blob/147de2ec76aeb8397cb813dab8572a151a1eb7e8/web/src/street-explorer/Popups.svelte#L26 of instantiating a Svelte component manually. Should we do the same here?

dabreegster:
I rebased after the TS strict changes and addressed comments, PTAL

src/pages/BrowseSchemes.svelte
Pete-Y-CS: Alright, Either of those does seem nicer than this, but this PR isn't changing our approach so I don't think this should hold up this PR, approving

Pete-Y-CS (approved):

PR #279, opened 2023-07-17
Enable TS strict mode and enforce it with GH actions

### LINKED ISSUE #89: Port to TypeScript

dabreegster:
Closes #89.

In recent PRs, there have been some simple errors that TS would've caught, but we're not enforcing checks yet. My goal here is to catch new problems with as many compile-time warnings as possible. There are still some difficult cases for us to fix, but we can `@ts-ignore` them and treat them as TODOs.

### DIFF
--- a/src/lib/draw/Toolbox.svelte
+++ b/src/lib/draw/Toolbox.svelte
@@ -55,17 +55,19 @@
     "street-view": new EventHandler(),
   };
 
-  MapEvents.forEach((eventName) => {
+  for (let eventName of MapEvents) {
     $map.on(eventName, (event) => {
+      // @ts-ignore TODO Can't express that MapEvents are the keys
       eventHandlers[get(currentMode)].mapHandlers[eventName](event);
     });
-  });
+  }
 
-  DocumentEvents.forEach((eventName) => {
+  for (let eventName of DocumentEvents) {
     document.addEventListener(eventName, (event) => {
+      // @ts-ignore TODO Can't express that DocumentEvents are the keys
       eventHandlers[get(currentMode)].documentHandlers[eventName](event);
     });
-  });
+  }
 
   // This must be used; don't manually change mode
   function changeMode(newMode: Mode) {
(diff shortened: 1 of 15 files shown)


src/lib/draw/Toolbox.svelte
```diff
@@ -55,17 +55,19 @@
     "street-view": new EventHandler(),
   };
 
-  MapEvents.forEach((eventName) => {
+  for (let eventName of MapEvents) {
```
Pete-Y-CS, line 58: Out of interest why do we prefer `for` over `forEach`?

Pete-Y-CS (approved):

src/lib/sidebar/InterventionList.svelte
```diff
@@ -17,8 +17,11 @@
       return feature.properties.planning?.name || "Untitled polygon";
     }
     if (schema == "v2") {
+      // TODO Revisit auto-generated enum types
```
dabreegster, line 20: Still needs more work to figure this out, but I  don't think it's worth blocking better checks

src/pages/App.svelte
```diff
@@ -1,4 +1,5 @@
 <script lang="ts">
+  // @ts-ignore no declarations
```
dabreegster, line 2: There's some way to define a .d.ts file to get TS declarations for external libraries, but haven't figured out exactly how, and it's not too important -- we're just calling this one init function

src/stores.ts
```diff
@@ -11,7 +11,8 @@ import {
 import { type RouteInfo } from "./worker";
 
 // A global singleton, containing a loaded map
-// TODO | null. When we enable strictNullChecks, this'll become a problem
+// @ts-ignore TODO By construction, no components using the store should be
```
dabreegster, line 14: We could make all components using $map handle the possibility of not being loaded. https://github.com/dimfeld/svelte-maplibre/ does this, but it's complex. For now, I think carrying on with the assumption of it being loaded is OK

src/lib/draw/Toolbox.svelte
dabreegster: Ah, will revert. I was trying a bunch of keyof/indexable type things that didn't work out. We use regular for loops in most places in the code, so it's default muscle memory for me

dabreegster: Discussed offline - in this case, not likely to share a separate function body anywhere. We'll prefer forEach / separate functions + functional methods especially in cases like #278 with the filter. Will go fix that up now

PR #281, opened 2023-07-21
Mechanical refactor to group imports from common and govuk directories

dabreegster:
Makes using these common libraries less verbose. https://javascript.info/import-export#re-export talks about this approach.

### DIFF
--- a/src/lib/BaselayerSwitcher.svelte
+++ b/src/lib/BaselayerSwitcher.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import Select from "./govuk/Select.svelte";
+  import { Select } from "./govuk";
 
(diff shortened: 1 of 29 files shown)


Pete-Y-CS (approved):
This PR looks fine but as I think about the task of putting the bid appraisal process into an app I wonder if we wanna put these gov uk tools into a separate svelte library which our other apps can import from?

dabreegster:
We will definitely want to share these! This is a step towards that at least; we could publish an NPM packge and then just swap "../govuk" with "govuk-svelte". Not exactly sure how to name that package yet... I wouldn't want to suggest to anyone external to ATE that it's ready to use. https://github.com/PaulioRandall/svelte-gds-design-system is the one related library I've found, also to consider using / taking inspiration from

PR #282, opened 2023-07-23
Add a layer to the browse page for schools and hospitals

dabreegster:
Demo: https://acteng.github.io/atip/poi_layers/browse.html

https://github.com/acteng/atip/assets/1664407/168302df-06ad-44d7-a336-18d91c60add3


src/lib/Legend.svelte
dabreegster, line 1: Note I still want to carry on with #277. This PR doesn't change anything for the sketching page, but starts to consolidate controls for the browse page. I'll revive that PR after this one.

src/lib/browse/HospitalsLayer.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+  import OptionalLayer from "./OptionalLayer.svelte";
+</script>
+
+<OptionalLayer
+  name="hospitals"
+  singularNoun="hospital"
+  pluralNoun="Hospitals"
+  color="blue"
```
dabreegster, line 9: Consider the colors temporary. I'll followup soon with something to let the user control how interventions are styled on the scheme page -- grouping by funding program and other fields might be useful. Not sure yet how to make contextual layers like this interact nicely with interventions.

src/lib/common/BaselayerSwitcher.svelte
dabreegster, line 1: Refactoring this component to just handle the logic itself, not also decide where the controls on the map go. I've preserved the current behavior in App.svelte for the scheme editor

src/lib/common/ColorLegend.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
```
dabreegster, line 1: Refactoring this out

src/lib/govuk/Checkbox.svelte
```diff
@@ -1,19 +1,17 @@
 <script lang="ts">
   // This component must be placed inside a CheckboxGroup
 
-  export let label: string;
```
dabreegster, line 4: I changed the API of this component, because we want to show the little colored box next to the label. There was only one other user, in the route controls, and it's not harder to use

src/maplibre_helpers.ts
```diff
@@ -99,11 +112,13 @@ export function overwritePolygonLayer(
     id: string;
     source: string;
     filter?: FilterSpecification;
+    // For vector (non GeoJSON) sources
+    sourceLayer?: string;
```
dabreegster, line 116: Vector tile sources themselves can have multiple layers packed in. Even when there's just one, you have to specify it. For the pmtiles we built, they match the filename -- schools.pmtiles -> schools

src/maplibre_helpers.ts
```diff
@@ -112,7 +127,11 @@ export function overwritePolygonLayer(
       "fill-color": params.color,
       "fill-opacity": params.opacity,
     },
-  });
+  };
+  if (params.sourceLayer) {
```
dabreegster, line 131: Bit verbose, but the only way to reasonably play nice with the TS checks

src/maplibre_helpers.ts
```diff
@@ -265,6 +296,10 @@ const layerZorder = [
   "route-lines",
   "route-polygons",
 
+  // Display optional layers on top of interventions
```
dabreegster, line 299: I don't know if this is the right choice yet. If there's a school on top of a big area and you hover, currently you'll get both popups. Need to think more and get feedback about how this contextual data will be used in decision-making.

dabreegster (commented):
See https://github.com/acteng/atip-data-prep/pull/13 for how the 2 pmtiles files are created

src/lib/Legend.svelte
Pete-Y-CS: Ideally our layer controls would share as much across pages as possible so I think waiting until this has everything we want before finishing that makes complete sense

src/lib/browse/HospitalsLayer.svelte
Pete-Y-CS: Out of interest why is `singluarNoun` not capitalised but `pluralNoun` is?

dabreegster: One is used for "Unnamed X" and the other for the checkbox "[color] Hospitals". I'll add a comment

robinlovelace-ate:
Looking good from vid. Will give it a spin on the link.

Pete-Y-CS (approved):

robinlovelace-ate:
Looking good for Leeds. Unrelated comment is that the left legend isn't looking that relevant. Overall: amazing!

image

robinlovelace-ate (approved):

dabreegster:
Thanks for the reviews!

> Unrelated comment is that the left legend isn't looking that relevant.

Agreed. I think a next step is to let users control the styling of those interventions. The breakdown by 4 intervention types is likely not useful here, and coloring by funding program, construction status, etc might be nicer.

PR #289, opened 2023-07-24
Use tooltips to show the area name on the homepage

dabreegster:
Not sure why we didn't do this originally, but it's much easier than glancing over to the side to the see the name.

https://github.com/acteng/atip/assets/1664407/1f1fcc3e-25f1-443d-a567-2030ddea1229

### DIFF
--- a/src/pages/ChooseArea.svelte
+++ b/src/pages/ChooseArea.svelte
@@ -13,10 +13,11 @@
     SecondaryButton,
   } from "../lib/govuk";
   import "maplibre-gl/dist/maplibre-gl.css";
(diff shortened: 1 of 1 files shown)


src/pages/ChooseArea.svelte
```diff
@@ -58,6 +58,7 @@
     });
     // Use map within onMount, so it's guaranteed to exist
     loadedMap = map;
+    mapStore.set(loadedMap);
```
dabreegster, line 61: Low-pri later refactor to use existing Svelte components on this page too

dabreegster:
Async feedback welcome

PR #290, opened 2023-07-24
Some usability improvements for browse page

dabreegster:
First off, our free-form text filter is confusing before this PR. If you search for "school", you see things on the map that just happen to be in the same scheme that mentions schools. This PR changes it so that you _only_ see interventions mentioning that word. And to make that clear, highlight the search result in the popup.

Second, I flipped the optional layers to be disabled by default. We can revisit this as we have more of them and understand how they're used, but at the moment, I think they're a distraction when loading scheme data.

Third, the "edit on map" button is disabled when the authority name is incorrect, so people don't even click and later get an error. I'd very much like to fix the root cause here and rebuild the all-scheme dataset from its original input, but this is a stop-gap.

https://github.com/acteng/atip/assets/1664407/2d117def-87f4-49f2-8706-80646c9a1912

### DIFF
--- a/src/pages/App.svelte
+++ b/src/pages/App.svelte
@@ -5,7 +5,6 @@
   import * as Comlink from "comlink";
   import type { FeatureCollection, Polygon } from "geojson";
   import { onMount } from "svelte";
-  import authoritiesUrl from "../../assets/authorities.geojson?url";
   import BoundaryLayer from "../lib/BoundaryLayer.svelte";
   import {
     BaselayerSwitcher,
@@ -13,6 +12,7 @@
     Layout,
     ZoomOutMap,
   } from "../lib/common";
+  import { getAuthoritiesGeoJson } from "../lib/common/data_getter";
   import HoverLayer from "../lib/draw/HoverLayer.svelte";
   import InterventionLayer from "../lib/draw/InterventionLayer.svelte";
   import Toolbox from "../lib/draw/Toolbox.svelte";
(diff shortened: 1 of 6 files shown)


src/pages/App.svelte
```diff
@@ -79,9 +79,7 @@
   });
 
   async function loadAuthorityBoundary(): Promise<FeatureCollection<Polygon>> {
-    const resp = await fetch(authoritiesUrl);
-    const body = await resp.text();
-    const geojson: FeatureCollection<Polygon> = JSON.parse(body);
+    let geojson = await getAuthoritiesGeoJson();
```
dabreegster, line 82: While I was using that helper function, I realized it wasn't used here, but it should be

src/pages/BrowseSchemes.svelte
```diff
@@ -25,9 +26,15 @@
...
 
-  onMount(() => {
+  // TODO Remove after the input data is fixed to plumb correct authority names.
+  let authorityNames: Set<string> | null = null;
+
+  onMount(async () => {
     // For govuk components. Must happen here.
     initAll();
+
+    let geojson = await getAuthoritiesGeoJson();
```
dabreegster, line 36: This is an extra file for us to load on this page, slows down startup slightly. So, fixing the upstream problem in the combo data should be important.

src/pages/BrowseSchemes.svelte
```diff
@@ -58,16 +66,29 @@
...
     }
     return html;
   }
+
+  // When the user is filtering name/description by freeform text, highlight the matching pieces.
+  function highlightFilter(input: string): string {
+    if (!filterText) {
+      return input;
+    }
+    return input.replace(
```
dabreegster, line 87: Courtesy https://bitsofco.de/a-one-line-solution-to-highlighting-search-matches/#one-line-match-highlighting

dabreegster:
Adding Robin for async feedback on the search changes, but going to merge now and handle any feedback later. Thanks :)

PR #292, opened 2023-07-26
Adjust color/styling of different layers

dabreegster:
Notably, let the user choose to color interventions by ATF funding round. As we get more scheme-level data (like construction/completion status), this could become more useful. I've adjusted the basemap style and most colors using my best judgment and some of the brand guidelines, but this is still very much in need of design help.
Screenshot from 2023-07-26 12-06-28
Most of the PR is refactoring, I'll comment inline


src/lib/browse/MrnLayer.svelte
```diff
@@ -54,6 +55,13 @@
           Major Road Network
         </ExternalLink>.
       </p>
+      <p>
```
dabreegster, line 58: Going back to the stand-up this morning, some kind of standardized data catalog where we record license, owner, when we accessed the data, etc would be handy, and we can surface some of this info if it's useful

src/lib/browse/StyleInterventions.svelte
```diff
@@ -0,0 +1,52 @@
...
+        "grey"
+      );
+      legendRows = [
+        ["ATF2", colors.atf2],
+        ["ATF3", colors.atf3],
+        ["ATF4", colors.atf4],
+      ];
+    }
+
+    $map.setPaintProperty("interventions-points", "circle-color", color);
```
dabreegster, line 30: This is how to change a layer's styling rules dynamically

src/lib/browse/data.ts
```diff
@@ -35,17 +34,12 @@ export function processInput(
   }
 
   for (let feature of gj.features) {
-    schemes.get(feature.properties!.scheme_reference).num_features++;
```
dabreegster, line 38: I cleaned up these problems in the latest version of the scheme file I've sent around

src/lib/common/BaselayerSwitcher.svelte
```diff
@@ -21,6 +21,7 @@
   choices={[
     ["streets", "Streets"],
     ["hybrid", "Satellite"],
+    ["dataviz", "Dataviz"],
```
dabreegster, line 24: MapTiler has a bunch of basemap styles. This one looks the most neutral to me, showing enough to help people orient, but getting out of the way to let the data on top show clearly

src/lib/common/Legend.svelte
dabreegster, line 1: Just refactored this multi-row color legend pattern

src/lib/draw/route/route_tool.ts
```diff
@@ -52,24 +53,19 @@ export class RouteTool {
       id: "route-points",
       source,
       filter: isPoint,
-      color: [
-        "match",
+      color: matchValue(
```
dabreegster, line 56: There were a few places where we were using maplibre case or match manually. All using this one helper method now.

src/schemas.ts
dabreegster, line 1: There were too many functions added to `types.ts`. Ideally these could be methods on the Schema type, but since it's just a wrapper around a string, haven't sorted out exactly how to do that yet

dabreegster (commented):
Async review welcome

PR #294, opened 2023-07-26
Generate random data for the scheme browse page

dabreegster:
Just a quick script to churn out vaguely sensible looking example data. We can use this to explore GCP deployment in the short-term, performance-test the browse page, take videos for GH pull requests, etc.

https://github.com/acteng/atip/assets/1664407/b4fb9751-eb23-40d1-b2bd-1bd33d788be4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@turf/bbox": "^6.5.0",
+        "@turf/boolean-contains": "^6.5.0",
         "@turf/boolean-point-in-polygon": "^6.5.0",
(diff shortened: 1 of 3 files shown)


Pete-Y-CS (commented):
lgtm! Is this something that should be in data prep though?

dabreegster:
Good question, I was really not sure where it fit. Think we should move it there? Will need to copy over a small amount of TS and node js config

Pete-Y-CS:
I mean it feels like ideally it belongs there but I don't mind it being here for expedience!

An aside: I just checked that expedience meant what I thought it did... It more or less does but I'm not sure I mean to imply this morally unacceptable 😅 
image

PR #295, opened 2023-07-27
Coloring changes

Pete-Y-CS:
Not fully finished reviewing the existing changes but here are some cosmetic changes that I _think_ would make it less confusing to me/other newcomers

### DIFF
--- a/src/lib/browse/InterventionColorSelector.svelte
+++ b/src/lib/browse/InterventionColorSelector.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { matchValue } from "../../maplibre_helpers";
+  import { get   return x as DataDrivenPropertyValueSpecification } from "../../maplibre_helpers";
   import { colorInterventionsBySchema, schemaLegend } from "../../schemas";
   import { map } from "../../stores";
   import { Legend } from "../common";
@@ -15,7 +15,7 @@
       color = colorInterventionsBySchema("v1");
       legendRows = schemaLegend("v1");
     } else {
-      color = matchValue(
+      color = get   return x as DataDrivenPropertyValueSpecification(
         ["get", "funding_programme"],
         { ATF2: colors.atf2, ATF3: colors.atf3, ATF4: colors.atf4 },
         "grey"
--- a/src/lib/browse/InterventionColorSelector.svelte
+++ b/src/lib/browse/InterventionColorSelector.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { get   return x as DataDrivenPropertyValueSpecification } from "../../maplibre_helpers";
+  import { constructMatchExpression } from "../../maplibre_helpers";
   import { colorInterventionsBySchema, schemaLegend } from "../../schemas";
   import { map } from "../../stores";
   import { Legend } from "../common";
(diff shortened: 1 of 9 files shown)


src/maplibre_helpers.ts
```diff
@@ -229,18 +229,26 @@ export function setPrecision(pt: Position): Position {
 //  Helper for https://maplibre.org/maplibre-style-spec/expressions/#match.
 //  Gets one feature property, uses a map to match a key to a value, and
 //  includes a fallback if no keys match.
-export function matchValue<Output>(
+export function getDataDrivenPropertyValueSpecification<OutputType>(
```
dabreegster, line 232: Hmm, not sure this name makes sense to me. DataDrivenPropertyValueSpecification is described (not with this name) at https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#data-expressions. It's a general way of describing things in maplibre styles that get calculated per feature. This method is much more specific -- it's just for `match`, linked above. It's basically a switch-case statement:

1) Fetch some property of the feature (getter)
2) Go through each of the key/value pairs in the mapping and try to match
3) If that fails, use a fallback value

If `matchValue` is too terse, how about something like `pickValueBasedOnKey` or something like that?

src/maplibre_helpers.ts
```diff
@@ -229,18 +229,26 @@ export function setPrecision(pt: Position): Position {
...
   getter: any[],
-  map: { [name: string]: Output },
-  fallback: Output
-): DataDrivenPropertyValueSpecification<Output> {
+  map: { [name: string]: OutputType },
+  fallback: OutputType
+): DataDrivenPropertyValueSpecification<OutputType> {
+  // This looks weird because it is weird. I hate to add a comment but here we are.
```
dabreegster, line 237: This is very verbose, do we want to check it in? Are the linked docs at https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#match confusing? The interface of the function is hopefully clear, or we can change param names to make it better. The implementation is hopefully clear with a glance at the docs

src/lib/browse/InterventionColorSelector.svelte
```diff
@@ -1,21 +1,21 @@
 <script lang="ts">
-  import { matchValue } from "../../maplibre_helpers";
+  import { get   return x as DataDrivenPropertyValueSpecification } from "../../maplibre_helpers";
```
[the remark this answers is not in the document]
Pete-Y-CS: Hmm, I don't think I opened this file, so I think somehow VSCode did this when I tried to reformat somehow?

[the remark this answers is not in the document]
Pete-Y-CS: No hang on this is one I did work on but I don't think I manually changed the imports, how werid

src/maplibre_helpers.ts
Pete-Y-CS: The thing is, it's not quite `pickValueBasedOnKey`, because you're getting an object from which you can pick a value right?

So my main issue with the name is that you're returning a data structure rather than the end matched result, to my understanding. I agree this name isn't great either though

src/maplibre_helpers.ts
Pete-Y-CS: I am going to blame it on my addled illness brain but I didn't see those docs. Those docs seem fine

src/maplibre_helpers.ts
Pete-Y-CS: Something like `getPropertyMatcher` or `constructMatchExpression`?

dabreegster: > you're returning a data structure rather than the end matched result

That's right, and it's how style expressions work in general. We're describing to maplibre how to choose a color, opacity, line width, etc.

Both those suggestions sound good to me, `constructMatchExpression` would get my vote

dabreegster (approved):

PR #297, opened 2023-07-28
Rename the layers to layerToggles

Pete-Y-CS:


dabreegster:
Hmm... "Hospital Layer Toggle" makes it sound like the component is just a toggle control, but it defines the layer itself. The checkbox and toggling visibility behavior is just part of it -- setting up the maplibre source and layer styles and the help button are just as important.

I'd agree `OptionalLayer` is a poor name at this point. It's more accurately `PolygonAmenityLayer`, a convenience for schools and hospitals and other similar layers to come. MRN and the PC boundary one need different things, so they don't use it.

Pete-Y-CS:
Well the components are the toggles, right? The fact that they have the logic to display the layer is secondary when you're using them in a page? To me it almost suggests that these should be called LayerToggle still, but call out to some ts  which actually draws the layer?

Pete-Y-CS:
I guess the thinking behind this: the name is for a Svelte component so it should describe its function in a page and what it actually renders.

dabreegster:
> the name is for a Svelte component so it should describe its function in a page and what it actually renders.

This makes sense to me. The function on the page includes the small colored box as legend, the toggle, and the help/more info button. So what if we name these "xyz LayerControl.svelte"? And OptionalLayer -> PolygonAmenityLayerControl?

Pete-Y-CS:
Sounds perfect

Issue #285, opened 2023-07-24
Admin boundary layers for the browse page


dabreegster:
- [ ] MCAs
- [ ] Local/ Highway/ Planning Authority boundaries
- [ ] ATE regional engagement)
- [ ] MP’s wards
- [ ] District Ward Boundaries

As much as possible, let's try to script the download + transform of these in the data-prep repo. We can explore the topojson format for minimizing file size + load time; https://github.com/mbloch/mapshaper is the tool that can convert GJ to topojson.

(This issue is related to #264, but more focused just on displaying layers in the browse page. We're not trying to change `authorities.geojson` in this repo, which is used for defining areas where people sketch schemes.)

Pete-Y-CS:
I was also going to look into LSOAs for what it's worth

dabreegster:
District wards seem visible at https://www.ordnancesurvey.co.uk/election-maps/gb/

dabreegster:
Inspiration for desired output: https://maproom.net/demo/election-map/0.html

dabreegster:
MP wards (aka parliamentary constituencies) are half-done, working on the rest. Others are up for grabs

Pete-Y-CS:
https://geoportal.statistics.gov.uk/datasets/ons::local-authority-districts-december-2022-boundaries-uk-buc-2/explore?location=54.141544%2C-1.532166%2C6.92

https://geoportal.statistics.gov.uk/datasets/ons::combined-authorities-december-2022-boundaries-en-buc/explore

Here are what I'm using for LAs and MCAs (metro combined authorities?)  Just putting here so it's logged somewhere

robinlovelace-ate:
Let's go with the May 2023 data as a starting point, 5 months more recent.

robinlovelace-ate:
Follow-up on this, here's a list of boundary layers from SchemeMapper:

<html>
<body>


Local Authority Districts (Lower Tier Local Authorities) |  
-- | --
  |   | Counties and Unitary Authorities (Upper Tier Local Authorities) |  
  |   | Constituencies & MP's |  
  |   | Government Regions |  
  |   | Sub-National Transport Bodies |  
  |   | Local Planning Authorities |  
  |   | Local Transport Authorities |  
  |   | England National Boundary |  
  |   | Combined Authorities |  
  |   | Local Enterprise Partnerships (Overlapping) |  
  |   | Local Enterprise Partnerships (Non-Overlapping) |  
  |   | ONS Built-up Areas |  
[...]

PR #298, opened 2023-07-31
Add layer with wards, LADs, and CAs. #285

dabreegster:
Thanks to Alex for pointing us to the data source! The import is in https://github.com/acteng/atip-data-prep/pull/17.

Demo: https://acteng.github.io/atip/wards/browse.html


src/lib/browse/ParliamentaryConstituenciesLayerControl.svelte
```diff
@@ -62,6 +62,7 @@
     let name = e.detail.properties.Name;
     name = name.replace(/ Boro Const$/, "");
     name = name.replace(/ Co Const$/, "");
+    name = encodeURIComponent(name);
```
dabreegster, line 65: Need to be paranoid about "&" and things

src/lib/browse/WardsLayerControl.svelte
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  function tooltip(feature: MapGeoJSONFeature): string {
+    return `<p>${feature.properties.name}</p>`;
+  }
+
+  function onClick(e: CustomEvent<MapGeoJSONFeature>) {
+    let name = encodeURIComponent(e.detail.properties.name);
+    // Help people find the councillor for this area
```
dabreegster, line 62: I could not do any better than Google this time :(

src/lib/browse/colors.ts
```diff
@@ -4,6 +4,7 @@ export const colors = {
   hospitals: "#B73D25",
   mrn: "#006478",
   parliamentary_constituencies: "#006E59",
+  wards: "purple",
```
dabreegster, line 7: Back to just making these up. We need a principled way of doing this.

dabreegster (commented):
UX problems when there are multiple boundaries turned on, but we need to know how people want to use this before figuring out a fix. (Like only allowing one boundary layer at a time)

robinlovelace-ate:
My ward is in there and up-to-date. 👍 on that basis, like the link to Google for further information, also not sure if we can do better than that... would hope so!

image

robinlovelace-ate:
Broader thought probably already discussed but just while in my head: easy to add these layers to the scheme editing and other pages at some point in the future?

src/lib/browse/CombinedAuthoritiesLayerControl.svelte
Pete-Y-CS, line 1: As I think we've discussed, a lot of these controls share stuff in common, maybe we should log an issue to tidy them up once we've created all the layers that exist and we can think clearly about what can be shared?

src/lib/browse/colors.ts
Pete-Y-CS: I think we're also gonna run out of significantly distinct colours.

Pete-Y-CS (approved):

src/lib/browse/colors.ts
dabreegster: We could maybe make all the boundary layers one color, and effectively change this to a radio button. But I'm not sure -- we need to know how people want to use this.

src/lib/browse/CombinedAuthoritiesLayerControl.svelte
dabreegster: Yep, 100% agree. I think we need a few more first to understand commonalities. The census ones are turning out to be special in a few ways...

src/lib/browse/colors.ts
Pete-Y-CS: Yeah, I also imagined we could have a set of colours which got applied to things as you ticked em, and you are only allowed a certain number of layers at once. A changing legend does sound like bad UX though so probably not a good idea

Issue #286, opened 2023-07-24
Sociodemographic layers for the browse page


dabreegster:
- [x] Car ownership (I've worked with this a bit in https://github.com/dabreegster/popgetter before)
- [x] IMD
- [x] Population density
- [ ] Employment density

https://nptscot.github.io has excellent references for how to visualize these. They have the option to visualize as a choropleth or dasymetric.

We can think through how to implement these. If multiple layers share the same geography, we don't need a copy of everything each time. We can load the geometry, then match up to each dataset by ID.

robinlovelace-ate:
Thumbs up on this.

robinlovelace-ate:
Key missing data on there:

- [ ] Mode split (2011 and possibly 2021 Census data although latter will be heavily distorted by distancing measures)

robinlovelace-ate:
Also: open access JTS / Connectivity data?

PR #301, opened 2023-08-02
Add 3 census layers at output area granularity

dabreegster:
Thanks to @cmconlan for finding and prepping the data. #286

This adds percent of households without a car, average cars per household, and population density, all at OA level from the 2021 census. Demo: https://acteng.github.io/atip/census_layers/browse.html

https://github.com/acteng/atip/assets/1664407/9e72aef2-9a26-40ee-b419-e2cfa33f8873

### Design issues

There are lots of these, but I want to time-box focusing on this and move onto more requests, and also get feedback from users on what we've got so far. If there are any short-term improvements, happy to do!

Before even getting to the question of how to show these layers in conjunction with any loaded scheme data (which is a problem with all of the layers right now)...

- We're using 3 checkboxes, but really they act like radio buttons, because you can only view one layer at a time. Alternatively we could have a checkbox for "census data at OA level" and then expose a dropdown or radio, but this makes discovering the layers harder.
- The checkboxes are enormous and not part of the regular group, because we need to dynamically add the color legend beneath them. The govuk checkbox group component breaks when I tried that.
- The scale for percent of households without a car and number of cars per household is flipped, so when you swap between the layers, the colors kind of invert. Potentially confusing?
- The [ONS map](https://www.ons.gov.uk/census/maps/choropleth/housing/number-of-cars-or-vans/number-of-cars-5a/no-cars-or-vans-in-household?oa=E00079677) used as reference changes the buckets/scale as you zoom around. We don't do that; the buckets are fixed.
- Things are choppy to load / move around, and the geometry is very simplified, at low zoom levels. It'd be better if we mimicked the ONS map and actually switched to using different data at lower granularity (like MSOA) when we're at low zoom.

It could be nice to have an optional for dasymetric maps instead of the choropleth, like nptscot.github.io. Consider it possible future work.


src/lib/browse/CensusOutputAreaLayerControl.svelte
```diff
@@ -0,0 +1,244 @@
...
+    } else {
+      window.open(
+        `https://www.ons.gov.uk/census/maps/choropleth/population/population-density/population-density/persons-per-square-kilometre?oa=${oa}`,
+        "_blank"
+      );
+    }
+  }
+
+  // Should return 6 things (1 more than colorScale)
+  function makeLimits(): number[] {
```
dabreegster, line 96: Chris, I still need to fix my Python dependencies to run your script and see what the cut-offs for the bandings you used. I tried out different quantile functions in pandas, but they don't seem to handle the min / max value. For MapLibre, we have to set a real max value, or it uses a fallback color. So I've made some simple first choices here, but happy to iterate on them

src/lib/browse/CensusOutputAreaLayerControl.svelte
```diff
@@ -0,0 +1,244 @@
...
+      return [0.0, 0.5, 1.0, 1.5, 2.0, 2.5];
+    } else {
+      // Use the same (slightly rounded) buckets as https://www.ons.gov.uk/census/maps/choropleth/population/population-density/population-density/persons-per-square-kilometre
+      return [0, 4700, 13000, 33000, 94000, 1980000];
+    }
+  }
+
+  function makeColorRamp(): any[] {
+    let limits = makeLimits();
+    let fillColor: any[] = ["step", ["get", colorBy]];
```
dabreegster, line 111: See https://maplibre.org/maplibre-style-spec/expressions/#step for reference... and note the docs don't match reality or the example code, because `stop_output_0` is a lie. Instead there's a fallback value at the very end. I'll file an issue / send a PR to fix the docs...

src/lib/browse/CensusOutputAreaLayerControl.svelte
```diff
@@ -0,0 +1,244 @@
...
+    // is used. It's effectively a bug -- the limits need to be set correctly.
+    fillColor.push("red");
+    return fillColor;
+  }
+</script>
+
+<Checkbox
+  id="percent_households_without_car"
+  bind:checked={showHouseholdsWithoutCar}
+  on:change={() => {
```
dabreegster, line 126: Manual radio button logic :)

src/lib/browse/colors.ts
```diff
@@ -8,6 +8,15 @@ export const colors = {
   combined_authorities: "cyan",
   local_authority_districts: "orange",
 
+  // Color ramp from https://www.ons.gov.uk/census/maps/choropleth
```
dabreegster, line 11: Other inspiration at https://alan-turing-institute.github.io/demoland-web, http://tristen.ca/hcl-picker/, https://nptscot.github.io, etc

src/pages/BrowseSchemes.svelte
```diff
@@ -149,6 +150,7 @@
             <CombinedAuthoritiesLayerControl />
             <LocalAuthorityDistrictsLayerControl />
           </CheckboxGroup>
+          <CensusOutputAreaLayerControl />
```
dabreegster, line 153: If I use the regular checkbox group above, the legends on different lines blow up:
Screenshot from 2023-08-02 13-04-48
Help welcome.

Pete-Y-CS:
> - We're using 3 checkboxes, but really they act like radio buttons, because you can only view one layer at a time. Alternatively we could have a checkbox for "census data at OA level" and then expose a dropdown or radio, but this makes discovering the layers harder.

I think it's not terrible behaviour to have these switch off other latyers, I agree let the users see and focus on feedback they give.

> - The checkboxes are enormous and not part of the regular group, because we need to dynamically add the color legend beneath them. The govuk checkbox group component breaks when I tried that.

I could look into trying to do some bespoke styling to make it a little less misbehaving if you like?

> - The scale for percent of households without a car and number of cars per household is flipped, so when you swap between the layers, the colors kind of invert. Potentially confusing?

Hmm I think we can probably just invert households without a car? make it households with 1 or more car and then it's okay?

>- The [ONS map](https://www.ons.gov.uk/census/maps/choropleth/housing/number-of-cars-or-vans/number-of-cars-5a/no-cars-or-vans-in-household?oa=E00079677) used as reference changes the buckets/scale as you zoom around. We don't do that; the buckets are fixed.

Yeah I was wondering if we wanted to integrate LSOAs as well as OAs. Think another question for users when they have a go

> - Things are choppy to load / move around, and the geometry is very simplified, at low zoom levels. It'd be better if we mimicked the ONS map and actually switched to using different data at lower granularity (like MSOA) when we're at low zoom.
That would be better but again, as you said, no need to prematurely optimise before users tell us what they want, right?

dabreegster:
> I could look into trying to do some bespoke styling to make it a little less misbehaving if you like?

If you have a spare moment, sure, but I think we could hold off. The flat list of layers is already too big; we won't be keeping this one CheckboxGroup for long

> Hmm I think we can probably just invert households without a car? make it households with 1 or more car and then it's okay?

You're right, this will be more straightforward. It also matches the bandings Chris created originally! I'll make this change later today

> Yeah I was wondering if we wanted to integrate LSOAs as well as OAs

The IMD and employment density data is at LSOA level, not OA, so we will have another layer for those soon. I think we can hold off on improving performance at low zooms for these layers right now, until we get feedback

Pete-Y-CS (approved):

Robinlovelace:
Great to see this and big :+1: from me, looking really strong.

cmconlan:
To add my comments on a few things:

- Yes, agree we could change "Percentage without a car" to "percentage with at least 1 car" or something to this effect.
- It would be a nice feature to switch out the granularity of the data depending on the zoom, but I expect most of the analysis performed in the tool will be very localised, so keeping it at the most granular level I think is fine. As has been mentioned we can see what feedback is from users in any case
- Longer term i think it's a nice feature to change the scale so it's relative to whatever is in view, but I don't think it significantly detracts from the utility of the analysis so we not urgent I don't think
- Think it's ok to have radio buttons, but UI may get confusing as we add more and more to it. Groupings would be nice.

Overall think it looks great - really happy with

No comments on code.

src/lib/browse/CensusOutputAreaLayerControl.svelte
yongrenjie: I've not gotten the opportunity to use this Svelte feature yet, but I wonder if https://svelte.dev/tutorial/sharing-code might be useful here?! Might just be a complicated answer to a simple problem 🙃

dabreegster: Ooh, thanks, that is clever! There could be a component wrapping these checkboxes. I'm not sure what the order would be though... does `on:change` happen after the checkbox state is toggled on? And in that shared `stopOthers` method, can we detect the "current" checkbox and keep it on? Maybe...

PR #304, opened 2023-08-04
Add LPA boundaries. #285

dabreegster:
https://github.com/acteng/atip/assets/1664407/d4727afa-e745-46ba-b23c-780819cbb840

Just something simple to start with

### DIFF
--- a/src/lib/browse/LocalPlanningAuthoritiesLayerControl.svelte
+++ b/src/lib/browse/LocalPlanningAuthoritiesLayerControl.svelte
@@ -1,0 +1,112 @@
+<script lang="ts">
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import {
+    hoveredToggle,
+    overwriteLineLayer,
+    overwritePmtilesSource,
+    overwritePolygonLayer,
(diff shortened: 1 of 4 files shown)


src/lib/browse/LocalPlanningAuthoritiesLayerControl.svelte
```diff
@@ -0,0 +1,112 @@
...
+    if ($map.getLayer(outlineLayer)) {
+      $map.setLayoutProperty(
+        outlineLayer,
+        "visibility",
+        show ? "visible" : "none"
+      );
+    }
+  }
+
+  // TODO Note there are overlapping features. InteractiveLayer only plumbs
```
dabreegster, line 56: We could improve this, at the moderate cost of making all the other users of this layer more complex. I'd almost want to use a distinct component with the more complicated API. For now, we don't even know if this is the correct layer to show, or what we want to do with it, so let's not pay that cost just yet

dabreegster (commented):
No click handler, because I couldn't find anything useful to show when googling the LPA codes or names

PR #305, opened 2023-08-07
More trip generators

Pete-Y-CS:


src/lib/browse/PointAmenityLayerControl.svelte
```diff
@@ -0,0 +1,61 @@
...
+      source: name,
+      filter: ["all", isPoint] as FilterSpecification,
+      color: color,
+      radius: circleRadius / 2,
+      // TODO Outline?
+    });
+    show = false;
+  };
+
+  overwriteGeojsonSource($map, name, `https://atip.uk/layers/v1/${name}.geojson`, layerLoadedCallback);
```
Pete-Y-CS, line 43: I am doing it with a callback because I think it needed to be async to handle getting the data,  so this behaves a bit differently to the polygon,  pmtile based layers

dabreegster:
https://acteng.github.io/atip/more-trip-generators/browse.html for the demo

Sports spaces look fine to me!

# Railway stations

- The tooltip for name isn't working
- The .geojson file adds 1MB to the page load, bringing it from 3 to 4MB total. (Ctrl+Shift+I, Network tab in Firefox shows a useful running total). The pmtiles files are nice here, since they only load a few KB at a time. We don't need to switch formats here, but we do need to clean up the geojson file by only keeping properties we care about (name) and get this file to be smaller

# Green spaces

I think we shouldn't add these in yet; they'll take more work to clean up and validate. First off, there are overlapping areas that wind up looking very weird:

https://github.com/acteng/atip/assets/1664407/b508f77d-c1e7-4d5a-a388-784f60bf22b5

We're picking up enormous conservation zones in the middle of the ocean:
Screenshot from 2023-08-07 21-12-01

We're missing valid green spaces like [Blakeley Forest](https://www.openstreetmap.org/way/27003796) that even have footpaths through them. (See https://acteng.github.io/atip/more-trip-generators/browse.html#14.69/53.52992/-2.23673)

Based on Anna's (colleague at Turing) experiences trying to generate similar data from OSM and OS for another project, I think it'll take a while to refine this layer. In the meantime, the basemap already shows green space in a minimal way, so it's probably not too urgent.

src/lib/browse/GreenSpacesLayerControl.svelte
```diff
@@ -0,0 +1,29 @@
...
+</script>
+
+<PolygonAmenityLayerControl
+  name="green_spaces"
+  singularNoun="green space"
+  pluralNoun="Green spaces"
+>
+  <p>
+    This shows <ExternalLink
+      href="https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dhospitalhttps://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpark"
```
dabreegster, line 13: Copy paste error, I still see hospital?

src/lib/browse/PointAmenityLayerControl.svelte
```diff
@@ -0,0 +1,61 @@
...
+  // The caller must also fill in the default slot with the contents of a help modal
+
+  // @ts-ignore TODO Also constrain name to exist in the colors type
+  let color = colors[name];
+
+  const layerLoadedCallback = () => {
+    overwriteCircleLayer($map, {
+      id: name,
+      source: name,
+      filter: ["all", isPoint] as FilterSpecification,
```
dabreegster, line 35: For our one use case so far of railway stations, this should be unnecessary. We should make sure the input data only contains points in the first place, or understand what stations aren't available as single points and deal with them somehow. Otherwise this is just hiding a problem of missing data

src/maplibre_helpers.ts
```diff
@@ -47,6 +47,25 @@ export function overwritePmtilesSource(map: Map, id: string, url: string) {
   });
 }
 
+// Like overwriteSource, but for geojson data hosted at a URL.
```
dabreegster, line 50: See line 32. `overwriteSource` takes GJ data or a URL to it, because the underlying maplibre API handles it. We don't need this

src/lib/browse/PointAmenityLayerControl.svelte
dabreegster: We don't need a callback; the MapLibre API handles all of this. We just have to add a source before adding the layer that references it. Under the hood, it fetches the data and feeds it to the layer rendering. See https://maplibre.org/maplibre-gl-js/docs/examples/geojson-polygon/ for a simple example

dabreegster:
Ah, the missing tooltips with railway stations are because the features don't have numeric IDs assigned in the geojson. Maplibre is frustratingly silent about this when they're missing :\

src/lib/browse/PointAmenityLayerControl.svelte
Pete-Y-CS: I wrongly thought you needed a filter and it wouldn't let me have an empty list. I really think if you have something that is represented as a list of some kind that you should represent its absence as an empty list. But that's by the by, removing

dabreegster: That's an unfortunate API choice in MapLibre styles generally. In methods where we take optional parameters, we use [defaults](https://github.com/acteng/atip/blob/3c844aeaaeb2acd5a1f9f91e050fe3bef4edb32c/src/maplibre_helpers.ts#L18) like this, because that's seemingly the only choice

src/lib/browse/PointAmenityLayerControl.svelte
Pete-Y-CS: Not sure why what I tried before didn't work but it's alright now

src/lib/browse/colors.ts
```diff
@@ -3,6 +3,9 @@ export const colors = {
   schools: "#007DBB",
   hospitals: "#B73D25",
   mrn: "#006478",
+  green_spaces: "green",
+  sports_spaces: "#013220",
+  railway_stations: "#7BD352",
```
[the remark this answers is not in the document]
dabreegster: We can remove green spaces for now

The light green for railway stations is a bit hard to see to me -- thoughts on `#A52A2A`?

src/maplibre_helpers.ts
```diff
@@ -292,7 +292,10 @@ const layerZorder = [
   // Then smaller optional layers on top
   "schools",
   "hospitals",
+  "green_spaces",
```
dabreegster, line 295:
```suggestion
```

src/lib/browse/PointAmenityLayerControl.svelte
```diff
@@ -0,0 +1,55 @@
...
+  } from "../../maplibre_helpers";
+  import { map } from "../../stores";
+  import { ColorLegend, HelpButton, InteractiveLayer } from "../common";
+  import { Checkbox } from "../govuk";
+  import { colors } from "./colors";
+
+  // This name is used for multiple things:
+  // - The name of a .pmtiles file
+  // - The layer name within that file
+  // - The layer name, for layerZorder
```
dabreegster, line 15:
```suggestion
  // - The name of a .geojson file
  // - The layer name, for layerZorder
```

dabreegster (approved):
Just a few cleanups, otherwise LGTM. Nice work!

PR #306, opened 2023-08-10
Add a layer showing bus routes with and without bus lanes

dabreegster:
Demo at https://acteng.github.io/atip/bus_routes/browse.html

Two colors and a tooltip to indicate if each road involving some bus route has a dedicated lane or not. Not attempting yet to show _direction_ of the bus lane. We could consider arrows or lines clearly on the left or right side of a road to do this in the future, or even showing a full-blown schematic like streetmix.net or osm2streets.org on hover.

The line width is kind of thick and awkward at low zooms. Existing problem with MRN too; might followup with a separate PR and try out changing width based on zoom level.

### DIFF
--- a/src/lib/browse/BusRoutesLayerControl.svelte
+++ b/src/lib/browse/BusRoutesLayerControl.svelte
@@ -1,0 +1,77 @@
+<script lang="ts">
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import {
+    hoveredToggle,
+    overwriteLineLayer,
+    overwritePmtilesSource,
+  } from "../../maplibre_helpers";
+  import { map } from "../../stores";
+  import {
+    ColorLegend,
+    ExternalLink,
+    HelpButton,
+    InteractiveLayer,
+  } from "../common";
+  import { Checkbox } from "../govuk";
+  import { colors } from "./colors";
+
+  let name = "bus_routes";
+
+  overwritePmtilesSource(
+    $map,
+    name,
+    `https://atip.uk/layers/v1/${name}.pmtiles`
+  );
+
+  overwriteLineLayer($map, {
+    id: name,
+    source: name,
+    sourceLayer: name,
+    color: [
+      "case",
+      ["boolean", ["get", "has_bus_lane"], false],
+      colors.bus_route_with_lane,
+      colors.bus_route_without_lane,
+    ],
+    width: 5,
+    opacity: hoveredToggle(0.5, 1.0),
+  });
+
+  let show = false;
+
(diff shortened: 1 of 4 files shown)


dabreegster:
https://github.com/acteng/atip/assets/1664407/bdf3e6e1-25cb-4dcf-8ea7-8b56bccae875

src/lib/browse/BusRoutesLayerControl.svelte
```diff
@@ -0,0 +1,77 @@
...
+    ],
+    width: 5,
+    opacity: hoveredToggle(0.5, 1.0),
+  });
+
+  let show = false;
+
+  function tooltip(feature: MapGeoJSONFeature): string {
+    if (feature.properties.has_bus_lane) {
+      return "<p>At least one bus route crosses here, with a bus lane in one or more directions</p>";
```
dabreegster, line 44: Is the tooltip too verbose and annoying?

src/lib/browse/colors.ts
```diff
@@ -10,6 +10,8 @@ export const colors = {
   combined_authorities: "cyan",
   local_authority_districts: "orange",
   local_planning_authorities: "red",
+  bus_route_with_lane: "#9362BA",
```
dabreegster, line 13: Just kind of picked a color and then found a darker shade of it. Suggestions very welcome

src/pages/BrowseSchemes.svelte
```diff
@@ -164,6 +165,7 @@
           <CollapsibleCard label="Road network">
```
dabreegster, line 165: Not sure if this category name is still the best, or how we'll group things as we pull in more

src/lib/browse/BusRoutesLayerControl.svelte
Pete-Y-CS: I don't think so, in any case one to await feedback: it doesn't really matter whether we think it's verbose or annoying!

src/lib/browse/colors.ts
Pete-Y-CS: I like it

src/pages/BrowseSchemes.svelte
Pete-Y-CS: Agreed it seemed a little odd to me but nowhere better for now. I guess in future we might want a public transport card

Pete-Y-CS: Train stations could be both in public transport and trip generators

Pete-Y-CS (approved):

PR #307, opened 2023-08-10
Explicitly list the date relevant for each source dataset

dabreegster:


dabreegster:
Needless to say, I regenerated all the OSM layers just now, not just making up a date :)

Pete-Y-CS (approved):

dabreegster:
The test failure is definitely a flake; it's on an unrelated page. Thanks for the review, merging

PR #308, opened 2023-08-11
Cycle parking layer

dabreegster:
Demo: https://acteng.github.io/atip/parking/browse.html

References:
- https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_parking
- https://bikedata.cyclestreets.net/tflcid/conversion/#cycle_parking
- https://bikedata.cyclestreets.net/cycleparking/#16.88/51.511361/-0.102809

I've chosen to only show capacity, not details like type, covered, public/private. We can extend this later with feedback.

Design-wise, only a few points are shown at low zoom, and at high zoom, the dots are not terribly significant. I think symbolic icons and https://maplibre.org/maplibre-gl-js/docs/examples/cluster/ will help with both issues, and also potentially be useful for other layers. We can explore them later.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+src/style/main.*
--- a/src/pages/BrowseSchemes.svelte
+++ b/src/pages/BrowseSchemes.svelte
@@ -7,6 +7,7 @@
   import BusRoutesLayerControl from "../lib/browse/BusRoutesLayerControl.svelte";
   import CensusOutputAreaLayerControl from "../lib/browse/CensusOutputAreaLayerControl.svelte";
   import CombinedAuthoritiesLayerControl from "../lib/browse/CombinedAuthoritiesLayerControl.svelte";
+  import CycleParkingLayerControl from "../lib/browse/CycleParkingLayerControl.svelte";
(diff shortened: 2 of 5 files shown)


.prettierignore
dabreegster, line 1: This makes `npm run fmt` run way faster; that auto-generated govuk file is big!

src/pages/BrowseSchemes.svelte
```diff
@@ -162,10 +163,11 @@
               <RailwayStationsLayerControl />
             </CheckboxGroup>
           </CollapsibleCard>
-          <CollapsibleCard label="Road network">
+          <CollapsibleCard label="Existing infrastructure">
```
dabreegster, line 166: Renamed this for the moment, I still expect churn as we add more and recategorize

Robinlovelace:
Big :+1: to cycle parking, a cost-effective low hanging fruit in many places.

dabreegster:
Gentle ping on this -- think it's OK for a first version? Icons and clustering will really help this and other layers, maybe we can get to that this week.

Pete-Y-CS (approved):

PR #309, opened 2023-08-14
Add an IMD layer

dabreegster:
https://github.com/acteng/atip/assets/1664407/10bac2b5-c674-43c1-8697-f2909175867a

https://acteng.github.io/atip/imd/browse.html

Hopefully a pretty straightforward one. I sanity checked compared ranks to http://dclgapps.communities.gov.uk/imd/iod_index.html.

### DIFF
--- a/src/lib/browse/ImdLayerControl.svelte
+++ b/src/lib/browse/ImdLayerControl.svelte
@@ -1,0 +1,96 @@
+<script lang="ts">
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import {
+    hoveredToggle,
+    makeColorRamp,
+    overwriteLineLayer,
+    overwritePmtilesSource,
+    overwritePolygonLayer,
+  } from "../../maplibre_helpers";
+  import { map } from "../../stores";
+  import { ExternalLink, HelpButton, InteractiveLayer } from "../common";
+  import { Checkbox } from "../govuk";
+  import { colors } from "./colors";
+  import SequentialLegend from "./SequentialLegend.svelte";
+
+  let name = "imd";
+  let outlineLayer = `${name}-outline`;
+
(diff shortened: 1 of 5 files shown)


src/lib/browse/ImdLayerControl.svelte
dabreegster, line 1: Couldn't find any useful reference to open for clicking; http://dclgapps.communities.gov.uk/imd/iod_index.html doesn't have URL controls

src/pages/BrowseSchemes.svelte
```diff
@@ -179,6 +180,7 @@
           </CollapsibleCard>
           <CollapsibleCard label="Census">
             <CensusOutputAreaLayerControl />
+            <ImdLayerControl />
```
dabreegster, line 183: It's possible to one of three OA-level census layers alongside the IMD layer. This is awkward, but I don't know how to solve it (UX-wise) yet. I think we only ever want one boundary/census/"covers everything" layer on at a time, and not sure how to communicate that as radio buttons across categories

src/lib/browse/ImdLayerControl.svelte
Pete-Y-CS: I wonder if it might be valuable to still send the user through to that page, even if we can't send them to the specific LSOA info?

dabreegster: It's listed in the help popup. Feels unintuitive to make a click on a specific LSOA in the map have the same effect everywhere

src/pages/BrowseSchemes.svelte
Pete-Y-CS: I agree. I don't know how to communicate it, and it's worth checking with users first, but making these and boundaries all mutually exclusive sounds like a good future task

Pete-Y-CS (approved):

PR #311, opened 2023-08-15
Organize browse layer code

dabreegster:
We have enough files to organize a bit now.

I'd still like to reduce verbosity / visual noise by removing the "LayerControl" suffix from the filenames, making LayerControls.svelte easier to skim, but that's optional


Pete-Y-CS:
I  was thinking this needed doing so this is very nice! I will say that I think removing layer control  from filename is fine if we renamed the directory to layer controls? I guess idk how essential file names are to how you have to call the thing you import/export from that Svelte file. Reason I ask is because I think it's nice to have `<XLayerControl>` as the svelte component because it describes what it does.

dabreegster:
We can do `import XyzLayerControl from "./layers/Xyz.svelte"`; the default export can be named however. I'm not too opinionated about it though. https://imfeld.dev/writing/svelte_domless_components is some inspiration for a Svelte component not representing anything in the DOM at all. But in our case, the one component mixes stateful map changes and a bit of DOM

Pete-Y-CS (approved):

PR #312, opened 2023-08-16
Refactor to use MapLibreUtils from an index as a local library

Pete-Y-CS:
this was painful and I now see can't automatically merge
Will rebase tomorrow

### DIFF
--- a/src/lib/layers/SpeedLimits.svelte
+++ b/src/lib/layers/SpeedLimits.svelte
@@ -9,14 +9,10 @@
   } from "maplibre-gl";
   import { onDestroy, onMount } from "svelte";
   import { lineWidth } from "../../colors";
-  import {
-    emptyGeojson,
-    overwriteLineLayer,
-    overwriteSource,
-  } from "../../maplibre_helpers";
   import { gjScheme, map, routeInfo } from "../../stores";
   import type { Feature } from "../../types";
   import { DiscreteLegend, HelpButton, InteractiveLayer } from "../common";
+  import { MapLibreUtils } from "../maplibre";
 
   // Show along a route if specified, or show all otherwise
   export let id: number | undefined;
@@ -62,8 +58,8 @@
   // NOTE! There's only ever one source and layer with this name. This component
   // (and the source and layer) will get destroyed frequently, but even if not,
   // overwriting should be safe.
-  overwriteSource($map, source, emptyGeojson());
-  overwriteLineLayer($map, {
+  MapLibreUtils.overwriteSource($map, source, MapLibreUtils.emptyGeojson());
+  MapLibreUtils.overwriteLineLayer($map, {
     id: layer,
     source,
     color,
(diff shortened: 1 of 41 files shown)


src/lib/layers/SpeedLimits.svelte
dabreegster, line 1: Hmm, this is a huge change, I meant we still import individual functions, but from "../maplibre". We'd have maplibre/index.ts that does stuff to re-export things from multiple files, like maplibre/utils.ts and maplibre/layers_helpers.ts. We can have this "singleton with a bunch of static methods" pattern (not the right term, that'd be Java) if it seems nice or we don't want to verbosely import individual methods, if you like it

src/lib/maplibre/maplibre_utils.ts
```diff
@@ -21,16 +21,16 @@ const defaultColor = "#000000";
 const defaultFilter = true;
 const defaultOpacity = 1;
 
-export const isPolygon: FilterSpecification = ["==", "$type", "Polygon"];
-export const isLine: FilterSpecification = ["==", "$type", "LineString"];
-export const isPoint: FilterSpecification = ["==", "$type", "Point"];
+const isPolygon: FilterSpecification = ["==", "$type", "Polygon"];
```
dabreegster, line 24: I think we'd keep the exports here

src/lib/maplibre/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as MapLibreUtils } from "./maplibre_utils";
```
dabreegster, line 1: I think something like `export * from "./utils"` should work. If it doesn't, we have to do something like `export { isPolygon, isLine, isPoint, ... } from "./utils"`. Can generate that list pretty quickly with grep and a regex if you like

src/lib/layers/SpeedLimits.svelte
dabreegster: Up to you, whatever's easiest -- guessing there's a slight sunk cost here if vscode or regexes wasn't the secret behind this PR so far ;)

Pete-Y-CS: There's a feeling of sunk cost but I just tried `export * from "./utils"` anyway and this didn't work as a first pass. If you're enough of a regex master to furnish me with a list quickly that would be ace

Pete-Y-CS: unlesss... give me a mo I might be wrong there

dabreegster: export { isPolygon, isLine, isPoint, overwriteSource, overwritePmtilesSource, overwriteLayer, overwritePolygonLayer, overwriteCircleLayer, overwriteLineLayer, emptyGeojson, pointFeature, setPrecision, constructMatchExpression, hoveredToggle, makeColorRamp, bbox, prettyPrintMeters, type FeatureWithProps } from "./utils"

PR #316, opened 2023-08-18
Rerework streetview

Pete-Y-CS:
This I think is mostly done. The final thing we might wanna do is adding a custom cursor. Also I have changed layer helper, it now gets all the default values when its created, which I think is sensible behaviour, but only resets values which have actually been changed. I think this is the most straightforward way which isn't taking unnecessary actions. Happy to revisit if you like but think it might be worth waiting to see if it gets used for other things with layers elsewhere


src/lib/common/StreetViewController.svelte
```diff
@@ -0,0 +1,85 @@
...
+  import { Radio } from "../govuk";
+  import DefaultButton from "../govuk/DefaultButton.svelte";
+  import SecondaryButton from "../govuk/SecondaryButton.svelte";
+  import { getRoadLayerHelpers, type LayerHelper } from "../maplibre";
+  import CollapsibleCard from "./CollapsibleCard.svelte";
+
+  export let displayEnableButton: boolean = false;
+  export let escapeKeyExits: boolean = false;
+  const roadLayerHelpers: Array<LayerHelper> = getRoadLayerHelpers();
+  let isActive: boolean;
```
dabreegster, line 13: The functionality is working great! Only big problem is when we have a browse layer turned on. Then a click will have two effects. I think we can fix this by:

1) Creating a store similar to `isAToolInUse` called `interactiveLayersEnabled` or something, defaulting to true
2) Using that in `lib/common/InteractiveLayer.svelte`. For the 3 mouse event handlers there, immediately check this store, and bail out when appropriate.
3) export `isActive` from StreetViewController.svelte and in LayerControls, bind it to the store

src/lib/common/StreetViewController.svelte
```diff
@@ -0,0 +1,85 @@
...
+    $map.off("click", handleMapClickEvent);
+    $map.getCanvas().style.cursor = "inherit";
+    roadLayerHelpers.forEach((roadLayerHelper) => {
+      roadLayerHelper.returnToDefaultPaintValues($map);
+    });
+  }
+</script>
+
+{#if displayEnableButton}
+  <DefaultButton on:click={enableStreetView}>Enable Street View</DefaultButton>
```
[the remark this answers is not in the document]
dabreegster: And should we use `SecondaryButton` when it's inactive? It isn't the main/primary action on the page we want to draw attention to

src/lib/common/StreetViewController.svelte
```diff
@@ -0,0 +1,99 @@
...
+  import type { MapMouseEvent } from "maplibre-gl";
+  import cameraCursorUrl from "../../../assets/camera_cursor.svg?url";
+  import { map, userSettings } from "../../stores";
+  import { Radio } from "../govuk";
+  import SecondaryButton from "../govuk/SecondaryButton.svelte";
+  import { getRoadLayerHelpers, type LayerHelper } from "../maplibre";
+  import CollapsibleCard from "./CollapsibleCard.svelte";
+  import { ExternalLink } from "./";
+
+  export let displayEnableButton: boolean = false;
```
dabreegster, line 11: Small code style nits, we could omit `: boolean` when there's a default value and the type is really clear. And generally we've been doing `T[]` instead of `Array<T>`, but I'm not really opinionated about it. Let's leave it for now, and maybe setup eslint or similar in the future, configure it how we like, and rely on it for enforcement

src/lib/common/StreetViewController.svelte
```diff
@@ -0,0 +1,99 @@
...
+  export let escapeKeyExits: boolean = false;
+  const roadLayerHelpers: Array<LayerHelper> = getRoadLayerHelpers();
+  export let isInactive: boolean;
+
+  export const handleMapClickEvent = (e: MapMouseEvent) => {
+    if (!isInactive) {
```
dabreegster, line 17: Nit: To avoid nesting, I'd slightly prefer `if (isInactive) { return; }`

src/lib/common/StreetViewController.svelte
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+  };
+
+  export function enableStreetView() {
+    isInactive = false;
+    $map.on("click", handleMapClickEvent);
+    const cursorStyle = `url(${cameraCursorUrl}), auto`;
+    console.log(cursorStyle);
+    $map.getCanvas().style.cursor = cursorStyle;
```
dabreegster, line 39:
```suggestion
    $map.getCanvas().style.cursor = `url(${cameraCursorUrl}), auto`;
```

src/lib/common/StreetViewController.svelte
```diff
@@ -0,0 +1,99 @@
...
+    roadLayerHelpers.forEach((roadLayerHelper) => {
```
dabreegster, line 40: Similarly here, we should consistently do `for (let x of list) {` or use `forEach` generally everywhere, but let's not worry about this yet

src/lib/common/StreetViewController.svelte
```diff
@@ -0,0 +1,99 @@
...
+      roadLayerHelper.setProperty($map, "line-color", "blue");
```
dabreegster, line 41: Maybe useful to put in `lib/browse/colors.ts` so when we find somebody with color sense, they can see everything in context. But FWIW I think the blue shows up very nicely ;)

src/lib/draw/StreetViewMode.svelte
```diff
@@ -1,40 +1,21 @@
...
-      );
-    } else if ($userSettings.streetViewImagery == "bing") {
-      window.open(
-        `https://www.bing.com/maps?cp=${lat}~${lon}&style=x`,
-        "_blank"
-      );
-    }
-  };
-
   eventHandler.documentHandlers.keydown = (e: KeyboardEvent) => {
```
dabreegster, line 19: We could consider using the escape key on the browse page too, but no need for it now

src/lib/maplibre/layer_helper_utils.ts
```diff
@@ -0,0 +1,37 @@
+import { get } from "svelte/store";
+import { map as mapStore } from "../../stores";
+import { LayerHelper } from "./layer_helper";
+
+export const roadLayerIds = [
```
dabreegster, line 5:
```suggestion
// These depend on MapTiler basemap styles
export const roadLayerIds = [
```

I see some footpaths missing, we can dig through later and improve this set

src/lib/maplibre/layer_helper.ts
```diff
@@ -0,0 +1,33 @@
+import type { Map } from "maplibre-gl";
+
+export class LayerHelper {
+  id: string;
+  defaultPaintValues: Array<[string, string]>;
+  changedValues: Array<string> = [];
```
dabreegster, line 6: If our approach is to scrape everything initially, then we can remove this -- it's not doing anything right now

src/lib/maplibre/layer_helper.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  constructor(layer: any) {
+    this.id = layer.id;
+    this.defaultPaintValues = this.getLayerValues(layer);
+  }
+
+  getLayerValues(layer: any): Array<[string, string]> {
+    return Object.keys(layer.paint ? layer.paint : {}).map(
```
dabreegster, line 14: Why not just `Object.entries` and maybe deep-cloning for sanity? That gives key/values outright

dabreegster (approved):
Some code style nitpicks, but happy to iterate on these later. The new functionality is so fantastic! \o/

dabreegster (approved):

PR #317, opened 2023-08-22
Streetview fixes

dabreegster:
1) Include paths in the highlighted layer, since it's likely of interest
2) Make the escape key also work in browse mode
3) Fix a bug in the scheme sketching mode where clicking would cause the streetview reaction twice

### DIFF
--- a/src/lib/draw/StreetViewMode.svelte
+++ b/src/lib/draw/StreetViewMode.svelte
@@ -9,7 +9,6 @@
   let streetViewController: StreetViewController;
 
   export function start() {
-    eventHandler.mapHandlers.click = streetViewController.handleMapClickEvent;
     streetViewController.enableStreetView();
   }
   export function stop() {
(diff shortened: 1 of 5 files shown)


src/lib/draw/StreetViewMode.svelte
```diff
@@ -9,7 +9,6 @@
   let streetViewController: StreetViewController;
 
   export function start() {
-    eventHandler.mapHandlers.click = streetViewController.handleMapClickEvent;
```
dabreegster, line 12: The enable / disable functions register a map event handler. So we don't need to do it twice.

Alternatives:
1) Plumb a property to controller saying whether or not the event handler should be registered
2) Make a component for the browse page that registers the callback
3) Have two separate Svelte components for browse and sketch and copy the small bits of code to wire them up in different ways. Share the radio button + help text and `handleMapClickEvent`.

Pete-Y-CS: I think this is the simplest for now so I'm happy with it but it does feel a little weird. 3 Is probably the nicest of these alternatives but I don't know that it's worth it

Pete-Y-CS (approved):

PR #324, opened 2023-08-24
Document the GeoJSON format

dabreegster:


docs/data_spec.md
```diff
@@ -0,0 +1,63 @@
+# ATIP GeoJSON specification
```
Pete-Y-CS, line 1: This looks good on a first skim. Only question is if we want to incorporate schemes, subschemes, and interventions as fully separate types of entity we may need to rework it slightly after this but this seems nice for now at least, and a great basis for editing if we need to

Pete-Y-CS (approved):

docs/data_spec.md
dabreegster: Yep! Just trying to describe what we have today. Hopefully makes it even more clear to then propose a diff for subschemes or something else.

Issue #318, opened 2023-08-23
DfT count layer


dabreegster:
Requested today. Click through to something like https://roadtraffic.dft.gov.uk/manualcountpoints/28512 maybe

Robinlovelace:
Impressive! The DfT count data provides counts for road sections that have start and end points so can be visualised as linestrings, e.g.

![image](https://github.com/acteng/atip/assets/1825120/21ca2286-0a63-417c-b010-369cc5dbf6f6)

Source: https://bikedata.cyclestreets.net/collisions,trafficcounts/#14.14/53.83128/-1.57107

Problem with that: it fails to snap the link to the road.

Robinlovelace:
Equivalent from ATIP:

![image](https://github.com/acteng/atip/assets/1825120/7d47c9da-4ac6-4459-985c-e8a2034f6b5d)


Source: https://acteng.github.io/atip/browse.html?authority=West%20Yorkshire%20Combined%20Authority#14.44/53.80831/-1.54725

Robinlovelace:
If you look at the actual data from the DfT, accessible by clicking on the dot (great feature BTW), you can see it has start and end junction and link length attributes:

![image](https://github.com/acteng/atip/assets/1825120/e57617f6-0fff-44c5-a9ce-4ded43f29f1d)

Source: https://roadtraffic.dft.gov.uk/manualcountpoints/17374

dabreegster:
If it'd be useful to show the linestring segment covered by the count, we could work on that, but we haven't had any feedback about the layer or how it's used yet

Robinlovelace:
I recall seeing Easting/Northing data of junctions at one point, more relevant for cars than bikes, but not sure if they are available in the latest version of the data.

Robinlovelace:
Worth asking if that would be useful. Hypothesis: yes.

PR #325, opened 2023-08-24
Add a vehicle counts layer. #318

dabreegster:
Demo: https://acteng.github.io/atip/vehicle_counts/browse.html

https://github.com/acteng/atip/assets/1664407/2f8c5780-dcdb-4a95-9055-177f21ffe024

This should be enough to start, but plenty of next steps:
- icons instead of points
- display all points when unzoomed (or at least some kind of clustering or heatmap to indicate where the points are)
- adjust the buckets -- not seeing the higher values much


dabreegster:
Started #326 to track problems like few points at low zoom

Pete-Y-CS (approved):
As discussed there are potential improvements but this is a great start and better to get input from inspectors before trying to refine

PR #327, opened 2023-08-29
Disable route info layers (lane details and speed limits) on the sket…

dabreegster:
…ch page, because they both show incomplete data, haven't had much user testing yet, and likely to just cause confusion

Only change is to remove two things from the UI:
Screenshot from 2023-08-29 10-10-12
Screenshot from 2023-08-29 10-10-00

We keep the route info web worker loading in the background, because it **is** useful for the "Auto-fill name" tool for routes.


dabreegster:
Closing this, since we're not imminently releasing anything new

PR #328, opened 2023-08-29
WIP: Add scheme-wide ATF4 category checkboxes

dabreegster:
Demo at https://acteng.github.io/atip/atf4_checkboxes/scheme.html?authority=Adur
Screenshot from 2023-08-29 10-49-29

Questions:

- This is a collection of checkboxes at the scheme level -- is that what we want? (Not at the intervention or sub-scheme level?) Do we want to require people to select at least one?
- Do we also want to ask people the high/medium/low complexity question for these?
- How should we encode the set in JSON?
  - An object with boolean values is awkward, because a missing key and a value of `false` mean the same thing. Maybe a sorted list of strings?
  - We can decouple the text shown on the page from the way we store the result. Is the mapping I made up reasonable? How's most convenient to later represent in the DB?
- Would a larger popup/modal be nicer than embedding the long list in the sidebar?
- Any changes to the exact wording of things?

### DIFF
--- a/docs/data_spec.md
+++ b/docs/data_spec.md
@@ -15,6 +15,8 @@ ATIP produces a [GeoJSON file](https://datatracker.ietf.org/doc/html/rfc7946). T
   - An optional string `authority`, belonging to [this set](https://github.com/acteng/atip/blob/main/assets/authority_names.json)
   - An optional free-form string `scheme_name`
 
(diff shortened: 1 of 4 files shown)


dabreegster:
We don't need this immediately, so putting back on the back-burner

Issue #323, opened 2023-08-23
View critical issues


dabreegster:
We can take the single .xslx that exists today and have people load it in the page, as we do with `all_schemes.geojson`. Later this can come from an API.

- Just display points, with tooltips + clicking
- Explore filter/color options (by "resolved" or not, maybe)
- One user story: how many criticals have we identified or resolved within 500m of a school?

PR #329, opened 2023-08-30
View critical issues

dabreegster:
#323. Demo at https://acteng.github.io/atip/view_criticals/browse.html, loading `Inspectors Log.xlsx` on the right. I'd encourage playing around with just the criticals shown, and both criticals and scheme data.

At low zooms, this clusters the points. You can click a cluster to expand. Once you're far enough in, individual criticals are shown as circles. Hovering on them gives a popup with all relevant info.

I also tweaked a few other things in this PR:
- Letting users hide the scheme data like a layer. Sometimes it gets in the way.
- Making error messages show up with govuk styles

There are other things I want to do with this layer, but probably as a followup PR:
- When you click a critical, make the popup "sticky" until you close it, so you could go copy text from it or something. (And we probably want similar for the scheme data.)
- Using icons instead of red circles (useful for other layers too)
- Larger user stories, like selecting a 500 meter buffer around a school and summarizing criticals. Or reporting of criticals by scheme, authority, etc.


package.json
```diff
@@ -49,6 +49,7 @@
     "humanize-string": "^3.0.0",
     "maplibre-gl": "^3.1.0",
     "pmtiles": "^2.10.0-beta.0",
+    "read-excel-file": "^5.6.1",
```
dabreegster, line 52: There's https://www.npmjs.com/package/xlsx and some other heavier options, but this one seems perfectly adequate for us

src/lib/browse/Filters.svelte
```diff
@@ -141,7 +162,11 @@
   <InterventionColorSelector />
 </CollapsibleCard>
 
-<p>
-  Showing {schemesToBeShown.size} schemes ({counts.route} routes, {counts.area} areas,
-  {counts.crossing} crossings, {counts.other} other)
-</p>
+<CheckboxGroup small>
+  <Checkbox id="show-schemes" bind:checked={show}>
+    Showing {schemesToBeShown.size.toLocaleString()} schemes ({counts.route.toLocaleString()}
```
dabreegster, line 167: Also adding commas every 3 orders of magnitude here, for consistency with other places in ATIP

src/lib/browse/layers/CriticalIssuesLayerControl.svelte
```diff
@@ -0,0 +1,153 @@
...
+  let show = true;
+  let numberIssues = 0;
+  let errorMessage = "";
+
+  cleanupSource($map, source);
+  $map.addSource(source, {
+    type: "geojson",
+    data: emptyGeojson(),
+    cluster: true,
+    clusterMaxZoom: 14,
```
dabreegster, line 31: Generally following https://maplibre.org/maplibre-gl-js/docs/examples/cluster/. Default params here seem fine to me.

src/lib/browse/layers/CriticalIssuesLayerControl.svelte
```diff
@@ -0,0 +1,153 @@
...
+      ($map.getSource(source) as GeoJSONSource).setData(gj);
+      numberIssues = gj.features.length;
+      errorMessage = "";
+    } catch (err) {
+      errorMessage = `The file you loaded is broken: ${err}`;
+    }
+  }
+
+  function tooltip(feature: MapGeoJSONFeature): string {
+    let x = `<div style="max-width: 30vw;">`;
```
dabreegster, line 90: Inconsistent use of bold/underline on the value / label for the value. Style suggestions welcome!

Issue #322, opened 2023-08-23
Critical issue data entry


dabreegster:
Investigate short-term options for entering criticals.

Input fields are:
- scheme reference and design stage
- one of 15 criticals (and then there are sub-questions on some)
- freeform notes
- a lat, long
- name of nearest street / intersection

Last two are slightly tedious to generate.

We can focus on making a nice UI for doing this and punt on how we submit the data somewhere. I thought we might be able to open an MS Forms URL and attach query parameters to pre-fill some fields, but apparently this may not be supported -- still need to try. Even still, it's not throwaway work, since eventually we'll be able to submit to an API and stash in a proper DB.

Pete-Y-CS:
Okay after discussion with Dustin we have agreed on some initial steps to implement then get feedback for iteration:

- At first we will implement a new page, which allows you to drop a pin then fill in a form.
- This will not be hooked up to any actual database or output, it's just in anticipation of scheme updater paving the way for us to have DB access
- You will be able to access this page directly, or by selecting a scheme on the scheme browse page, and the page will be opened centred on that scheme's location

A second stage of work will allow the user to see existing entered criticals, again not actually hooked up to a backend but demonstrating how it would work with a working DB integration

PR #330, opened 2023-09-01
Start a page for critical issue entry

dabreegster:
Demo at https://acteng.github.io/atip/critical_entry_mvp/critical_entry.html. A start to #322:

https://github.com/acteng/atip/assets/1664407/7c6b7b41-d9e9-4dfc-9b5a-092e6dab962b

Pasting the resulting row in the xlsx appears to work for me, but of course we should test more carefully end-to-end before recommending people use this.

Future steps:

- [ ] Make it nicer to enter multiple criticals at a time. Need to decide if previously confirmed ones should be editable or not.
- [ ] Showing existing criticals and/or the relevant scheme data could be useful. Sourcing this data, or how people would arrive at the page, is unclear before we have a DB / API.

### DIFF
--- a/src/pages/CriticalIssueEntry.svelte
+++ b/src/pages/CriticalIssueEntry.svelte
@@ -21,7 +21,7 @@
     <h1>Critical issue entry</h1>
     {#if markerPosition}
       <p>Critical at {markerPosition}</p>
-      <Form />
+      <Form pt={markerPosition} />
     {:else}
       <p>Click the map to add a critical</p>
     {/if}
--- a/src/pages/CriticalIssueEntry.svelte
+++ b/src/pages/CriticalIssueEntry.svelte
@@ -27,7 +27,7 @@
   }
 </script>
 
-<Layout>
+<Layout sidebarWidth="35rem">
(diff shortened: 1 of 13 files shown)


src/pages/CriticalIssueEntry.svelte
```diff
@@ -0,0 +1,64 @@
...
+  import Form from "../lib/critical_entry/Form.svelte";
+  import Pin from "../lib/critical_entry/Pin.svelte";
+
+  onMount(() => {
+    // For govuk components. Must happen here.
+    initAll();
+  });
+
+  let markerPosition: LngLat | null = null;
+  let streetviewOff = true;
```
dabreegster, line 18: Bit of odd mutual exclusion logic here. If stview is on, clicking shouldn't create a new pin. And while stview is on, you can't switch basemap, because the road helper stuff would then need to detect that change and fiddle with the correct layers.

dabreegster:
Rebased, and fixed a bug Pete found where the contents of the textbox in the popup didn't update (but the clipboard did)

Issue #257, opened 2023-06-29
Investigate GCP hosting strategies


dabreegster:
Goal: Serve a static website on GCP, but restrict access to it.

# GCS + GAE + IAP

The simplest approach conceptually would be to use a private GCS bucket, then serve it using IAP. It appears this isn't supported: https://issuetracker.google.com/issues/114133245?pli=1

The next simplest approach I can find is to use AppEngine. The static website would be hosted in a private GCS bucket, we'd use [AppEngine to serve it](https://cloud.google.com/appengine/docs/legacy/standard/python/getting-started/hosting-a-static-website), then we'd use IAP to sit in front of AppEngine.

I haven't yet found docs about gov.uk integration with IAP. The [One Login docs](https://docs.sign-in.service.gov.uk/) are lower level, talking about OAuth flows. My hope is that the IAP layer can completely take care of that for us.

# Adding a backend

Many of the backend APIs we may consider could be hosted in Cloud Run or Cloud Functions or similar, since they'd be stateless servers just talking to a DB.  IAP can also sit in front of Cloud Run, it appears. So the frontend code running in the browser could make requests to the backend, also going through IAP and verifying the identity again.

dabreegster:
This is past mvp, and remaining TODOs are tracked in `backend/README.md`

PR #331, opened 2023-09-04
Move API key and resource base URL into vite config, in preparation f…

dabreegster:
…or #257

This uses https://vitejs.dev/guide/env-and-mode.html, and has no behavioral change. It's just a simple change that can be done independently of the rest of the GCP deployment work.

When we're ready, we can define a `.env.prod` file with a different resource base URL (the private GCS bucket).

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+VITE_MAPTILER_API_KEY="MZEJTanw3WpxRvt7qDfo"
+VITE_RESOURCE_BASE="https://atip.uk"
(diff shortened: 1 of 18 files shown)


Pete-Y-CS:
Do we need to start using gitcrypt to store things like API keys? Still this change is orthogonal to that I think

dabreegster:
> Do we need to start using gitcrypt to store things like API keys?

No, because the API key shows up in what we serve. It's trivial to figure it out even if we obfuscated the JS sent to the client, because the browser will make requests using the key. https://documentation.maptiler.com/hc/en-us/articles/360020806037-How-to-protect-your-map-key describes how to protect the key -- setting up allowed domains. #59 tracks that; it hasn't been a problem yet.

PR #332, opened 2023-09-04
Stop using relative module imports

dabreegster:
@yongrenjie pointed me at this awesome trick, which'll let us stop using relative paths for imports and instead treat `src/` as our root. This would make changing directory structure _way_ easier. Pete, if this sample change looks good and it works in your IDE, I'll make it everywhere!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,8 @@
         "ts-node": "^10.9.1",
         "typescript": "^5.1.3",
         "vite": "^4.1.1",
-        "vite-plugin-wasm-pack": "0.1.11"
+        "vite-plugin-wasm-pack": "0.1.11",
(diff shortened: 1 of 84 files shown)


Pete-Y-CS:
This looks amazing! Will have a quick look locally but I don't see why it would be an issue! Sure I can make VSCode work with it

Pete-Y-CS:
Yup gave it a quick test drive! Fire at will!

dabreegster:
Alright, used pretty much everywhere, except in a few cases for "package private" stuff. Manual testing shows no behavioral diff. Once playwright tests pass, I'll merge.

As a future followup, I'd like to consider moving most things in `src/*ts` into `src/lib`. There's not a great reason for the separation right now, and importing from "types", "stores", etc feels confusingly like it could be an NPM package.

Issue #337, opened 2023-09-12
Addition data layers


robinlovelace-ate:
- [ ] Traffic volumes
- [ ] Traffic speeds
- [ ] Crossing frequency (linked to critical of crossing per 500 m)
- [ ] PCT layer, see https://www.carbon.place/#11.8/51.50321/-0.09312
- [ ] Historic LTNs, see bikedata

dabreegster:
- Speed: https://www.ordnancesurvey.co.uk/products/os-mastermap-highways-network-with-speed-data
- Crossing frequency is a coloring per road. Not sure what counts as a good crossing

PR #339, opened 2023-09-14
Add a PCT layer. #337

dabreegster:
Demo: https://acteng.github.io/atip/pct/browse.html (should be deployed by ~12:45)

### DIFF
--- a/src/lib/browse/LayerControls.svelte
+++ b/src/lib/browse/LayerControls.svelte
@@ -18,6 +18,7 @@
   import LocalPlanningAuthoritiesLayerControl from "./layers/LocalPlanningAuthoritiesLayerControl.svelte";
   import MrnLayerControl from "./layers/MrnLayerControl.svelte";
   import ParliamentaryConstituenciesLayerControl from "./layers/ParliamentaryConstituenciesLayerControl.svelte";
+  import PctLayerControl from "./layers/PctLayerControl.svelte";
   import RailwayStationsLayerControl from "./layers/RailwayStationsLayerControl.svelte";
   import SchoolsLayerControl from "./layers/SchoolsLayerControl.svelte";
   import SportsSpacesLayerControl from "./layers/SportsSpacesLayerControl.svelte";
@@ -69,6 +70,7 @@
   </CollapsibleCard>
   <CollapsibleCard label="Other">
     <VehicleCountsLayerControl />
+    <PctLayerControl />
   </CollapsibleCard>
   <StreetViewController
     bind:this={streetViewController}
--- a/src/lib/browse/layers/PctLayerControl.svelte
+++ b/src/lib/browse/layers/PctLayerControl.svelte
@@ -74,9 +74,9 @@
   <span slot="right">
     <HelpButton>
       <p>
-        The PCT shows roads where cycling levels are predicted to increase,
-        given different assumptions about investments. The layer shows daily
-        counts for commuting to work or school.
+        The PCT shows major links on the transport network on which there is high
+        latent demand for cycling under scenarios of change, based on the 'fastest'
+        routes from CycleStreets. Counts approximate daily rush hour traffic levels.
       </p>
       <p>
         See <ExternalLink href="https://www.pct.bike">
--- a/src/lib/browse/layers/PctLayerControl.svelte
+++ b/src/lib/browse/layers/PctLayerControl.svelte
@@ -1,0 +1,130 @@
+<script lang="ts">
+  import { ExternalLink, HelpButton, InteractiveLayer } from "lib/common";
+  import { Checkbox, Radio, Select } from "lib/govuk";
+  import {
+    hoveredToggle,
+    makeColorRamp,
+    overwriteLineLayer,
+    overwritePmtilesSource,
+  } from "lib/maplibre";
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import { map } from "stores";
+  import { colors, denseLineWidth } from "../colors";
+  import SequentialLegend from "./SequentialLegend.svelte";
+
+  // TODO It'd be much simpler to have one source with both attributes
+  let nameCommute = "pct_commute";
+  let nameSchool = "pct_school";
+  let colorScale = colors.sequential_low_to_high;
+  let limits = [0, 50, 100, 500, 1000, 2000];
+
+  let tripPurpose = nameCommute;
+  let scenario = "baseline";
+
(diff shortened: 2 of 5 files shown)


src/lib/browse/LayerControls.svelte
```diff
@@ -69,6 +70,7 @@
   </CollapsibleCard>
   <CollapsibleCard label="Other">
```
dabreegster, line 71: "Statistics"? "Counts"?

src/lib/browse/layers/PctLayerControl.svelte
```diff
@@ -0,0 +1,130 @@
...
+    makeColorRamp,
+    overwriteLineLayer,
+    overwritePmtilesSource,
+  } from "lib/maplibre";
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import { map } from "stores";
+  import { colors, denseLineWidth } from "../colors";
+  import SequentialLegend from "./SequentialLegend.svelte";
+
+  // TODO It'd be much simpler to have one source with both attributes
```
dabreegster, line 15: See https://github.com/acteng/atip-data-prep/pull/34 for the data import

src/lib/browse/layers/PctLayerControl.svelte
```diff
@@ -0,0 +1,130 @@
...
+  let nameCommute = "pct_commute";
+  let nameSchool = "pct_school";
+  let colorScale = colors.sequential_low_to_high;
+  let limits = [0, 50, 100, 500, 1000, 2000];
```
dabreegster, line 19: I looked at the buckets used by https://www.carbon.place, but we need 6 for our current color scale. Kind of made these up, but they seem reasonably useful

src/lib/browse/layers/PctLayerControl.svelte
```diff
@@ -0,0 +1,130 @@
...
+    );
+  }
+</script>
+
+<Checkbox id={nameCommute} bind:checked={show}>
+  Propensity to Cycle Tool
+  <span slot="right">
+    <HelpButton>
+      <p>
+        The PCT shows roads where more cycling levels are predicted to increase,
```
dabreegster, line 77: @robinlovelace-ate , any suggestions on what to say here? What can we quickly say about data freshness -- based on 2011 census? And what's the license on the final output data we're using here?

dabreegster (commented):
CC @stuartlynn for UX opinions. This is another example of a layer with extra controls and a legend that appears when enabled

src/lib/browse/layers/PctLayerControl.svelte
Pete-Y-CS: This current wording is a bit confusing to me. At the very least I think dropping the word `more` would help. I think also either here, by the legend, or ideally both say that it's (presumably) cycle trips per day?

src/lib/browse/LayerControls.svelte
Pete-Y-CS: I think the problem is that that there are things that fit those descriptions in other categories. Maybe merge with census and call em statistics?

Incidentally, should criticals really be over on the right now that I'm thinking about these layers categories? It seems like there's a delineation between left and right: left is ATE data which we might want to interact with directly, right is precomputed context data which is static as far as ATIP is concerned?

Pete-Y-CS (commented):
Code looks good but a couple of UX/wording thoughts

src/lib/browse/layers/PctLayerControl.svelte
dabreegster: Sliightly adjusted it, any better? I'm also not 100% sure the data is daily trips, because the PCT output files I grabbed don't have data definitions anywhere handy, and I haven't found the right bits of the methodology paper.

src/lib/browse/LayerControls.svelte
dabreegster: I agree the left/right differences are very weird right now. @stuartlynn described some nice ideas about rethinking the UI more generally here. First a user would load data and search through layers to enable. (And that bit could do a better job maybe of preventing you from enabling multiple boundary layers at once, or a boundary and a census layer.)  Then we have a single sidebar/control that only has a legend and any filters/controls for those layers. The user could go back and opt into different layers.

If we went that route, it'd also be a potentially nice time to try and encode the current settings for enabled layers (and their settings) in some way that could be shared as a permalink, expressed in the URL, etc.

src/lib/browse/layers/PctLayerControl.svelte
Pete-Y-CS: Looks better to me, pending confirmation of it being accurate (presumably from Robin)!

src/lib/browse/LayerControls.svelte
stuartlynn: Going to try and have a prototype for this approatch early next week so we can discuss it's merits and if it feels like an improvment over the current approatch

src/lib/browse/layers/PctLayerControl.svelte
Robinlovelace: Let me check what it says now...

src/lib/browse/layers/PctLayerControl.svelte
```diff
@@ -0,0 +1,130 @@
...
+</script>
+
+<Checkbox id={nameCommute} bind:checked={show}>
+  Propensity to Cycle Tool
+  <span slot="right">
+    <HelpButton>
+      <p>
+        The PCT shows roads where cycling levels are predicted to increase,
+        given different assumptions about investments. The layer shows daily
+        counts for commuting to work or school.
```
Robinlovelace, line 79:
```suggestion
        The PCT shows major links on the transport network on which there is high
        latent demand for cycling under scenarios of change, based on the 'fastest'
        routes from CycleStreets. Counts approximate daily rush hour traffic levels.
```

Reasoning: prediction is not the right word, all depends on the backcasting scenario and deterministic routing. Also captures that these are 'one way' flows which I didn't add because that in itself is confusing and loaded implying direction which the PCT lacks, rush hour traffic is reasonable. The fact that this is the fastest network is important. Do we also want to mention that this is based on old OSM data (not just old OD data)?

Robinlovelace (approved):

src/lib/browse/layers/PctLayerControl.svelte
dabreegster: Thanks! I added a note about old OSM data too and a general caution. Hope we can get the next-gen thing out there soon :)

PR #340, opened 2023-09-14
Show the version of the app

dabreegster:
Now that we have more deployment environments, make it clear where we are. Also provide some warnings and contact emails.


src/lib/browse/AppVersion.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+  import { HelpButton } from "lib/common";
+
+  let version = "unknown";
+  if (window.location.hostname == "localhost") {
+    version = "Local development";
+  } else if (window.location.hostname == "acteng.github.io") {
+    if (window.location.pathname == "/atip/browse.html") {
+      version = "Current development (public)";
```
dabreegster, line 9: Better terms/names welcome

src/pages/BrowseSchemes.svelte
```diff
@@ -2,6 +2,7 @@
   // @ts-ignore no declarations
   import { initAll } from "govuk-frontend";
   import "../style/main.css";
+  import AppVersion from "lib/browse/AppVersion.svelte";
```
Pete-Y-CS, line 5: Should this be on all the pages? I know what we currently have is specific to browse schemes but it feels like we might want something similar in other places?

dabreegster: PTAL -- changed a few things about detecting the environment, mentioning it on multiple pages, etc

Pete-Y-CS (approved):

src/lib/browse/AppVersion.svelte
dabreegster: Thanks! Updated the names in 88cad1af11a81a39d29c90bec285ad5e9d616b54

Issue #326, opened 2023-08-24
Adjust layers at low zoom


dabreegster:
Some related problems:

- at low zooms, trip generator layers based on polygons (the first 3) are not really visible
- at low zooms, point layers like cycle parking and traffic counts (and soon stats19) appear to just sample points. It's unclear where to zoom to see more detail
- Use of color + circles/polygons to represent many of these layers isn't ideal. Symbols/icons could be much nicer.
- at low zooms, some objects have tooltips and are clickable, but this feels inappropriate -- it's practically impossible to tell what object is being hovered on when it's so tiny

Possible solutions (not mutually exclusive):

- https://maplibre.org/maplibre-gl-js/docs/examples/cluster/ for points. From quickly trying this out, this only works on GeoJSON sources, not vector tiles like pmtiles. So only OK if we can afford to load the whole geojson file at once.
- https://maplibre.org/maplibre-gl-js/docs/examples/heatmap-layer/ for points. This does seem to work with pmtiles.
- Adding `-zg` to `tippecanoe` makes more/all points show up at low zooms. If the size per tile gets too big, we could maybe consider dropping some attributes at low zooms -- especially if we decide we don't need hovering/clicking at that level. https://github.com/felt/tippecanoe has lots of docs/examples.
- Use maplibre symbol layers

PR #343, opened 2023-09-15
Crossings layer for browse page

Pete-Y-CS:
Crossings with click-through to OSM. Needs changes to live hosting but this should be pretty representative


src/lib/browse/layers/PointAmenityLayerControl.svelte
```diff
@@ -12,23 +12,27 @@
...
 
+  export let url: string;
+
+  export let tooltip: (feature: MapGeoJSONFeature) => string;
+
+  const originalOnClick = (e: CustomEvent<MapGeoJSONFeature>) => {};
+
+  export let onClick = originalOnClick; 
+
+  $: clickable = onClick !== originalOnClick;
```
Pete-Y-CS, line 28: This is a bit funky I'll admit, might be a better way to do

src/lib/browse/layers/PointAmenityLayerControl.svelte
```diff
@@ -38,11 +42,6 @@
     // TODO Outline?
   });
   let show = false;
-
-  function tooltip(feature: MapGeoJSONFeature): string {
```
Pete-Y-CS, line 42: Although we didn't need it before, I definitely think it makes more sense for the specific layer control to determine how the tooltip is constructed

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,64 @@
+<script lang="ts">
+  import { circleRadius } from "colors";
```
dabreegster, line 2: The dots are hard to hover on, even at tiny zooms. How do we feel about something like `["interpolate", ["linear"], ["zoom"], 1, 2, 8, 3, 13, 10]` from the cycle parking layer? (In one of the other PRs out right now, I refactored another of these zoom-dependent patterns -- we could do the same). Also related to #326 more generally for UX at some zoom levels

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,64 @@
...
+  function tooltip(feature: MapGeoJSONFeature): string {
+    const crossingTypeString: string = feature.properties["crossing"];
+    if (crossingTypeString === "no")
+      return "Location where crossing is impossible/illegal but where there is a clear desire line to cross.";
+    let result = "Crossing with the following features.";
+    const initialTooltipSentence =
+      initialTooltipSentences.get(crossingTypeString);
+    if (initialTooltipSentence) result = initialTooltipSentence;
+
+    return result + `. More information can be found by clicking.`;
```
dabreegster, line 19: I think you need to put the entire thing in a `<p>` so govuk styles apply. The text is small right now.

dabreegster, line 19: "Click for details" is shorter and direct?

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,64 @@
...
+  const initialTooltipSentences: Map<string, string> = new Map<string, string>([
+    ["traffic_signals", "Signalised crossing"],
+    ["marked", "Crossing with no traffic signals"],
+    ["uncontrolled", "Crossing with no traffic signals"],
+    ["unmarked", "Crossing with no markings or signals"],
+    ["zebra", "Zebra crossing"],
+    ["island", "Crossing with an island"],
+    [
+      "informal",
+      "Infomral crossing where there crossing is not intended by the municipality but there is an obviously desire line",
```
dabreegster, line 31: Typos and odd grammar...

"Informal crossing with an obvious desire line, but no official infrastructure to support it" ?

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,64 @@
...
+      return "Location where crossing is impossible/illegal but where there is a clear desire line to cross.";
+    let result = "Crossing with the following features.";
+    const initialTooltipSentence =
+      initialTooltipSentences.get(crossingTypeString);
+    if (initialTooltipSentence) result = initialTooltipSentence;
+
+    return result + `. More information can be found by clicking.`;
+  }
+
+  const initialTooltipSentences: Map<string, string> = new Map<string, string>([
```
dabreegster, line 22: In general in TS, I don't think you need to do `let x: Type = new Foo<Generic>(...` -- probably just the first is sufficient.

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,64 @@
...
+  import { ExternalLink } from "lib/common";
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import crossingsUrl from "../../../../assets/crossings.geojson?url";
+  import PointAmenityLayerControl from "./PointAmenityLayerControl.svelte";
+
+  const name = "crossings";
+
+  function tooltip(feature: MapGeoJSONFeature): string {
+    const crossingTypeString: string = feature.properties["crossing"];
+    if (crossingTypeString === "no")
```
dabreegster, line 12: Does `npm run fmt` keep this? Lack of braces for an `if` are weird

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,64 @@
...
+  import { circleRadius } from "colors";
+  import { ExternalLink } from "lib/common";
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import crossingsUrl from "../../../../assets/crossings.geojson?url";
+  import PointAmenityLayerControl from "./PointAmenityLayerControl.svelte";
+
+  const name = "crossings";
+
+  function tooltip(feature: MapGeoJSONFeature): string {
+    const crossingTypeString: string = feature.properties["crossing"];
```
dabreegster, line 11: Does type inference not figure out `string` here? Also you can do `properties.crossing` to be more consistent with other code

src/lib/browse/layers/CrossingsLayerControl.svelte
dabreegster, line 19: We've been using `${result}. Foo bar xyz` for string interpolation, not concat generally, in this codebase

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,64 @@
+<script lang="ts">
+  import { circleRadius } from "colors";
+  import { ExternalLink } from "lib/common";
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import crossingsUrl from "../../../../assets/crossings.geojson?url";
```
dabreegster, line 5: Can switch to pmtiles in the usual place now

src/lib/browse/layers/PointAmenityLayerControl.svelte
dabreegster, line 1: I don't think this component is a useful indirection, and some of the awkwardness plumbing through different options for railways vs crossings comes up here. `CycleParkingLayerControl.svelte` is a simple and short example not using the indirection.

src/lib/browse/layers/PointAmenityLayerControl.svelte
dabreegster: My preference is to not have this indirection at all (see above). But if we keep it, this doesn't need to be reactive, because `InteractiveLayer` and such don't expect things like this to change after creation. It could just be a regular `let clickable = ....` calculated once.

assets/crossings.geojson
dabreegster, line 1: Can remove this now

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,93 @@
...
+  import type { MapGeoJSONFeature } from "maplibre-gl";
+  import { map } from "stores";
+  import { colors } from "../colors";
+
+  const name = "crossings";
+  let show = false;
+
+  function tooltip(feature: MapGeoJSONFeature): string {
+    const crossingTypeString = feature.properties.crossing;
+    if (crossingTypeString === "no")
```
dabreegster, line 19: Why is this special and not in the mapping below?  (and in case my other comment got lost, the style of using `if` without braces is different than everywhere else in the codebase. `prettier` doesn't change it right now, so just be aware it introduces inconsistency

src/lib/browse/layers/CrossingsLayerControl.svelte
```diff
@@ -0,0 +1,93 @@
...
+      return "Location where crossing is impossible/illegal but where there is a clear desire line to cross.";
+    let result = "Crossing with the following features.";
```
dabreegster, line 21: This looks out-of-date and almost never used. I think you can just  change all of this code to do `descriptions.get(crossingType) ?? crossingType`

src/lib/browse/layers/RailwayStationsLayerControl.svelte
```diff
@@ -1,28 +1,64 @@
...
-      Open Data Commons Open Database License
-    </ExternalLink>
-  </p>
-</PointAmenityLayerControl>
+<InteractiveLayer
+  layer={name}
+  {tooltip}
+  {show}
+  clickable={false}
+  on:click={() => {}}
```
dabreegster, line 63: You can remove this

dabreegster (approved):
Looks good after a few more fixes. Nice!

PR #347, opened 2023-09-18
Make use of full stop consistent

Pete-Y-CS:


dabreegster (commented):
Ah good catch! But I would vote going the other way; keep the period in the tooltip function, not each description?

dabreegster (approved):

Issue #319, opened 2023-08-23
Speed limit layer


dabreegster:
Mentioned today. OSM doesn't have this filled out everywhere. Check into the public OS datasets and see if it's there.

dabreegster:
Done, using private OS data

PR #348, opened 2023-09-20
Start a layer for OS road widths and speeds

dabreegster:
#319. https://github.com/acteng/atip-data-prep/pull/37 had the data prep.

Using this requires `private_layers/v1/road_widths.pmtiles` and `private_layers/v1/road_speeds.pmtiles` in the GCS bucket, or linking to a local file in `assets` and doing the `import url from "../../../../assets/road_widths.pmtiles?url";` trick temporarily.

Both layers have dataviz issues, but I'm hoping to get feedback ASAP and iterate.


src/lib/browse/layers/RoadWidthsLayerControl.svelte
```diff
@@ -0,0 +1,70 @@
...
+    name,
+    `${import.meta.env.VITE_RESOURCE_BASE}/private_layers/v1/${name}.pmtiles`
+  );
+
+  overwriteLineLayer($map, {
+    id: name,
+    source: name,
+    sourceLayer: name,
+    color: makeColorRamp(["get", "average"], limits, colorScale),
+    // TODO Try showing the actual width, in meters and not pixels
```
dabreegster, line 30: In other words, once you're zoomed in, you see "actual" carriageways as polygons, not as thickened lines that're always the same pixel-size at any zoom. I briefly played around with this:
```
-    width: [
-      "interpolate",
-      ["exponential", 2],
-      ["zoom"],
-      10,
-      ["*", ["get", "average"], ["^", 2, -6]],
-      24,
-      ["*", ["get", "average"], ["^", 2, 8]],
-    ],
```

but it's going to take more work. It's useless (and slow) at low zoom, and "zoom" can only be used in top-level interpolate/step expressions. I think we might have to use two separate line layers glued to the same source, with different min/max zooms. Leaving as future work

Pete-Y-CS: That's a super cool idea big fan

Pete-Y-CS (approved):

Pete-Y-CS (changes requested):
We need a readme update for private_layers!

dabreegster:
> We need a readme update for private_layers!

Good call! Added

Issue #205, opened 2023-05-31
Add more basemaps


robinlovelace-ate:
Currently we only have 2 basemaps.

I think basemaps that show a bit more detail on existing infrastructure could be good, e.g.:

- [ ] TransportMap which has good data on public transport routes: https://wiki.openstreetmap.org/wiki/Transport_Map
- [ ] CyclOSM: https://leaflet-extras.github.io/leaflet-providers/preview/#filter=CyclOSM

dabreegster:
More basemaps would be great! I'd also love to find a more detailed satellite layer for the UK that we have a license to use.

Implementation notes:
- Right now we're wired to use maptiler styles: https://github.com/acteng/atip/blob/cbdd647ee4231975cbdb81628dbb3d0414f3aefe/src/lib/Map.svelte#L20. We can of course support more.
- Need to update map attribution text
- Need to make sure the layer order works with other sources: https://github.com/acteng/atip/blob/cbdd647ee4231975cbdb81628dbb3d0414f3aefe/src/maplibre_helpers.ts#L240
- All of our color scheme / design is mostly tuned (very loosely speaking!) for the default style, and looks bad with the satellite option. All of the colors/styling may need to change to reasonably fit other basemaps.

dabreegster:
Closing old issue. We've got Google satellite now

PR #349, opened 2023-09-20
Add 3 private OS raster basemaps

dabreegster:
#205.

To use this, first follow the new instructions in backend docs and make the `api_keys.json` in GCS.

The new layers look great! Unsolved future work:

- [ ] Adjust all our colors/styling to handle the new styles
- [ ] Investigate OS vector tiles, so we don't cover up road/place labels with our stuff
- [ ] Limit the bounds you can pan, since some of the raster layers don't cover anything outside England


src/lib/maplibre/layer_helper_utils.ts
```diff
@@ -34,6 +34,7 @@ function streetsLayers() {
 
 const satelliteLayers = ["Path", "Road", "Tunnel"];
 
+// OS raster layers don't have anything we can use
```
dabreegster, line 37: So the Streetview tool just doesn't highlight anything

src/lib/maplibre/utils.ts
```diff
@@ -287,6 +291,63 @@ export function prettyPrintMeters(x: number): string {
...
+      import.meta.env.VITE_MAPTILER_API_KEY
+    }`;
+  }
+
+  let resp = await fetch(
+    `${import.meta.env.VITE_RESOURCE_BASE}/private_layers/api_keys.json`
+  );
+  let apiKeys = await resp.json();
+
+  // OS Raster styles
```
dabreegster, line 323: From https://labs.os.uk/public/os-data-hub-examples/os-maps-api/zxy-example-change-style#maplibre-gl-js

src/pages/CriticalIssueEntry.svelte
```diff
@@ -20,6 +20,9 @@
     initAll();
   });
 
+  let defaultStyle =
```
dabreegster, line 23: I thought about doing the same for the browse page, but because most of our layer styling choices are tuned for the greyscale dataviz one, don't want to change it yet.

Pete-Y-CS (approved):

PR #350, opened 2023-09-21
Improve developer experience running locally

dabreegster:
This PR solves two problems:

1) When working on a new layer or fixing an existing one, you have to manually fiddle with the code to import a local file by URL and use that.
2) When running locally, you don't see private layers, OS basemaps, etc.

With this PR, you can do `VITE_RESOURCE_BASE="" VITE_MIMIC_GCP_LOCALLY="true" npm run dev` and fix these problems. You must first set up `public/layers` and `public/private_layers` as described. (I'm using symlinks for this, personally, and linking to atip-data-prep.)

**Important**: I changed the expected path of `random_schemes.geojson` to live inside `private_layers`. Please run `gsutil mv gs://dft-rlg-atip-dev/random_schemes.geojson gs://dft-rlg-atip-dev/private_layers/v1/random_schemes.geojson` before deploying this PR.


dabreegster:
Alright, rebased and ready for review!

Pete-Y-CS (approved):

PR #352, opened 2023-09-25
Remove unused contextual layers from the sketch page

dabreegster:
Cleans up old experiments on the scheme sketch page:

- a speed limit layer (map-wide or for a route)
- an attempt to render lane-level details using osm2streets along a drawn route

The reasoning:
- The functionality has always been experimental / unused
- We've done way more work on contextual layers since then for the browse page. If any might be important to use for sketching, we could take advantage of the new work (and improved tech/tricks, like pmtiles covering all of England)
- The speed limit data from OSM was always patchy. We have better private data from OS now
- The map-matching to show speed or lane rendering along a particular route has always had bugs, and needs tests
- Less WASM to build, faster page load. I'd like to reconsider the remaining use of the `route_info` service for auto-naming the routes. That feature works great and is very handy, but there might be a cheaper way for us to do the same thing.


route_info/src/map_matching.rs
```diff
@@ -27,55 +25,4 @@ impl RouteInfo {
             bail!("not a LineString");
         }
     }
-
-    /// Interpret waypoints from route-snapper in the context of this StreetNetwork.
```
dabreegster, line 31: Some of the code could be useful in some unknown future, but git means we lose nothing, and it's unused in the meantime

Pete-Y-CS (approved):
Nothing more beautiful than many red-highlighted lines in a PR

PR #354, opened 2023-09-26
Line measuring

Pete-Y-CS:
Draw lines on the map and show how long they are. Currently only in browse page. Just as I was typing this realised this is going to be breaking for Criticals so making it a WIP whilst I make Critical Entry work with new way that Pins works

### DIFF
--- a/src/colors.ts
+++ b/src/colors.ts
@@ -14,7 +14,7 @@ export const colors = {
   hovering: "black",
   lineEndpointColor: "black",
   streetview: "blue",
-  measuringLine: "grey",
+  measuringLine: "cyan",
 };
 
 // For points
--- a/src/colors.ts
+++ b/src/colors.ts
@@ -14,6 +14,7 @@ export const colors = {
   hovering: "black",
   lineEndpointColor: "black",
   streetview: "blue",
+  measuringLine: "grey",
 };
 
 // For points
--- a/src/lib/common/DraggablePin.svelte
+++ b/src/lib/common/DraggablePin.svelte
@@ -15,9 +15,11 @@
       $map.getCanvas().style.cursor = "grabbing";
     });
     marker.on("drag", () => {
-      $map.getCanvas().style.cursor = "inherit";
       markerPosition = marker!.getLngLat();
       markerPositionUpdated();
+    });
+    marker.on("dragend", () => {
+      $map.getCanvas().style.cursor = "inherit";
     });
   });
   onDestroy(() => {
--- a/src/lib/common/DraggablePin.svelte
+++ b/src/lib/common/DraggablePin.svelte
@@ -1,0 +1,28 @@
+<script lang="ts">
+  import { Marker, type LngLat } from "maplibre-gl";
+  import { map } from "stores";
+  import { onDestroy, onMount } from "svelte";
+
+  let marker: Marker | null = null;
+  export let markerPosition: LngLat;
+  export let markerPositionUpdated: () => any;
+
+  onMount(() => {
+    marker = new Marker({ draggable: true })
+      .setLngLat(markerPosition)
+      .addTo($map);
+    marker.on("dragstart", () => {
+      $map.getCanvas().style.cursor = "grabbing";
+    });
+    marker.on("drag", () => {
+      $map.getCanvas().style.cursor = "inherit";
+      markerPosition = marker!.getLngLat();
+      markerPositionUpdated();
+    });
+  });
+  onDestroy(() => {
+    if (marker) {
+      marker.remove();
+    }
+  });
+</script>
--- a/src/lib/common/LineMeasureController.svelte
+++ b/src/lib/common/LineMeasureController.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { lineString, type Feature, type LineString } from "@turf/helpers";
+  import length from "@turf/length";
   import Pin from "lib/critical_entry/Pin.svelte";
   import { SecondaryButton } from "lib/govuk";
   import { LngLat, MapMouseEvent } from "maplibre-gl";
@@ -8,7 +9,7 @@
 
   let isInactive = true;
   let waypoints: any[] = [];
-  let linestring: Feature<LineString> | undefined = undefined;
+  let lineToMeasure: Feature<LineString> | undefined = undefined;
   let isShiftDown = false;
   let nextId = 0;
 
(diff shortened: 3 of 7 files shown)


dabreegster:
Demo at https://acteng.github.io/atip/LineMeasuring/browse.html

dabreegster (commented):
First just UX feedback:

- [ ] Trim the precision of printed lengths, `toFixed(2)` should do it
- [ ] The tool shouldn't be startable at the same time as the street view tool
- [x] The escape key to quit doesn't work
- [ ] If other layers are active and you try to measure things over them, the hover effect clashes.

We don't have to tackle all of these upfront. Looking at code now

src/colors.ts
```diff
@@ -14,6 +14,7 @@ export const colors = {
   hovering: "black",
   lineEndpointColor: "black",
   streetview: "blue",
+  measuringLine: "grey",
```
dabreegster, line 17: I'd actually vote for something a little louder like red or cyan. It's tough to see against satellite:
Screenshot from 2023-09-27 09-42-37

src/pages/CriticalIssueEntry.svelte
dabreegster, line 1: If it's simpler, I would not attempt to share `Pin.svelte` between the two pages at all. They're both a thin wrapper over the maplibre marker, they have different behavior for getting created, it's not even very much code in the first place, and it's liable to all change to use https://svelte-maplibre.vercel.app/examples/draggable_custom_marker anyway

src/lib/critical_entry/Pin.svelte
```diff
@@ -1,39 +1,28 @@
...
+  import { Marker, type LngLat } from "maplibre-gl";
   import { map } from "stores";
   import { onDestroy, onMount } from "svelte";
 
-  export let enableAdding: boolean;
-
   let marker: Marker | null = null;
-  export let markerPosition: LngLat | null = null;
+  export let markerPosition: LngLat;
+  export let markerPositionUpdated: () => any;
```
dabreegster, line 8: Hmm, needing a callback feels like a code smell, since Svelte should do reactivity for us. I guess the problem is somewhere in binding a member of an array during `#each` and not having that update make it look like the entire array has changed? Kind of hard to know without trying something out

src/lib/common/LineMeasureController.svelte
```diff
@@ -0,0 +1,167 @@
...
+          waypointLngLat.lng != waypointToRemove?.lng ||
+          waypointLngLat.lat != waypointToRemove?.lat
+        );
+      });
+    }
+  }
+
+  function enableMeasurement() {
+    isInactive = false;
+    $map.on("click", handleMapClickEvent);
```
dabreegster, line 64: We could just install/uninstall the handlers once in onMount/onDestroy. The first thing it does is check if we're active here; both the check and adding/removing constantly are redundant

src/lib/common/LineMeasureController.svelte
```diff
@@ -0,0 +1,167 @@
...
+
+  function disableMeasurement() {
+    isInactive = true;
+    $map.off("click", handleMapClickEvent);
+    waypoints = [];
+    removeLineFromMap();
+  }
+
+  function keyDown(keyDownEvent: KeyboardEvent) {
+    if (keyDownEvent.key === "Shift") isShiftDown = true;
```
dabreegster, line 75: We're not using one-line style without braces anywhere else in this codebase

src/lib/common/LineMeasureController.svelte
```diff
@@ -0,0 +1,167 @@
...
+        id: layerName,
+        type: "line",
+        source: layerName,
+        paint: {
+          "line-color": colors.measuringLine,
+          "line-width": 5,
+        },
+      });
+    } else {
+      // @ts-ignore setData not happy for some reason
```
dabreegster, line 121: Grep the rest of the codebase for reference, there's a typecast

src/lib/common/LineMeasureController.svelte
```diff
@@ -0,0 +1,167 @@
...
+    <p>Length: {lineToMeasure.properties?.length}m</p>
+  {/if}
+  <CollapsibleCard label="Help">
+    <ul>
+      <li>
+        <b>Click</b>
+        on the map to place a new waypoint
+      </li>
+      <li>
+        <b>Click & Drag</b>
```
dabreegster, line 147: "and drag"

btw "&" isn't supposed to be used directly in HTML, since it's used for special codes. &amp;

src/lib/common/LineMeasureController.svelte
```diff
@@ -0,0 +1,167 @@
...
+      return [waypoint.lngLat.lng, waypoint.lngLat.lat];
+    });
+    lineToMeasure = lineString(waypointsAsPosition, {});
+    if (lineToMeasure.properties)
+      lineToMeasure.properties.length = length(lineToMeasure) * 1000;
+    updateLinestring();
+  }
+
+  function removeLineFromMap() {
+    if ($map.getSource(layerName)) {
```
dabreegster, line 98: Constantly adding / removing the source and layer is very confusing to keep track of... the style we use in the rest of the codebase is to declare the source/layer once using the `overwrite` methods, and then `setData`

src/lib/critical_entry/Pin.svelte
Pete-Y-CS: I agree. The issue was that I didn't know how to make the lines redraw  on the map live without it.

Pete-Y-CS:
Resolved the UX issues of esc not used, too much precision and colour of line. 

I think the clash of the tools should be part of a bigger piece of work about how we organise our UI/UX for different layers and tools on the browse page

src/lib/common/DraggablePin.svelte
```diff
@@ -0,0 +1,28 @@
...
+
+  onMount(() => {
+    marker = new Marker({ draggable: true })
+      .setLngLat(markerPosition)
+      .addTo($map);
+    marker.on("dragstart", () => {
+      $map.getCanvas().style.cursor = "grabbing";
+    });
+    marker.on("drag", () => {
+      $map.getCanvas().style.cursor = "inherit";
```
dabreegster, line 18: Ah, actually we need to do this on `dragend` to reset it. `drag` is continuously called during drag

dabreegster (approved):
Just one change, and looks good! We can iterate on it later; this is a perfectly usable start. Awesome!

PR #361, opened 2023-10-03
Use new route-snappers, revamping how auto-named routes work

dabreegster:
This is a followup to #352. We're going to remove the `route-info` WASM library, which only has one useful working feature: auto-naming routes based on the two intersection endpoints. This PR changes that feature to instead use a new version of the route-snapper library. It also slightly improves the UX, by populating the name automatically when a route is first made. Users can still override the name, or later auto-fill it (to correct things after a split, for instance).

The benefits to dismantling the route-info service are:

- Page load performance: almost half the amount of data to load, because now we only load a route snapper graph, not a second file. The snapper file has grown slightly now that it has names, but it's a small increase when gzipped, and the overall change is way less to download.
- Code simplicity: we don't have two WASM libraries doing confusingly similar things
- Deployment performance: now we don't need to compile any WASM packages in this repo, so CI is going to get way faster.

On that last point, I'm going to remove the old route-info service next. It has to be separate, due to how tricky it is to upload the GH workflow script and handle all outstanding branches.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "js-cookie": "^3.0.5",
         "maplibre-gl": "^3.3.1",
         "read-excel-file": "^5.6.1",
-        "route-snapper": "^0.1.15",
+        "route-snapper": "^0.2.1",
         "svelte": "^4.0.0",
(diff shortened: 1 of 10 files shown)


Robinlovelace:
Awesome, great to see these quality of life improvements rolling in.

Issue #362, opened 2023-10-05
Validations in the sketcher


dabreegster:
Goal: Let somebody upload an arbitrary GeoJSON file into the sketch tool, then give them an easy way to fix any problems that make the file non-compliant with the ATIP schema.

Edge cases:

- [ ] Handle MultiLineString and other odd geometry types

PR #363, opened 2023-10-05
Guide users to fix problems in the sketch tool

dabreegster:
For #362. This effectively adds support for people to load arbitrary GeoJSON files into the sketch tool. Any features that don't match our expected schema (or that have other problems we can auto-detect) show a clear warning, and there's a summary of how many problems the file has. There's also handling for removing extra properties.

https://github.com/acteng/atip/assets/1664407/bc304e69-348c-4f66-a2f8-68de3c88cc86


src/lib/sidebar/scheme_data.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+  return `Untitled ${noun}`;
+}
+
+export function interventionWarning(
+  schema: Schema,
+  feature: FeatureUnion
+): string | null {
+  // Only worry about one schema for now.
```
dabreegster, line 69: In the future, these could come from some kind of external JSON schema validation library maybe.

We could also add more cases to catch polygons marked as route, and other problems found recently in ATF data.

dabreegster:
As #362 notes, another thing we might want to do is automatically split Multi LineStrings/Polygons. We've seen em for real in submitted data, and they're not the nicest thing particularly to support.

Issue #288, opened 2023-07-24
Add a geocoder to the browse page


dabreegster:
Typing in an address or post code to navigate a map is very handy. https://maplibre.org/maplibre-gl-js/docs/examples/geocoder/ is a starter. We need to pick a geocoding service and figure out pricing.

robinlovelace-ate:
Yes and to the editing pages also please, was meaning to open an issue on this a while back. Great thinking Dustin!

dabreegster:
Possible sources:
- Nominatim: free but extreme quality problems from my tests
- OpenCage
- OS Places API
- MapTiler
- ArcGIS
- https://geocode.earth/

dabreegster:
Been working for a while, just closing old issue

PR #366, opened 2023-10-11
Add a geocoder using MapTiler

dabreegster:
#288 

I started with maplibre-gl-geocoder and Nominatim API, following https://maplibre.org/maplibre-gl-js/docs/examples/geocoder/, but the plugin's UX was clunky and results from Nominatim are not great. Since we're using MapTiler for some basemaps anyway and already have an API key, I decided to try their plugin + API instead, and I think it works much more smoothly.

Try at https://acteng.github.io/atip/geocoder_maptiler/browse.html. It's still imperfect, but I think this is enough to start.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "atip",
       "version": "0.0.0",
       "dependencies": {
+        "@maptiler/geocoding-control": "^0.0.99",
         "@turf/bbox": "^6.5.0",
         "@turf/boolean-contains": "^6.5.0",
         "@turf/boolean-point-in-polygon": "^6.5.0",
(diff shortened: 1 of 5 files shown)


robinlovelace-ate:
Works great. Takes me there!

image

Source (your URL isn't quite right above Dustin): https://acteng.github.io/atip/geocoder_maptiler/browse.html?style=dataviz#13.38/53.82764/-1.54093

robinlovelace-ate (approved):

robinlovelace-ate:
I guess it would be easy to add this to other tools including sketch. Not that we should but could be handy in other tools at some point.

PR #367, opened 2023-10-11
A few UX fixes

dabreegster:
- When there are overlapping layers, only let the topmost one have interactions. (I accidentally lost this behavior during the svelte-maplibre shuffle, when switching from my fork to the upstream)
- Add the geocoder to critical entry and sketch pages
- Add the line measure tool to the critical entry page (but it partly breaks / conflicts with the main tool there)

Demo at https://acteng.github.io/atip/topmost/browse.html


robinlovelace-ate:
Will give it a spin.

robinlovelace-ate:
Tested with 2 layers and, indeed, only got popup on top one, as shown below:

image

Issue #248, opened 2023-06-25
Rethink edit attributes vs edit geometry mode


dabreegster:
#108 introduced two different modes, edit attributes (for stuff in the sidebar) and edit geometry (for the geospatial representation). The motivation was that the form was going to soon outgrow the sidebar and become a modal. A big modal would cover up the map, making it impossible to edit the geometry too.

I'd like to revisit that design choice, after the UX session with Kelin. I think one "Edit" mode (or "select" or some other name) is fine. Clicking or selecting an object would activate the appropriate controls on the map, but ALSO open up the form in the sidebar. If our forms ever outgrow the sidebar, we simply add a prominent button there to pop-out into a full modal and cover part/all of the map. If a user is trying to fill out this larger form for many objects, we can even have some kind of "sticky" behavior about showing that bigger modal by default or not.

Thoughts?

robinlovelace-ate:
I don't fully understand the suggestion but my thinking is that, given that the attribute editing form will grow, it makes sense to keep geometry editing and attribute editing functionality separate.

dabreegster:
Try a quick prototype at https://acteng.github.io/atip/merge_edit/scheme.html?authority=Adur (give it a few mins to deploy, and keep in mind there are lots of bugs left in this -- it's just a quick example)

https://github.com/acteng/atip/assets/1664407/eea6ece2-ad51-4985-a29f-f17b2e06ad37

Even in the current approach with a separate "Edit attributes", we're only using the sidebar. The idea with the new approach is that we'd add a new button to show a full-screen form:
![Screenshot from 2023-06-26 13-37-29](https://github.com/acteng/atip/assets/1664407/c4bad9f1-51e5-4986-8010-d56d4e7f277e)

I've seen a few people get confused by the two separate edit modes, and when using ATIP myself, it's a bit frustrating to switch back and forth.

robinlovelace-ate:
Amazing progress Dustin! I'm in strategy mode but will aim to test soon.

dabreegster:
Thanks Stuart! You've got me thinking about a few other old ideas for simplifying the UX flow... One note -- we don't have the ability (or any intended use case for this) to select multiple features at a time today.

Thinking through the UX again... this idea of a list and a focus view really sounds solid:

**List view**
- The left sidebar lists all features and has the file management / scheme-wide forms (possibly as a modal, if they grow to include more)
- The map shows all features
- Clicking one feature either from the sidebar or map switches into the focus view
[...]

PR #370, opened 2023-10-16
Overhaul sketch page UX

dabreegster:
#248, #228, #134

I'll describe the changes when I mark this PR as ready for review. For now, please give feedback about https://acteng.github.io/atip/modal_shift/scheme.html?authority=Adur. Robin doesn't like the new top toolbar.

My TODOs:

- [x] Make sure constructors aren't happening before map is ready
- [x] Move controls to left sidebar: streetview, route's shift toggle
- [x] Restyle the progress bar for the horizontal format
- [x] Rewrite instructions
- [x] Check with other schemas
- [x] Rewrite playwright tests

Bigger followup ideas:

- [ ] Display key shortcut icons next to buttons?
- [ ] Modal for scheme details (especially since this'll grow)
- [ ] Modal for "new object", so we can guide people on matching "school street" to "linestring" for example
- [ ] Merge edit form + geometry modes
- [ ] Maybe put the basemap picker on the map, not sidebar
- [ ] Maybe put "split route" into edit form/geometry mode


Pete-Y-CS:
Some UI niggles:
image
which intervention has a problem?


image
Filling in a form on the left and having to confirm in the top bar is a little confusing (keyboard shortcut would be nice too)


more significant bug:
image
It seems that the route snapper finishing loading doesn't enable route and snapped polygon: these have remained disabled for me until I input e.g. a freehand polygon, and once I finish doing that new route and snapped polygon are enabled

Pete-Y-CS:
Another thought, well two thoughts, well three thoughts:
1. a lot of LCWIPs have routes sharing some sections, depending on how we represent might be good to allow the user to colour code paths
2. some LCWIPs include existing, to be upgraded, fully new in their LCWIP, I suppose we want to collect all 3 of these categories? The alternative would be upgraded and fully new only I suppose
3. How do we want to represent shared paths? Here's an example from Peterborough where there are 3 conceptually distinct routes stacked together for most of their lengths:
image

Pete-Y-CS:
This isn't really about the rework but I'm gonna add here anyway: I think a useful feature would be freehand when editing an existing node on a path using edit geometry i.e. hold shift  and the link to the previous and next node become a straight line and you can place the edited node whereever

dabreegster:
Thanks for all the feedback! Starting to look into some of these issues.

> I think a useful feature would be freehand when editing an existing node on a path using edit geometry i.e. hold shift and the link to the previous and next node become a straight line and you can place the edited node whereever

Not sure I understood -- do you mean converting an existing node from snapped to freehand, probably in the middle of the route somewhere? This indeed seems useful. Maybe it'd be helpful to style snapped and freehand nodes differently, to make it even more clear which is which?

> Filling in a form on the left and having to confirm in the top bar is a little confusing (keyboard shortcut would be nice too)

Agreed. The intention was for the main navigation / state-changing buttons to consistently be in the top bar. But maybe this is confusing. What if we instead just use the left sidebar for all of the controls? The current top bar will **only** be present in the neutral list mode, to give room for all of the "new" buttons. Or we could even change that -- go back to the right sidebar (and have it vanish once you're in any other mode) or have one "new" button in the left sidebar or something else. Suggestions / mockups welcome!

Pete-Y-CS:
Fully removing the top bar whilst filling in the form at least feels worth trying. 

>Not sure I understood -- do you mean converting an existing node from snapped to freehand, probably in the middle of the route somewhere? This indeed seems useful. Maybe it'd be helpful to style snapped and freehand nodes differently, to make it even more clear which is which?

Yeah so you start dragging a node, press shift, and suddenly it becomes freehand, exactly. Styling them differently seems like a nice touch too

dabreegster:
> Fully removing the top bar whilst filling in the form at least feels worth trying.

Done. I think this looks better. Should we do it for all modes?

src/lib/draw/route/RouteControls.svelte
```diff
@@ -1,59 +1,75 @@
...
   // TODO When editing, we should save in the route and use the previous value
-  $: routeTool.setRouteConfig({
+  $: $routeTool!.setRouteConfig({
     avoid_doubling_back: $userSettings.avoidDoublingBack,
     extend_route: extendRoute,
   });
 
-  // TODO Show if shift is held or not
-  // TODO Disable finish when the route is invalid
+  // TODO It'd be nice to wire up reactivity to RouteTool directly.
```
dabreegster, line 15: In Svelte 5, this will be possible. In the meantime, I am thinking about rewriting the 3 .ts tools as components, to make use of reactivity directly and to not dodge around svelte-maplibre

src/lib/draw/route/RouteSnapperLoader.svelte
dabreegster, line 1: Extracted loading logic here

src/lib/sidebar/About.svelte
```diff
@@ -21,11 +21,18 @@
           Dustin Carlino
         </ExternalLink>: lead developer, from The Alan Turing Institute
       </li>
+      <li>
```
dabreegster, line 24: Added you here -- long overdue!

src/lib/sidebar/EditForm.svelte
```diff
@@ -0,0 +1,111 @@
...
+      // Let the delete key work in forms
+      if (tag == "INPUT" || tag == "TEXTAREA") {
+        return;
+      }
+      e.stopPropagation();
+
+      deleteIntervention(id);
+    }
+
+    if (e.key == "e") {
```
dabreegster, line 74: New hotkey to edit geometry!

src/lib/sidebar/Instructions.svelte
dabreegster, line 1: The deleted sections aren't really relevant anymore... the UI should be way more self-explanatory now?

src/lib/sidebar/LeftSidebar.svelte
```diff
@@ -0,0 +1,84 @@
...
+  import RouteSnapperLoader from "../draw/route/RouteSnapperLoader.svelte";
+  import SnapPolygonMode from "../draw/snap_polygon/SnapPolygonMode.svelte";
+  import StreetViewMode from "../draw/StreetViewMode.svelte";
+  import EditForm from "./EditForm.svelte";
+  import ListMode from "./ListMode.svelte";
+
+  export let schema: Schema;
+  export let routeSnapperUrl: string;
+
+  $: if ($map && !$pointTool) {
```
dabreegster, line 22: We have to wait for the map to be loaded to create these

src/lib/sidebar/LeftSidebar.svelte
```diff
@@ -0,0 +1,84 @@
...
+  <h2>New route</h2>
+  <RouteMode />
+{:else if $mode.mode == "new-freehand-polygon"}
+  <h2>New polygon (freehand)</h2>
+  <PolygonMode />
+{:else if $mode.mode == "new-snapped-polygon"}
+  <h2>New polygon (snapped)</h2>
+  <SnapPolygonMode />
+{:else if $mode.mode == "split-route"}
+  <h2>Split route</h2>
```
dabreegster, line 61: See comments in SplitRouteMode.svelte. This one is structured a little weirdly because it has to be nested underneath Map, not in this component.

dabreegster (commented):
Alright, this is now ready for review! IMO, all that's left to merge this PR is Playwright tests. Then we can tackle smaller followups

scheme.html
```diff
@@ -6,7 +6,7 @@
       name="viewport"
       content="initial-scale=1,maximum-scale=1,user-scalable=no"
     />
-    <title>ATIP v2 Scheme Entry</title>
+    <title>ATIP v2 Scheme Sketcher</title>
```
Pete-Y-CS, line 9: Somehow this change feels quite momentous

dabreegster: Just something I noticed. Do we have a better name? >_<

Pete-Y-CS: I don't think so, I just mean that we've been talking for a while about various issues with naming and so on, and rebranding to scheme sketcher. Now it's here!

dabreegster:
Alright, all the tests fixed up -- I think this MVP is ready!

src/pages/ChooseArea.svelte
```diff
@@ -111,7 +111,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-one-half left">
-    <h1 class="govuk-heading-l">Welcome to ATIP v2</h1>
+    <h1 class="govuk-heading-l">Welcome to ATIP Scheme Sketcher v2</h1>
```
Pete-Y-CS, line 114: 🔥

Pete-Y-CS (approved):
Nothing  that holds up merging but a few instances where I've commented because I might do it differently

src/lib/draw/EditGeometryMode.svelte
```diff
@@ -0,0 +1,157 @@
...
+
+    if (feature.geometry.type == "LineString") {
+      $routeTool?.editExistingRoute(feature as Feature<LineString>);
+      $routeTool?.addEventListenerSuccess(onSuccess);
+      $routeTool?.addEventListenerUpdated(onUpdate);
+      $routeTool?.addEventListenerFailure(onFailure);
+      controls = "route";
+    } else if (feature.geometry.type == "Polygon") {
+      if (feature.properties.waypoints) {
+        $routeTool?.editExistingArea(feature as Feature<Polygon>);
```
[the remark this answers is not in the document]
dabreegster: Hmm, the temptation to refactor a bit is hampered slightly by the point tool being an exception to not handling updates. Maybe also the tools could have a single method `setEventListeners(onSuccess, onUpdate, onFailure)`, since there's now not actually any use for them having multiple callbacks at a time. I'll table this for the moment since the explicit logic is working and isolated here, but happy to revisit

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -132,53 +122,48 @@
...
-  <p>
-    Finish drawing on the map and/or select "Edit attributes" to use these
-    options.
-  </p>
+    <div>
+      <WarningButton
+        on:click={() => (displayClearAllConfirmation = true)}
+        disabled={$gjScheme.features.length == 0}
+      >
+        Start new scheme
```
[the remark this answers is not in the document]
dabreegster: Leaving as TODO, because I want to think through what the "scheme-wide" LCWIP fields will be and make the UI nicer. There's 3 buttons here now that are logically "New file", "Open file", and "Save file", but they don't really look that way now. If we move them all out, the left sidebar gets huge again and you can't see the list:
Screenshot from 2023-10-21 13-59-38
and they're also not things people should really think about much. (Arguably that's true  for about / instructions / the title too)

Issue #345, opened 2023-09-15
Regenerate route snappers with fixed boundaries


dabreegster:
The input boundary tries to exclude the Thames, so the route snapper dutifully avoids it:
![Screenshot from 2023-09-15 22-49-31](https://github.com/acteng/atip/assets/1664407/2a36734c-e45a-48bc-8a6d-304f40e9af4d)

dabreegster:
Status:

- New snappers are regenerated and uploaded (as v2.3) for everywhere except TA_West of England Combined Authority
- https://github.com/acteng/atip/tree/new_boundaries uses the new boundaries, and handles the new naming scheme for authorities (`${level}_${authorityName}`)

TODO:

- [ ] Fix the one broken area. (Short-term fix in osm2streets, but longer term, I want to simplify how we build the files from OSM data directly, and also add support for pbf import for faster import)
- [ ] Update `all_scheme_data.gj`with the new authority names, which'll also fix problems with the browse page, where the original authority name was lost (Ideally we start a new internal git repo to manage this)

dabreegster:
Ah, and also, 4 areas in the new boundaries use MultiPolygons. This breaks two different things:
- The route snapper will just use the first polygon. It needs one contiguous boundary to use. Right now it just uses the first polygon in that list. This is OK for most of the areas -- Devon is one example where there's a tiny island associated with it. But the Isles of Scilly break, because there's legitimately a few different islands there.
- I'm using a Turf library to match GJ schemes to the authority boundary containing them. It doesn't handle MultiPolygons either. I'm going to do the same "use the first polygon" hack.

It might be nicer to simplify upstream and fix `authorities.geojson` to handle this more carefully. But I don't want to do any more work on that boundary file until we can actually reproducibly recreate it from the steps in data-prep.

PR #375, opened 2023-10-23
Update route snappers and handle new authority names

dabreegster:
#345 

This PR makes the authority name more specific, including the LAD_ or TA_ prefix. And since we're changing boundary names, there's new logic to match a GJ scheme to the smallest boundary fully containing it. That also helps backfill previously broken data in the browse page.

After merging this PR, manual deployment steps will be:

- [x] Uploading a new `all_scheme_data.geojson` to Sharepoint where people will find it
- [ ] Regenerating and uploading `random_schemes.geojson` to the test bucket
- [x] Manually deploying to GCP (since we still don't have service account automation)


src/scripts/assign_authority.ts
dabreegster, line 1: Oh yeah, and I want to start that internal repo for managing raw input data and centralizing all the data-cleaning scripts. Previous one is at https://github.com/acteng/atip_analysis/blob/main/js/index.js for reference.

Pete-Y-CS (approved):

dabreegster:
I've updated `all_schemes_data.geojson` in the Sharepoint folder. At your leisure, could you sync past this PR, `npm run generate-random-schemes`, and replace the file in GCS? Thanks!

PR #376, opened 2023-10-25
Update route-snapper

dabreegster:
Brings in changes from https://github.com/dabreegster/route_snapper/blob/main/CHANGES.md#022.

https://github.com/acteng/atip/assets/1664407/490bed16-37e8-42d4-ae16-1546897fbaa4

Demo at https://acteng.github.io/atip/route_snapper_update/scheme.html?authority=Adur.


src/lib/draw/route/route_tool.ts
```diff
@@ -343,9 +334,10 @@ export class RouteTool {
   }
 
   private redraw() {
-    (this.map.getSource(source) as GeoJSONSource).setData(
-      JSON.parse(this.inner.renderGeojson())
-    );
+    let gj = JSON.parse(this.inner.renderGeojson());
+    (this.map.getSource(source) as GeoJSONSource).setData(gj);
+    this.map.getCanvas().style.cursor = gj.cursor;
+    this.snapMode.set(gj.snap_mode);
```
dabreegster, line 340: Since we're in a .ts file, we can't change  a variable and have the Controls.svelte update easily. Using a store to get around that currently.

dabreegster:
Pete, could you try using this for that example case and see if it feels better?

robinlovelace-ate:
Minor aesthetic niggle, really not a fan of this empty space that is too big not to be noticed but too small to be useful:

image

robinlovelace-ate:
Either full banner (that's foldable?) or keep in a left or right panel would be neater, like (hate to admit it's good) Google Maps : )

robinlovelace-ate:
Probably not foldable as it has one job.

robinlovelace-ate:
In terms of extra space on right: love it! Minor tweak can overcome my niggle above, make it full width is first thought.

dabreegster:
Thanks for the feedback! Specifically can you create a new route and make something that involves both snapping and freehand points? That's the big change here, but will also incorporate the top bar design

src/lib/draw/route/RouteControls.svelte
```diff
@@ -11,53 +11,36 @@
...
+{#if $routeToolSnapMode}
+  <p style="background: red; padding 8px;">
+    Snapping to existing roads. Press <b>s</b>
     to draw anywhere
   </p>
 {:else}
-  <p>
-    Drawing points anywhere. Release <b>Shift</b>
+  <p style="background: blue; padding: 8px;">
+    Drawing points anywhere. Release <b>s</b>
```
Pete-Y-CS, line 23: Two small notes: it's really `press s to switch back to snapped mode` since you're not holding s down for this.

The other one I'm a little less certain of but the behaviour when you switch back to snapped isn't entirely obvious (although I can't think it would better in the more intuitive way). Difficult to explain in text maybe have a quick call in a mo?

dabreegster: Fixed the text and coloring lines now to be very clear

dabreegster:
PTAL, demo is now at https://acteng.github.io/atip/route_snapper_update/scheme.html?authority=LAD_Adur.

- Always start in snapped mode
- Fixed up the red/blue text box a bit
- Coloring the lines too. You can wind up with this for example:
Screenshot from 2023-10-26 11-03-12

When the black dots are dense on a red line, it's maybe a bit hard to see. It looked worse to me to put the line on top. Styling suggestions welcome!

Pete-Y-CS:
Changes look great!

PR #377, opened 2023-10-25
Add an alpha banner to most pages

dabreegster:
Demos:
https://acteng.github.io/atip/alpha/browse.html
https://acteng.github.io/atip/alpha/critical_entry.html
https://acteng.github.io/atip/alpha/scheme.html?authority=Adur

### DIFF
--- a/src/lib/browse/AppVersion.svelte
+++ b/src/lib/browse/AppVersion.svelte
@@ -14,11 +14,5 @@
       </p>
     {/if}
     <p>Note the ATF scheme data has not been validated.</p>
-    <p>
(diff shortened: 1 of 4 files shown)


Pete-Y-CS:
I did wonder if we wanted to do it depending on environment (e.g. when deployed on GCP), but easier just to leave it in everywhere.

Pete-Y-CS (approved):

dabreegster:
Happy to drop it from the GH deployments if we get that feedback. But in the short term, our partner will likely be using that version, so good for it to be on the sketch page at least

Issue #368, opened 2023-10-12
DEFRA layers


dabreegster:
From https://uk-air.defra.gov.uk/data/gis-mapping/: NOx, PM2.5, PM10. Not sure about background/roadside yet, with/without sea salt.

http://www.extrium.co.uk/noiseviewer.html  Noise pollution from traffic, not sure exactly which layer, or about the data availability

dabreegster:
Possible links for the noise: https://environment.data.gov.uk/dataset/fd1c6327-ad77-42ae-a761-7c6a0866523d, https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/902825/strategic-noise-mapping-round3.pdf. We probably want LAeq,16h.

Both roadside + background for pollution. Look into ways to use roadside and fallback to background.

dabreegster:
Started in https://github.com/acteng/atip/tree/defra. But it's better to do this properly and first upstream nice raster support in https://github.com/dimfeld/svelte-maplibre/pull/92

dabreegster:
Closing this in favor of #490. We want to find raw data and make our own tiles, preferably

Issue #359, opened 2023-10-02
Cleanup after svelte-maplibre refactor


dabreegster:
- [ ] Plumb paint properties in `InterventionColorSelector` instead of setting directly
- [ ] Rethink behavior of multiple overlapping features, especially for the LPA layer
- [ ] Rethink the route and polygon `.ts` tools managing sources/layers, and maybe cleanup the old maplibre util functions
- [ ] Stop using the `gjScheme` store in the browse page
- [ ] Use svelte-maplibre for the critical entry draggable marker (only geocode on dragend)
- [ ] Use svelte-maplibre for the line measurement tool
- [ ] Refactor something for areas-with-outlines
- [ ] Update dev docs (some HMR guidance refers to old methods)
- [ ] Consider new UX patterns, like showing a popup both on hover, and then making it "sticky" when clicked
- [ ] Revisit `interactiveMapLayersEnabled` and make sure browse layers are inactive when other tools are active. Try an invisible layer that covers everything?
- [ ] Improve critical clustering (see below)

Critical clustering has changed and feels a bit more buggy to me. I can't figure out why yet; svelte-maplibre seems to plumb through the cluster options just fine. Before:
![Screenshot from 2023-10-02 12-00-59](https://github.com/acteng/atip/assets/1664407/6ec4fff3-6afb-46a8-a3ba-3b70662f0d86)
After:
![Screenshot from 2023-10-02 12-02-16](https://github.com/acteng/atip/assets/1664407/75e18149-a784-44f0-8f3d-85beb1bc7600)
Note some of the text counts don't show up.

PR #378, opened 2023-10-25
Add pollution layers

dabreegster:
#368. Demo at https://acteng.github.io/atip/defra/browse.html.

It would be much nicer to show a value by the mouse as you move around, but these pre-built raster layers served through WMS don't easily support it. We could try out the `GetFeatureInfo` API in the future (maybe on-click; as you hover won't be fast enough). Or we could process the raw data ourselves and produce raster tiles in some format that's easier to query.

It would also be nice to match the NOx metric used by inspections: >40, 40-32, <32. But again with prebuilt raster data, we can't do much.

Note this PR contains an attempt at the noise layer from http://www.extrium.co.uk/noiseviewer.html. It's disabled for two reasons:
1) I'm not sure they mean for it to be used publicly, even though the WMS server happily serves content without CORS problems.
2) They only have an HTTP endpoint. It works with `npm run dev`, but not on GH pages or GCS because of https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,6 @@
     "read-excel-file": "^5.6.1",
     "route-snapper": "^0.2.1",
     "svelte": "^4.0.0",
-    "svelte-maplibre": "^0.5.0"
+    "svelte-maplibre": "github:dabreegster/svelte-maplibre#changes_for_atip"
   }
 }
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "govuk-frontend": "^4.6.0",
     "humanize-string": "^3.0.0",
     "js-cookie": "^3.0.5",
-    "maplibre-gl": "^3.3.1",
+    "maplibre-gl": "^3.5.1",
     "read-excel-file": "^5.6.1",
     "route-snapper": "^0.2.1",
     "svelte": "^4.0.0",
--- a/src/lib/browse/layers/areas/Pollution.svelte
+++ b/src/lib/browse/layers/areas/Pollution.svelte
@@ -1,0 +1,56 @@
+<script lang="ts">
+  import { ExternalLink, HelpButton, Popup } from "lib/common";
+  import { Checkbox } from "lib/govuk";
+  import { layerId } from "lib/maplibre";
+  import { map } from "stores";
+  import { onDestroy, onMount } from "svelte";
+  import OsOglLicense from "../OsOglLicense.svelte";
+
+  let name = "pollution";
+
+  let show = false;
+
+  // legend
+  //"https://ukair.maps.rcdo.co.uk/ukairserver/services/aq_amb_2021/NOx_viridis/MapServer/WMSServer?request=GetLegendGraphic%26version=1.3.0%26format=image/png%26layer=21"
+
+  onMount(() => {
+    $map.addSource(name, {
+      type: "raster",
+      tiles: [
+        "https://ukair.maps.rcdo.co.uk/ukairserver/services/aq_amb_2021/NOx_viridis/MapServer/WMSServer?request=GetMap&version=1.3.0&format=image/png&crs=EPSG:3857&width=256&height=256&styles=&bbox={bbox-epsg-3857}&layers=21",
+      ],
+      tileSize: 256,
+    });
+    $map.addLayer({
+      id: "pollution-layer",
+      type: "raster",
+      source: name,
+      paint: {
+        "raster-opacity": 0.5,
+      },
+      layout: {
+        visibility: "none",
+      },
+    });
+  });
+  onDestroy(() => {
+    $map?.removeLayer("pollution-layer");
+    $map?.removeSource(name);
+  });
+
+  $: $map.setLayoutProperty(
+    "pollution-layer",
+    "visibility",
+    show ? "visible" : "none"
+  );
+</script>
+
+<Checkbox id={name} bind:checked={show}>
+  Pollution
+  <span slot="right">
+    <HelpButton>
+      <p>TODO</p>
+      <OsOglLicense />
+    </HelpButton>
+  </span>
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -48,10 +48,10 @@
     "govuk-frontend": "^4.6.0",
     "humanize-string": "^3.0.0",
     "js-cookie": "^3.0.5",
-    "maplibre-gl": "^3.3.1",
+    "maplibre-gl": "^3.5.1",
```
dabreegster, line 51: Needed for https://github.com/maplibre/maplibre-gl-js/blob/main/CHANGELOG.md#350

package.json
```diff
@@ -48,10 +48,10 @@
...
     "read-excel-file": "^5.6.1",
     "route-snapper": "^0.2.1",
     "svelte": "^4.0.0",
-    "svelte-maplibre": "^0.5.0"
+    "svelte-maplibre": "github:dabreegster/svelte-maplibre#changes_for_atip"
```
dabreegster, line 55: Switching back to my fork because https://github.com/dimfeld/svelte-maplibre/pull/92 won't be reviewed for about a week while maintainer is on holiday.

Also note this updates svelte-maplibre and fixes some HMR issues! #359

src/lib/browse/layers/areas/Pollution.svelte
```diff
@@ -0,0 +1,144 @@
...
+    </label>
+  </div>
+
+  {#if pollutant == "Noise"}
+    <SequentialLegend
+      colorScale={["#FF6600", "#FF3333", "#990033", "#AD9AD6", "#0000E0"]}
+      limits={["55", "60", "65", "70", "75", ">"]}
+    />
+  {:else}
+    <img
```
dabreegster, line 126: It's near illegible, unfortunately. https://docs.geoserver.org/main/en/user/services/wms/get_legend_graphic/index.html is supposed to let us customize font size and DPI and stuff, but none of it seems to have an effect here. Maybe only GeoServer implements these options and DEFRA is using something else?

Pete-Y-CS: Also not accessible for e.g. screen reader. Although it's for understanding visual information anyway so it might not be as much of a concern...

src/lib/browse/layers/areas/Pollution.svelte
```diff
@@ -0,0 +1,145 @@
...
+
+  let show = false;
+  let opacity = 50;
+  let pollutant = "PM25_viridis";
+
+  // URLs and layers found from https://uk-air.defra.gov.uk/data/wms-services
+  // and QGIS, or http://www.extrium.co.uk/noiseviewer.html for the noise layer
+  $: info = {
+    NOx_viridis: ["22", "Data for 2022"],
+    PM25_viridis: ["21", "Data for 2022"],
```
Pete-Y-CS, line 17: is this supposed to say for 2021?

src/lib/browse/layers/areas/Pollution.svelte
```diff
@@ -0,0 +1,145 @@
...
+    PM10_viridis: ["22", "Data for 2022"],
+    NOxRoads_viridis: ["22", "Data for 2022"],
+    PM25Roads_viridis: ["14", "Data for 2022"],
```
Pete-Y-CS, line 20: likewise with 2014?

src/lib/browse/layers/areas/Pollution.svelte
```diff
@@ -0,0 +1,145 @@
...
+  <Select
+    label="Pollutant"
+    id="pollutant"
+    choices={[
+      ["PM25_viridis", "Background PM2.5"],
+      ["PM10_viridis", "Background PM10"],
+      ["NOx_viridis", "Background NOx"],
+      ["PM25Roads_viridis", "Roadside PM2.5"],
+      ["PM10Roads_viridis", "Roadside PM10"],
+      ["NOxRoads_viridis", "Roadside NOx"],
```
Pete-Y-CS, line 108: Does this need `["Noise", "Noises],` adding?

Pete-Y-CS (changes requested):

src/lib/browse/layers/areas/Pollution.svelte
dabreegster: Note this PR contains an attempt at the noise layer from http://www.extrium.co.uk/noiseviewer.html. It's disabled for two reasons:

    I'm not sure they mean for it to be used publicly, even though the WMS server happily serves content without CORS problems.
    They only have an HTTP endpoint. It works with npm run dev, but not on GH pages or GCS because of https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content.

src/lib/browse/layers/areas/Pollution.svelte
dabreegster: Based on the very end of https://ukair.maps.rcdo.co.uk/ukairserver/services/aq_amb_2022/PM25_viridis/MapServer/WMSServer?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities and QGIS, no. The layer IDs are not consistent
Screenshot from 2023-10-26 10-01-42

src/lib/browse/layers/areas/Pollution.svelte
dabreegster: Double checked https://ukair.maps.rcdo.co.uk/ukairserver/services/aq_amb_2022/PM25Roads_viridis/MapServer/WMSServer?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities, 14 is right

Pete-Y-CS (approved):

PR #381, opened 2023-11-01
Plumb along schema in URL

dabreegster:
Two quick fixes:
1) If you go the homepage with ?schema=foo, then we preserve that when you click an area. Otherwise our partner will have to manually add `schema=pipeline` in the near-future
2) Stop using .txt for downloaded files. This was something we did for better compatibility with SmartSurvey. Since we're not using a  new version of the sketch tool anytime soon, let's use .geojson, so people understand what the files are and they work with normal GIS tools

### DIFF
--- a/tests/homepage.spec.ts
+++ b/tests/homepage.spec.ts
@@ -1,12 +1,5 @@
 import { expect, test } from "@playwright/test";
 
-test("front page has welcome text", async ({ page }) => {
-  await page.goto("/");
-  await expect(
-    page.getByRole("heading", { name: "Welcome to ATIP Scheme Sketcher v2" })
-  ).toBeVisible();
-});
-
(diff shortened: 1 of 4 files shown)


tests/homepage.spec.ts
```diff
@@ -1,12 +1,5 @@
 import { expect, test } from "@playwright/test";
 
-test("front page has welcome text", async ({ page }) => {
```
dabreegster, line 3: I think this was the original playwright test just to get started. It's not too useful now :)

Pete-Y-CS (changes requested):
image
Can we apply this to here too (`change area` button)

dabreegster:
Ooh good catch, thanks! OK to merge?

Pete-Y-CS (approved):

PR #385, opened 2023-11-06
Superscheme tidy up

Pete-Y-CS:
Change subscheme to scheme when giving default name to "sub"schemes
Don't require user to expand 'sub'scheme to see its interventions

### DIFF
--- a/src/lib/sidebar/scheme_data.ts
+++ b/src/lib/sidebar/scheme_data.ts
@@ -50,7 +50,7 @@ export function backfillSuperscheme(
   if (!json.scheme_name) {
     json.scheme_name = authorityName
       ? `${authorityName} LCWIP`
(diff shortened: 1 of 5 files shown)


src/lib/sidebar/scheme_data.ts
```diff
@@ -50,15 +50,15 @@ export function backfillSuperscheme(
   if (!json.scheme_name) {
     json.scheme_name = authorityName
       ? `${authorityName} LCWIP`
-      : "Untitled Supserscheme";
+      : "Unnamed Supserscheme";
```
dabreegster, line 53: Typo! Superscheme

dabreegster (approved):
New behavior is much more understandable to me, thanks!

PR #386, opened 2023-11-08
Default to the pipeline schema on that branch

dabreegster:
When someone visits https://acteng.github.io/atip/pipeline/scheme.html, set the default URL parameter appropriately. We only want this for this branch right now

### DIFF
--- a/src/pages/ChooseArea.svelte
+++ b/src/pages/ChooseArea.svelte
@@ -15,6 +15,7 @@
   import { findSmallestAuthority, type AuthorityBoundaries } from "boundaries";
   import {
     appVersion,
+    defaultSchema,
     FileInput,
     getAuthoritiesGeoJson,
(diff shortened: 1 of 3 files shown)


src/pages/ChooseArea.svelte
```diff
@@ -102,7 +103,7 @@
   function detectSchema(gj: FeatureCollection): Schema {
     if (gj.features.length > 0) {
       let props = gj.features[0].properties;
-      for (let schema of ["planning", "v2", "atf4"]) {
+      for (let schema of ["planning", "v2", "atf4", "pipeline"]) {
```
dabreegster, line 106: This is for the choose area page to auto-detect a file with pipeline properties

dabreegster:
Non-controversial PR, so merging and asking for review later

Issue #152, opened 2023-05-04
Import multiple files


dabreegster:
There are some use cases where we want to load all submitted ATIPs in one area at the same time, or maybe show proposed infra alongside a file representing existing infra.

For context, #94 was an old experiment to do this, showing one file at a time.

## UX

It could be useful to draw/symbolize objects in two different ways. Right now we color by some kind of object type. The user should be able to switch between this, or coloring by the file/layer.

The user-facing concept of a layer seems perfect here. Someone should be able to toggle a file/layer on and off, add more layers, etc.

**Big question**: how should editing work? In QGIS and similar, I think every layer has a state -- hidden, readonly, or editing. Multiple layers can be editable at the same time, by clicking different objects. But if you draw a new object, we have to decide what layer to place it in! And potentially have controls to move objects between different layers.

## Implementation

This is complex enough to be worth figuring out the intended UX first before diving in. Today we have one `gjScheme` store and a bunch of IDs everywhere.

One option is to keep that -- flatten/union multiple files/layers into one GeoJSON object internally. One flat ID space. On every feature, add a property to indicate what layer it lives in. When you export or save the file, we can then split things again and produce multiple GeoJSON files. Numeric feature IDs would jump around wildly if people load multiple files in different combinations or orders, but -- that's kind of OK, because even today those IDs aren't considered stable over time.

Another option is to plumb through a dictionary of layer names -> GeoJSON files as the main app state. Anything in the code touching this state has to also use a layer ID. Any feature IDs maybe become a tuple (layer ID, feature ID).

dabreegster:
This kind of functionality now exists in SB (showing all local schemes in one place), and the user can "add scheme from file" to edit multiple things at once

PR #387, opened 2023-11-08
Revert superscheme

dabreegster:
There's a critical problem with the superscheme PR: edits to the per-scheme pipeline form aren't preserved across page refresh. For the workshop tomorrow, it's more important to start collecting that form than to try showing multiple schemes at the same time.

A few problems that looked like they were introduced in the merge/rebase:
- The top-level GJ object has both a `subschemes` and a `pipeline` field: https://github.com/acteng/atip/blob/d9c99cac39163a4b975d7adaee5c2eef525bc2c6/src/types.ts#L15. The intention of `pipeline` was to wrap all the new scheme-level data in one place. `subschemes` and the format there don't match the browse page -- it's a third thing. We need to agree on what the GJ representation is -- see https://github.com/acteng/atip/pull/379#issuecomment-1784928203
- The `PipelineScheme` type is duplicated https://github.com/acteng/atip/blob/d9c99cac39163a4b975d7adaee5c2eef525bc2c6/src/types.ts#L107 and https://github.com/acteng/atip/blob/d9c99cac39163a4b975d7adaee5c2eef525bc2c6/src/types.ts#L122 -- I'm not sure what downstream type-checking issues this could cause
- There's some code duplication I find very confusing. `PipelineSchemeDetails.svelte` and `EntireScheme.svelte` have lots of overlap, and the new `backfillSuperscheme` is repeating the original `backfill`

I think my biggest confusion is why this new feature is specialized for the pipeline schema. In my mind, this should be tackling #152 -- a way to edit multiple files/layers/schemes for the same authority boundary in the sketch tool at the same time. It has nothing particular to do with the attributes we're collecting in the pipeline schema. We should think through https://github.com/acteng/atip/pull/379#issuecomment-1784928203 and settle what the GJ data looks like, how the UI should look, and then the implementation can follow from that.

My vote for the GJ representation is to support only two types of files: a single scheme and a combined scheme. We have both today (the second in the browse tool), and we can adjust them to better include the new pipeline fields


robinlovelace-ate:
Important fix ready for workshop by the looks.

robinlovelace-ate (approved):

dabreegster:
I'll merge to simplify all the rebasing for the `pipeline` branch. I'm also going to take a step back in the workshop tomorrow and ask:

- If it's clear when one scheme has many routes, vs when there are many schemes in the same area
- What the desired workflow will be for people filling out data in those areas with many schemes
- If they even want to edit multiple schemes at once, or if that'll be too confusing
- About the UI they'd expect

Issue #389, opened 2023-11-13
Sketch features / UI improvements


dabreegster:
We checked how useful some ideas would be in the workshop last week, and all of these seem relevant.

- [x] Adding a georeferenced image to sketch over, partly dependent on if high-res maps are available. The initial prototype in `move_images` branch doesn't work yet, and also, our partner asked for "control points". Click two points in the image loaded, and click the corresponding spots on the map to automatically find scale+translation+rotation.
- [ ] Optionally show tooltips near route snapper waypoints as you drag around
- [x] Use the geocoder to add waypoints to the route snapper
- [x] Undo/redo
- [x] Add support for specifying a single optional coverage polygon

as well as all the multiple-scheme stuff, will put in the other issue

dabreegster:
Just dumping TODOs from meetings:

- [ ] Overhaul budget section. One total numeric cost, two checkboxes for development/construction funded. Multiple funding sources possible, and add a freetext Other. Check all of this.
- [ ] Revisit the sub-types with Robin
- [ ] Accuracy of mapped data needs to be required
- [ ] Check copy/paste with govuk textboxes
- [ ] Add optional cost and estimated completion year to per-intervention form. In scheme-wide details, show sum of intervention costs and max year as a hint to fill out the scheme-wide one.
- [ ] Revisit ATF4 dropdown for interventions -- maybe should be checkboxes
- [ ] Lift the "file" menu out of the scheme controls. Open file, save file, reset everything. But critically, the new control is to **add** a file to the current loaded set. The workflow will be several people independently mapping schemes, then one person collecting them together and making sure they join up properly.
- [ ] Stop autofilling route name in pipeline mode, but keep as an option. Add hint text to say use the LCWIP PDF.

dabreegster:
Remaining things here don't seem relevant anymore, so closing

PR #390, opened 2023-11-13
Let waypoints be added to the route snapper using the geocoder

dabreegster:
https://github.com/acteng/atip/assets/1664407/1ebb8c54-2c3a-4718-93fc-3c9a951d1195
Demo at https://acteng.github.io/atip/route_snapper_geocoder/scheme.html?authority=LAD_Adur   (in a few minutes)

#389 
The UX isn't perfect, but feels like a usable start to me

### DIFF
--- a/src/lib/draw/route/GeocoderControls.svelte
+++ b/src/lib/draw/route/GeocoderControls.svelte
@@ -1,0 +1,50 @@
+<script lang="ts">
+  import type { FeatureCollection } from "@maptiler/geocoding-control/types";
+  import type { Position } from "geojson";
+  import { TextInput } from "lib/govuk";
+  import { emptyGeojson } from "lib/maplibre";
+  import { map, routeTool } from "stores";
+  import { geocoderGj } from "./stores";
+
+  let query = "";
+
+  async function runQuery(query: string) {
+    if (!query) {
+      $geocoderGj = emptyGeojson() as FeatureCollection;
+      return;
+    }
+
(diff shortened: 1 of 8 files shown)


src/lib/draw/route/GeocoderControls.svelte
```diff
@@ -0,0 +1,50 @@
...
+      return;
+    }
+
+    let params = new URLSearchParams({
+      country: "gb",
+      proximity: $map.getCenter().toArray().join(","),
+      fuzzyMatch: "true",
+      key: import.meta.env.VITE_MAPTILER_API_KEY,
+    }).toString();
+    let url = `https://api.maptiler.com/geocoding/${query}.json?${params}`;
```
dabreegster, line 23: https://docs.maptiler.com/cloud/api/geocoding/ for reference. Maybe we should specify the bbox as the authority boundary area and further limit results?

src/pages/SketchSchemes.svelte
```diff
@@ -117,6 +118,7 @@
       {:else if $mode.mode == "split-route"}
         <SplitRouteMode />
       {/if}
+      <GeocoderLayer />
```
dabreegster, line 121: It's kind of annoying to split the maplibre and sidebar components for this functionality that logically belongs together, but it's necessary because of how svelte-maplibre components need to be nested in the right place to work

Pete-Y-CS (approved):

PR #391, opened 2023-11-13
Allow user to draw a polygon denoting the coverage of a scheme/lcwip …

Pete-Y-CS:
…(NB this needs to be clear whether it's for a whole pipeline or individual schemes, a requirement for ATE internally to explain)

Just need to make coverage polygons pale on the map

### DIFF
--- a/src/lib/draw/InterventionLayer.svelte
+++ b/src/lib/draw/InterventionLayer.svelte
@@ -4,6 +4,7 @@
   import {
     addLineStringEndpoints,
     isLine,
+    isNotCoveragePolygon,
     isPoint,
     isPolygon,
     layerId,
@@ -131,7 +132,7 @@
 
   <FillLayer
     {...layerId("interventions-polygons")}
-    filter={["all", isPolygon, hideWhileEditing]}
+    filter={["all", isPolygon, hideWhileEditing, isNotCoveragePolygon]}
     paint={{
       "fill-color": color,
       "fill-opacity": 0.2,
--- a/src/lib/draw/InterventionLayer.svelte
+++ b/src/lib/draw/InterventionLayer.svelte
@@ -157,6 +157,7 @@
       "line-width": 0.7 * lineWidth,
     }}
   />
+
   <LineLayer
     {...layerId("interventions-coverage-polygons-outlines")}
     filter={["all", isPolygon, hideWhileEditing, isCoveragePolygon]}
@@ -165,5 +166,14 @@
       "line-opacity": 0.5,
       "line-width": 0.4 * lineWidth,
     }}
-  />
+    hoverCursor={clickable ? "pointer" : undefined}
+    on:click={onClick}
+    manageHoverState={clickable}
+  >
+    {#if clickable}
+      <Popup openOn="hover" openIfTopMost let:features>
+        <div class="govuk-prose"><p>{tooltip(features)}</p></div>
+      </Popup>
+    {/if}
+  </LineLayer>
 </GeoJSON>
--- a/src/lib/draw/InterventionLayer.svelte
+++ b/src/lib/draw/InterventionLayer.svelte
@@ -3,6 +3,7 @@
   import type { Feature } from "geojson";
   import {
     addLineStringEndpoints,
+    isCoveragePolygon,
     isLine,
     isNotCoveragePolygon,
     isPoint,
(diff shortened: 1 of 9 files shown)


src/lib/forms/PipelineForm.svelte
```diff
@@ -18,8 +18,14 @@
     atf4_type: "",
     accuracy: "",
     is_alternative: false,
+    is_coverage_polygon: false,
   };
 
+  const shouldDisplayCoveragePolygonQuestion: boolean =
+    $gjScheme.features.filter(
```
dabreegster, line 25: Also only if the current feature `geometry.type == "Polygon"` -- no point for lines/points

dabreegster, line 25: Also only if the current feature `geometry.type == "Polygon"` -- no point for lines/points

dabreegster (commented):
Looking good so far! I'd suggest a few next steps...

- near the Scheme Details button, add a checkbox for "show coverage polygon". Disable it if there's not one defined yet. It can go in the userSettings store
- in `draw/InterventionLayer.svelte`, filter out the polygon if it has that property and that "show coverage polygon" boolean is enabled

We could also mark the polygon in the sidebar list, but hopefully people name it something obvious like "Coverage area". Maybe we can even clobber the name when they tick the box

src/types.ts
```diff
@@ -87,6 +87,8 @@ export interface InterventionProps {
   length_meters?: number;
   // For LineStrings and Polygons only
   waypoints?: Waypoint[];
+  //For polygons only
+  is_coverage_polygon: boolean;
```
Pete-Y-CS, line 91: I don't think I can make this optional because I don't think I can ts-ignore the warning when it's assigned to a checkbox in a form

dabreegster: Hmm... this makes previous files incompatible. I think it _can_ be optional, and in the pipeline form, you can see if it's undef and make it be false before the checkbox

dabreegster (commented):
Still in meetings, sorry for rushed review

src/lib/draw/InterventionLayer.svelte
```diff
@@ -3,7 +3,9 @@
   import type { Feature } from "geojson";
   import {
     addLineStringEndpoints,
+    isCoveragePolygon,
     isLine,
+    isNotCoveragePolygon,
```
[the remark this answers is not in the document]
Pete-Y-CS:
```
<FillLayer
    {...layerId("interventions-polygons")}
    filter={["all", isPolygon, hideWhileEditing, ["!", isCoveragePolygon]]}
    paint={{
      "fill-color": color,
      "fill-opacity": 0.2,
    }}
    hoverCursor={clickable ? "pointer" : undefined}
    on:click={onClick}
    manageHoverState={clickable}
  >
  ```
  
  Is this what you had in mind? I agree it should be nicer but doesn't seem to work :(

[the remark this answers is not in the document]
Comment (dabreegster) [reply src/lib/draw/InterventionLayer.svelte:8]: Yeah, that doesn't work? Hmm weird, I can poke at it later, but fine to ignore for now :\

File: src/lib/draw/polygon/PolygonMode.svelte [hunk 4]
```diff
@@ -21,6 +21,7 @@
     gjScheme.update((gj) => {
       feature.id = newFeatureId(gj);
       feature.properties.intervention_type = "area";
+      feature.properties.is_coverage_polygon = "false";
```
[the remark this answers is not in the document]
Pete-Y-CS: huh good point that is odd...

Pete-Y-CS:
> Still in meetings, sorry for rushed review

All good, addressed what I can but the `["!", expression]` expression was not behaving as antcipated

src/lib/forms/PipelineForm.svelte
```diff
@@ -72,4 +81,13 @@
   >
     Is this an alternative route and not the default option?
   </Checkbox>
+  {#if shouldDisplayCoveragePolygonQuestion && props.is_coverage_polygon !== undefined}
+    <Checkbox
+      id={"coverage-polygon-" + id}
+      bind:checked={props.is_coverage_polygon}
+    >
+      Is this a polygon to manually denote what the coverage of the scheme/lcwip
```
dabreegster, line 89: I don't want us to be in the business of figuring out wording here... suggestion:
"Does this polygon show the coverage of the scheme? (All area considered while making the scheme)"

dabreegster (commented):
This feels pretty ready to me! Main thing is that I can't seem to get tooltips / click on the coverage polygon once I've made it. Will leave a comment in the code by that...

src/lib/draw/InterventionLayer.svelte
```diff
@@ -148,11 +150,20 @@
...
     }}
   />
+  <LineLayer
+    {...layerId("interventions-coverage-polygons-outlines")}
+    filter={["all", isPolygon, hideWhileEditing, isCoveragePolygon]}
+    paint={{
+      "line-color": color,
+      "line-opacity": 0.5,
+      "line-width": 0.4 * lineWidth,
+    }}
```
dabreegster, line 167: To make this interactive, we need to copy in these props too:
```
    hoverCursor={clickable ? "pointer" : undefined}
    on:click={onClick}
    manageHoverState={clickable}
  >
    {#if clickable}
      <Popup openOn="hover" openIfTopMost let:features>
        <div class="govuk-prose"><p>{tooltip(features)}</p></div>
      </Popup>
    {/if}
```

src/lib/forms/PipelineForm.svelte
Pete-Y-CS: I've got no strong preference, happy to change this but wording will become clear when we know better what this is for 😅

dabreegster (approved):
Works great for me! Nice

PR #393, opened 2023-11-14
Route snapper maplibre cleanup

dabreegster:
This is diffbased on #392, please review that first. No behavioral change here, just part of #359 cleanup. It _almost_ gets us to showing tooltips with road names for waypoints, but I need to work on some fixes upstream in svelte-maplibre first.


src/lib/draw/polygon/polygon_tool.ts
```diff
@@ -293,7 +246,7 @@ export class PolygonTool {
       gj.features.push(polygon);
     }
 
-    (this.map.getSource(source) as GeoJSONSource).setData(gj);
+    polygonToolGj.set(gj);
```
dabreegster, line 249: This is the big trick I didn't think about before. We have non .svelte files doing stuff, but we want to do the rendering in normal svelte. All we have to do is plumb the GJ back to svelte using stores

Pete-Y-CS (approved):

PR #396, opened 2023-11-16
Use scheme collection geojson

Pete-Y-CS:


src/types.ts
```diff
@@ -7,14 +7,20 @@ export type Schema = "v1" | "v2" | "planning" | "atf4" | "pipeline";
...
 // default GeoJSON types and specify feature properties.
-export interface Scheme {
+export interface SchemeCollection {
   type: "FeatureCollection";
   features: FeatureUnion[];
   // Foreign members
-  scheme_name?: string;
   authority?: string;
   origin?: string;
+  schemes: { [refrence: string]: SchemeData };
```
dabreegster, line 16:
```suggestion
  // Keyed by scheme_reference, which doesn't change over the lifetime of the sketch tool
  schemes: { [refrence: string]: SchemeData };
```

src/types.ts
```diff
@@ -7,14 +7,20 @@ export type Schema = "v1" | "v2" | "planning" | "atf4" | "pipeline";
...
+}
+
+export interface SchemeData {
+  scheme_name: string;
```
dabreegster, line 20: This was optional previously. Why don't we keep it like that? It'll simplify a bit of the code in getEmptyScheme

src/lib/sidebar/scheme_data.ts
```diff
@@ -22,6 +23,10 @@ export function backfill(json: Scheme) {
     // Always overwrite IDs, and follow what newFeatureId requires.
     f.id = idCounter++;
   }
+  const uuid = uuidv4();
+  json.schemes = json.schemes || {
+    uuid: getEmptySchemeData(uuid),
```
dabreegster, line 28: I'm fairly sure this creates a dictionary literally with the key "uuid", not the variable. How about
```
if (!json.schemes) {
  json.schemes = {};
  json.schemes[uuid] = {
    scheme_reference: uuid,
  };
}
```

src/lib/sidebar/scheme_data.ts
```diff
@@ -132,3 +137,22 @@ export function getUnexpectedProperties(
 
   return copy;
 }
+
+export function getFirstSchemeOrEmptyScheme(
+  schemeCollection: SchemeCollection
+): SchemeData {
+  if (schemeCollection.schemes != undefined) {
+    const schemeReferences = Object.keys(schemeCollection.schemes);
```
dabreegster, line 145: Does something like `Object.keys(schemeCollection.schemes ?? {})` work, to remove the branching?

src/lib/sidebar/PipelineSchemeForm.svelte
```diff
@@ -10,15 +10,18 @@
     TextArea,
     TextInput,
   } from "lib/govuk";
-  import { gjScheme } from "stores";
+  import { gjSchemeCollection } from "stores";
   import ATF4Type from "../forms/ATF4Type.svelte";
+  import { getFirstSchemeOrEmptyScheme } from "./scheme_data";
 
   let showModal = false;
+  let scheme = getFirstSchemeOrEmptyScheme($gjSchemeCollection);
```
dabreegster, line 18: A TODO here would be good to mark we'll change this in the next step, which'll add UI support for multiple schemes. Right now we're assuming there's 0 or 1 schemes filled out

src/lib/sidebar/PipelineSchemeForm.svelte
```diff
@@ -10,15 +10,18 @@
...
+  $gjSchemeCollection.schemes[scheme.scheme_reference] = scheme;
 
   // This component is only created once, but gjScheme could be cleared
   // completely multipel times. Set defaults anytime the modal is open.
   $: if (showModal) {
-    $gjScheme.pipeline ||= {
+    scheme.pipeline ||= {
```
dabreegster, line 24: I can check when I'm home, but can you make sure the indirection works with Svelte? Every time you edit a pipeline form field, in the console you should see it saving the GJ to local storage. This was a subtle bug a few PRs back

EDIT: never mind, you're copying things over in `onCompletion`. That should be fine. Maybe a comment there to note nothing is saved until then, because it's unfortunate Svelte subtlety

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -24,6 +28,10 @@
 
   let loaded = false;
   let displayClearAllConfirmation = false;
+  let scheme: SchemeData;
```
dabreegster, line 31: `$: scheme = getFirst($gjSchemeCollection)` is equivalent and handles cleaning up the subscription. But... I need to think about this a bit more; this will run constantly every single time anything in the app does anything. It's only meant to handle loading a new file, clearing everything, etc?

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -112,63 +125,67 @@
...
-
-    <SecondaryButton on:click={exportToGeojson}>
-      Export to GeoJSON
-    </SecondaryButton>
-
-    <div>
-      <WarningButton
-        on:click={() => (displayClearAllConfirmation = true)}
-        disabled={$gjScheme.features.length == 0}
+{#if scheme}
```
dabreegster, line 140: Just to reduce nesting and the git diff, could we combine? `{#if scheme && $mode.mode == "list}`

src/lib/sidebar/scheme_data.ts
```diff
@@ -22,6 +23,10 @@ export function backfill(json: Scheme) {
     // Always overwrite IDs, and follow what newFeatureId requires.
     f.id = idCounter++;
   }
```
dabreegster, line 25: I think we need logic here to backfill the previous format we've been using. We might have `json.scheme_name` set. If so, we need to move it over to the default `schemes` thing that gets created

src/lib/sidebar/EntireScheme.svelte
```diff
@@ -36,23 +44,28 @@
...
     } else if (loadLocal) {
       try {
-        gjScheme.set(backfill(JSON.parse(loadLocal)));
+        gjSchemeCollection.set(backfill(JSON.parse(loadLocal)));
       } catch (err) {
         console.log(`Failed to load from local storage: ${err}`);
       }
     }
     loaded = true;
+    scheme = getFirstSchemeOrEmptyScheme($gjSchemeCollection);
```
dabreegster, line 59: It feels like a few places set up this default empty scheme if it's missing. Is it something `backfill` can do? We always call that when loading a file from anywhere. It should return a valid SchemeCollection. Part of that could be filling out this default empty bit

src/lib/sidebar/EntireScheme.svelte
Pete-Y-CS: What you've said presumably still works but the reason I have to do this probably goes away with the next piece of work: it's so that when you update the scheme in the pipeline form the name update gets propagated

src/lib/sidebar/scheme_data.ts
Pete-Y-CS: You are absolutely right: I have encountered and fixed this during fixing the tests 😅

src/lib/sidebar/EntireScheme.svelte
Pete-Y-CS: (Which presumably is resolved more nicely when we properly support multiple schemes on sketch page)

dabreegster (approved):
We should be good now! After merging, need to update Sharepoint and the GCS bucket

PR #397, opened 2023-11-16
Let the user temporarily add an image to the map for georeferencing

dabreegster:
#389 
Demo: https://acteng.github.io/atip/move_images_v3/scheme.html?authority=LAD_Adur

Here's an example from Hackney: https://consultation.hackney.gov.uk/streetscene/butterfield-green/results/butterfieldgreen-delegatedpowersdecision.docx2.pdf

It took me a bit to align things, but here's a demo of how it looks in the end:

https://github.com/acteng/atip/assets/1664407/48490635-7291-4df7-9bf3-a4c92541c65c

This is just a first cut. Next, I'll work on picking any two control points on the raw image, then finding the same places on the map. Going by corners is hard!

### DIFF
--- a/src/lib/draw/image/ImageLayer.svelte
+++ b/src/lib/draw/image/ImageLayer.svelte
@@ -1,0 +1,96 @@
+<script lang="ts">
+  import type { ImageSource } from "maplibre-gl";
+  import { map, mode } from "stores";
+  import { onDestroy } from "svelte";
+  import { Marker } from "svelte-maplibre";
+  import { imgSrc, opacity } from "./stores";
+
+  // TODO Upstream an ImageLayer component
+
+  $: setupSource($imgSrc);
+
+  let source = "image-source";
+  let layer = "image-layer";
+
+  let topLeft = { lng: 0, lat: 0 };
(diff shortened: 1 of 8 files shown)


src/lib/draw/image/ImageLayer.svelte
```diff
@@ -0,0 +1,96 @@
+<script lang="ts">
+  import type { ImageSource } from "maplibre-gl";
+  import { map, mode } from "stores";
+  import { onDestroy } from "svelte";
+  import { Marker } from "svelte-maplibre";
+  import { imgSrc, opacity } from "./stores";
+
+  // TODO Upstream an ImageLayer component
```
dabreegster, line 8: Most of the code here can go away soon. I'll add something to svelte-maplibre to wrap up the reactivity properly

dabreegster (commented):
Just recording things that didn't work:
- a symbol layer, dynamically setting icon size based on zoom... it also needs to know latitude
- a polygon layer with a texture. maplibre insists on tiling the image.

Pete-Y-CS (approved):

PR #398, opened 2023-11-17
Form changes for pipeline

Pete-Y-CS:
Handling a couple of aspects of https://github.com/acteng/atip/issues/389#issuecomment-1814961227 :

-  Add optional cost and estimated completion year to per-intervention form. In scheme-wide details, show sum of intervention costs and max year as a hint to fill out the scheme-wide one.
-  Stop autofilling route name in pipeline mode, but keep as an option. Add hint text to say use the LCWIP PDF
-  Accuracy of mapped data needs to be required

### DIFF
--- a/src/lib/forms/PipelineForm.svelte
+++ b/src/lib/forms/PipelineForm.svelte
@@ -115,10 +115,7 @@
     </Checkbox>
   {/if}
 
-  <TimescaleSubform
-    forIntervention
-    bind:timescale={props.pipeline.intervention_timescale}
-  />
+  <TimescaleSubform bind:timescale={props.pipeline.intervention_timescale} />
   <NumberInput
     label="Cost"
     width={10}
--- a/src/lib/forms/PipelineForm.svelte
+++ b/src/lib/forms/PipelineForm.svelte
@@ -6,8 +6,10 @@
     SecondaryButton,
     Select,
     TextArea,
+    NumberInput
   } from "lib/govuk";
   import { prettyPrintMeters } from "lib/maplibre";
+  import TimescaleSubform from "lib/sidebar/pipeline_forms/TimescaleSubform.svelte";
   import { gjSchemeCollection, routeTool } from "stores";
   import type { InterventionProps } from "types";
   import ATF4Type from "./ATF4Type.svelte";
@@ -19,6 +21,16 @@
     atf4_type: "",
     accuracy: "",
     is_alternative: false,
+    cost: 0,
+    intervention_timescale: {
+      status: "",
+      timescale: "",
+    },
+  };
+
+  props.pipeline.intervention_timescale ||= {
+    status: "",
+    timescale: "",
   };
 
   props.is_coverage_polygon ||= false;
@@ -83,6 +95,7 @@
       ["medium", "Medium"],
       ["low", "Low"],
     ]}
+    required
     inlineSmall
     bind:value={props.pipeline.accuracy}
   />
@@ -102,4 +115,15 @@
       while making the scheme)
     </Checkbox>
   {/if}
+
+  <TimescaleSubform
+    forIntervention
+    bind:timescale={props.pipeline.intervention_timescale}
+  />
+  <NumberInput
+    label="Cost"
+    width={10}
+    min={0}
+    bind:value={props.pipeline.cost}
+  />
 {/if}
--- a/src/lib/forms/PipelineForm.svelte
+++ b/src/lib/forms/PipelineForm.svelte
@@ -2,11 +2,11 @@
   import {
     Checkbox,
     FormElement,
+    NumberInput,
     Radio,
     SecondaryButton,
     Select,
     TextArea,
-    NumberInput
   } from "lib/govuk";
   import { prettyPrintMeters } from "lib/maplibre";
   import TimescaleSubform from "lib/sidebar/pipeline_forms/TimescaleSubform.svelte";
--- a/src/lib/forms/PipelineForm.svelte
+++ b/src/lib/forms/PipelineForm.svelte
@@ -21,7 +21,6 @@
     atf4_type: "",
     accuracy: "",
     is_alternative: false,
-    cost: 0,
     intervention_timescale: {
       status: "",
       timescale: "",
--- a/src/lib/forms/PipelineForm.svelte
+++ b/src/lib/forms/PipelineForm.svelte
@@ -117,7 +117,7 @@
 
   <TimescaleSubform bind:timescale={props.pipeline.intervention_timescale} />
   <NumberInput
-    label="Cost"
+    label="Cost (GBP)"
     width={10}
     min={0}
     bind:value={props.pipeline.cost}
--- a/src/lib/govuk/NumberInput.svelte
+++ b/src/lib/govuk/NumberInput.svelte
@@ -43,7 +43,7 @@
 
 <FormElement {label} id={label}>
   {#if hint}
-    <p>{hint}</p>
+    <div class="govuk-hint">{hint}</div>
   {/if}
   <ErrorMessage errorMessage={validate(stringValue)} />
   <input
--- a/src/lib/govuk/NumberInput.svelte
+++ b/src/lib/govuk/NumberInput.svelte
@@ -9,6 +9,7 @@
   // Inclusive
   export let min: number | undefined = undefined;
   export let max: number | undefined = undefined;
+  export let hint: string | undefined = undefined;
 
   let stringValue: string | undefined = value?.toString();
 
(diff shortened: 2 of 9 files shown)


src/lib/forms/PipelineForm.svelte
```diff
@@ -18,6 +19,24 @@
     atf4_type: "",
     accuracy: "",
     is_alternative: false,
+    intervention_budget: {
```
dabreegster, line 22: This part is redundant with below. The below one is nicer, because it'll handle files we've been making the last few weeks

src/lib/forms/PipelineForm.svelte
```diff
@@ -29,25 +48,11 @@
       props.is_coverage_polygon === true) &&
     props.intervention_type === "area";
 
-  // Sets the intervention name to "From {road1 and road2} to {road3 and
```
dabreegster, line 32: So for this, I think we settled on leaving the Autofill button, but not filling out the name by default. So I think the change is near https://github.com/acteng/atip/blob/62cd8ea959440f65ebc102612d6ab8b44d434387/src/lib/draw/route/RouteMode.svelte#L27, if the schema is pipeline, don't copy over the route_name we get. Same in EditGeometryMode, it copies route_name sometimes

src/lib/sidebar/PipelineSchemeForm.svelte
```diff
@@ -25,17 +25,73 @@
     scheme = getArbitraryScheme($gjSchemeCollection);
     scheme.pipeline ||= {
       scheme_type: "",
-      status: "",
-      timescale: "",
+      scheme_timescale: { status: "", timescale: "" },
```
dabreegster, line 28: Similar to the other file, I think it's safer to do these as separate ||= statements below to keep backwards compatibility. If scheme.pipeline.scheme_budget is missing, set these defaults.

(I'm getting errors when I click scheme details with existing GJ files and with new ones)

src/lib/forms/PipelineForm.svelte
Pete-Y-CS: Hmm it's annoying that the RouteMode needs to know about the schema now
I wonder if we can get around it

dabreegster: We should probably stick schema in a global store; plumbing it is annoying

src/lib/sidebar/pipeline_forms/BudgetSubform.svelte
dabreegster, line 1: A few problems with sharing this per intervention and scheme...

1) at the intervention level, the fields shouldn't be required
2) We only talked about capturing a cost and completion year for interventions. I don't think we want the full forms for each intervention, unless we hear otherwise. It seems like hugely more difficult to ask for this level of detail

Pete-Y-CS: Hmm okay, do you still want GBP funded/Unfunded?

src/types.ts
```diff
@@ -122,6 +129,8 @@ export interface PipelineIntervention {
   accuracy: "high" | "medium" | "low" | "";
   is_alternative: boolean;
 
+  intervention_timescale: PipelineTimescale;
```
dabreegster, line 132: Just an optional estimated completion year, not the full thing, based on the meeting yesterday?

src/types.ts
```diff
@@ -122,6 +129,8 @@ export interface PipelineIntervention {
...
+  cost: number;
```
dabreegster, line 133: Shouldn't this be optional too? `NumberInput` handles this without hassle

src/lib/forms/PipelineForm.svelte
dabreegster: Still need to revert the changes here -- keep the autofill button, we just want to not populate it by default

dabreegster (commented):
Haven't looked at everything yet; it's changing from underneath me ;)

src/types.ts
dabreegster: This is still the whole timescale sub-form -- I think we only want an optional completion year

src/lib/govuk/NumberInput.svelte
```diff
@@ -41,6 +42,9 @@
 </script>
 
 <FormElement {label} id={label}>
+  {#if hint}
+    <p>{hint}</p>
```
dabreegster, line 46: How's `<div class="govuk-hint">` look? From "hint text" section at https://design-system.service.gov.uk/components/text-input/

src/lib/govuk/Radio.svelte
```diff
@@ -23,6 +25,9 @@
     <legend class="govuk-fieldset__legend govuk-label--s">
       {legend}
     </legend>
+    {#if hint}
+      <p>{hint}</p>
```
dabreegster, line 29: Same here

src/lib/forms/PipelineForm.svelte
```diff
@@ -19,6 +21,16 @@
     atf4_type: "",
     accuracy: "",
     is_alternative: false,
+    cost: 0,
```
dabreegster, line 24: We can leave this off and not fill it out. I agree it's a bit inconsistent/confusing now, but in `types.ts` if it's optional, then we don't have to set a default. Since we collect cost with `NumberInput` and that handles undefined for us, it should work out OK

src/lib/forms/PipelineForm.svelte
```diff
@@ -101,4 +117,15 @@
       while making the scheme)
     </Checkbox>
   {/if}
+
+  <TimescaleSubform
+    forIntervention
+    bind:timescale={props.pipeline.intervention_timescale}
+  />
+  <NumberInput
+    label="Cost"
```
dabreegster, line 126: Cost (GBP)

dabreegster (commented):
Partly done

src/lib/sidebar/scheme_data.ts
```diff
@@ -197,3 +204,112 @@ export function getArbitraryScheme(
 ): SchemeData {
   return Object.values(schemeCollection.schemes)[0];
 }
+
+export function getRectifiedPipelineSchemeWithSuggestedValues(
```
dabreegster, line 208: Are we going with the hint approach? Why are we modifying `scheme` at all? Can the implementation of this just be:

1) In `PipelineschemeForm`, call a helper function that finds the max estimated year from features, or undefined if nothing's filled out. Display a bit of text near that field with the value.
2) Similar for summing cost per feature

I'm getting errors when I open "scheme details" with prior data; people are going to hit those tomorrow. We can fix these cases, but I'm finding code in here very confusing -- it's mutating the scheme and trying to fill out more default values. If we need to do that, centralized in `backfill` like we've been doing? But do we need to at all for hints -- shouldn't it just be a few readonly methods?

src/lib/sidebar/pipeline_forms/TimescaleSubform.svelte
```diff
@@ -0,0 +1,89 @@
+<script lang="ts">
+  import { NumberInput, Radio } from "lib/govuk";
+  import type { PipelineTimescale } from "types";
+
+  export let timescale: PipelineTimescale;
+  export let maxTimescaleFromInterventions: string | undefined = undefined;
```
dabreegster, line 6: Do we have to plumb this around? If `forIntervention` is false and we're in Scheme details, we can call a helper function that looks at $gjSchemeCollection

src/lib/sidebar/PipelineSchemeForm.svelte
```diff
@@ -101,95 +103,11 @@
...
-
-      <Checkbox id="scheme-funded" bind:checked={scheme.pipeline.funded}>
-        Is the scheme fully funded?
-      </Checkbox>
-    </fieldset>
+    <TimescaleSubform
+      bind:timescale={scheme.pipeline.scheme_timescale}
+      maxTimescaleFromInterventions={maxTimescale}
+    />
+    <BudgetSubform bind:budget={scheme.pipeline.scheme_budget} {summedCost} />
```
dabreegster, line 110: Similar comment here -- why is the parent component bothering with the summedCost and then passing it down? BudgetSubForm can just call a read-only helper function on $gjSchemeCollection and get back a hint to diplay

dabreegster (approved):
A backfill case or two is broken, but we'll check in an imminent meeting about if we can safely break some of these cases

dabreegster (commented):
Just noting we decided to hold fire on this. We might decide to copy down the timescale and budget sections down to per-intervention, but have to think through the UX of the forms carefully

dabreegster:
Should we close this PR (but keep the branch) just to avoid accidentally merging?

PR #402, opened 2023-11-21
When concatting imported schemes reset all IDs

Pete-Y-CS:


dabreegster (approved):
Thanks for fixing this! I made it start at 1 to be consistent with https://github.com/acteng/atip/blob/c26054991ca5c3ce7bd26a8112e70d0cedb2c3d4/src/stores.ts#L50 -- some of the libraries using maplibre are not careful enough about the truthiness of 0 vs null

PR #403, opened 2023-11-21
Allow reordering of schemes in list

Pete-Y-CS:
Currently have logic lifted to ListMode because it feels like that's what should be managing the order of schemes but I don't mind putting it in PerSchemeControls if we prefer

### DIFF
--- a/src/lib/sidebar/ListMode.svelte
+++ b/src/lib/sidebar/ListMode.svelte
@@ -40,6 +40,56 @@
       errorFromFile = `Couldn't load scheme from a file: ${err}`;
     }
   }
+
+  function moveSchemeInList(scheme_reference: string, direction: number) {
+    return function () {
+      const currentSchemeOrder = JSON.parse(
+        JSON.stringify(Object.keys($gjSchemeCollection.schemes))
+      );
+      const currentIndex = currentSchemeOrder.indexOf(scheme_reference);
+      if (
+        currentIndex + direction >= 0 &&
+        currentIndex + direction < currentSchemeOrder.length
+      ) {
+        const newSchemeOrder = swapArrayValues(
+          currentSchemeOrder,
+          currentIndex,
+          currentIndex + direction
+        );
+        const newSchemesObject = getReorderedSchemesObject(
+          newSchemeOrder,
+          $gjSchemeCollection.schemes
+        );
+        $gjSchemeCollection.schemes = newSchemesObject;
+        $gjSchemeCollection = $gjSchemeCollection;
+      }
+    };
+  }
+
+  function swapArrayValues(
+    array: any[],
+    firstIndex: number,
+    secondIndex: number
+  ) {
+    const firstValue = array[firstIndex];
+    const secondValue = array[secondIndex];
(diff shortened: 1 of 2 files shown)


src/lib/sidebar/ListMode.svelte
```diff
@@ -55,6 +105,15 @@
 <hr />
 
 {#each Object.keys($gjSchemeCollection.schemes) as scheme_reference}
-  <PerSchemeControls {scheme_reference} />
+  <PerSchemeControls {scheme_reference}>
+    {#if Object.keys($gjSchemeCollection.schemes).length > 1}
+      <SecondaryButton on:click={moveSchemeInList(scheme_reference, -1)}>
```
dabreegster, line 110:
```suggestion
      <SecondaryButton on:click={() => moveSchemeInList(scheme_reference, -1)}>
```
then you don't need to return a callback from the function

src/lib/sidebar/ListMode.svelte
```diff
@@ -40,6 +40,56 @@
...
+          newSchemeOrder,
+          $gjSchemeCollection.schemes
+        );
+        $gjSchemeCollection.schemes = newSchemesObject;
+        $gjSchemeCollection = $gjSchemeCollection;
+      }
+    };
+  }
+
+  function swapArrayValues(
```
dabreegster, line 69: Optional: https://www.freecodecamp.org/news/swap-two-array-elements-in-javascript/ "How to Swap Two Array Elements by Destructuring" looks a bit nicer to me, just a one-liner

src/lib/sidebar/ListMode.svelte
```diff
@@ -40,6 +40,56 @@
...
+  ) {
+    const firstValue = array[firstIndex];
+    const secondValue = array[secondIndex];
+    array[secondIndex] = firstValue;
+    array[firstIndex] = secondValue;
+
+    return array;
+  }
+
+  function getReorderedSchemesObject(
```
dabreegster, line 82: All of this works fine for me in modern Firefox, but according to MDN, `Object.keys` does not guarantee any kind of ordering. https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order is a minefield; I couldn't find a canonical newer answer (preferably in someplace like MDN). We should maaaybe use something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames instead, but it seems to work now, so let's just go with it.

dabreegster (approved):
Nice and simple!

src/lib/sidebar/ListMode.svelte
Pete-Y-CS: I don't think this works?

Pete-Y-CS: Or at least I don't understand it, so for now I'm gonna leave this change

src/lib/sidebar/ListMode.svelte
Pete-Y-CS: done, just testing

PR #409, opened 2023-11-23
Be very careful with pipeline scheme data

dabreegster:
1) Add a test checking that the warnings for missing data behave as expected, and that data gets saved and restored after a refresh
2) Simplify how the pipeline backfill logic happens
3) Fix an existing bug where scheme details aren't actually saved if you close the modal by clicking the map!

### DIFF
--- a/src/lib/sidebar/GenericSchemeForm.svelte
+++ b/src/lib/sidebar/GenericSchemeForm.svelte
@@ -2,29 +2,16 @@
   import { Modal } from "lib/common";
   import { DefaultButton, SecondaryButton, TextInput } from "lib/govuk";
   import { gjSchemeCollection } from "stores";
-  import type { SchemeData } from "types";
 
   export let scheme_reference: string;
 
   let showModal = false;
-  let scheme: SchemeData | null = null;
-
-  $: if (showModal) {
-    scheme = $gjSchemeCollection.schemes[scheme_reference];
-  }
 
   function onKeyDown(e: KeyboardEvent) {
     if (showModal && e.key == "Escape") {
       e.stopPropagation();
-      finish();
+      showModal = false;
     }
-  }
-
-  // No changes in the form are saved until this happens
-  function finish() {
-    $gjSchemeCollection.schemes[scheme!.scheme_reference] = scheme!;
-    $gjSchemeCollection = $gjSchemeCollection;
-    showModal = false;
   }
 </script>
 
@@ -34,9 +21,11 @@
   Scheme details
 </SecondaryButton>
 <Modal title="Scheme details" bind:open={showModal}>
-  {#if scheme}
-    <TextInput label="Scheme name" required bind:value={scheme.scheme_name} />
+  <TextInput
+    label="Scheme name"
+    required
+    bind:value={$gjSchemeCollection.schemes[scheme_reference].scheme_name}
+  />
 
-    <DefaultButton on:click={finish}>Save</DefaultButton>
-  {/if}
+  <DefaultButton on:click={() => (showModal = false)}>Save</DefaultButton>
 </Modal>
--- a/src/lib/sidebar/ListMode.svelte
+++ b/src/lib/sidebar/ListMode.svelte
@@ -1,18 +1,12 @@
 <script lang="ts">
-  import { randomSchemeColor } from "colors";
   import { FileInput } from "lib/common";
   import { ErrorMessage, SecondaryButton } from "lib/govuk";
   import { gjSchemeCollection } from "stores";
-  import { v4 as uuidv4 } from "uuid";
   import PerSchemeControls from "./PerSchemeControls.svelte";
-  import { backfill } from "./scheme_data";
+  import { addEmptyScheme, backfill } from "./scheme_data";
 
   function newBlankScheme() {
-    let scheme_reference = uuidv4();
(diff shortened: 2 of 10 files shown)


src/lib/sidebar/GenericSchemeForm.svelte
dabreegster, line 1: The "click outside the modal" bug was present here. Unlike the pipeline case, it's simpler to just directly reference the one value we need, so changes are saved ASAP

src/lib/sidebar/ListMode.svelte
```diff
@@ -107,7 +101,7 @@
 
 <hr />
 
-{#each Object.keys($gjSchemeCollection.schemes) as scheme_reference, i}
+{#each Object.keys($gjSchemeCollection.schemes) as scheme_reference, i (scheme_reference)}
```
dabreegster, line 104: There was a major bug here! At the very start, $gjSchemeCollection is just the emptyCollection. This code runs and creates these components, which have an error, since the blank scheme has no name (or other details). We soon load data from local storage. If it had all the details filled out, you'd expect the error message to go away. But very subtly, `PerSchemeControls` wasn't getting recreated for the new scheme(s) loaded. Instead the existing component was just having its `scheme_reference` changed. Keyed each blocks fix. `PerSchemeControls` should only get mounted once per scheme.

src/lib/sidebar/PipelineSchemeForm.svelte
```diff
@@ -11,26 +11,20 @@
...
-      scheme_type: "",
-      status: "",
-      timescale: "",
-      atf4_lead_type: "",
-      scheme_description: "",
-      funding_source: "",
-      funded: false,
-    };
-  }
+  // This is just for conveience below, but it means most changes are not
```
dabreegster, line 19: It's quite frustrating that the Svelte compiler can't trace through a simple alias for typing convenience like this. Maybe something in Svelte 5 / runes improves it.

src/lib/sidebar/scheme_data.ts
```diff
@@ -39,20 +44,14 @@ export function backfill(json: SchemeCollection) {
     };
 
     // Move over the optional scheme_name from the old place
-    // @ts-ignore handling old format
```
dabreegster, line 42: By declaring the input as `any`, we can get rid of most of this noise. `any` is more appropriate anyway; we don't expect the input to backfill to be valid at all!

src/lib/sidebar/scheme_data.ts
```diff
@@ -62,25 +61,52 @@ export function backfill(json: SchemeCollection) {
...
+  let schema = get(schemaStore);
   for (let scheme of Object.values(json.schemes)) {
-    scheme.color ||= randomSchemeColor();
+    // Ensure every scheme has some color
+    (scheme as any).color ??= randomSchemeColor();
+
+    // Ensure pipeline defaults are set
+    if (schema == "pipeline") {
+      (scheme as any).pipeline ??= emptyPipelineScheme();
+      // Any changes to PipelineScheme after 21 November 2023 must be handled here
```
dabreegster, line 72: There'll be a few in the next PR, but I expect it to be nice and simple to handle here only.

tests/pipeline.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  await page.getByRole("button", { name: "Save" }).click();
+  await expect(page.getByText("Missing some required data")).not.toBeVisible();
+
+  // Check the data in local storage
+  let json = await getLocalStorage(page, "LAD_Adur_pipeline");
+  let scheme = Object.values(json.schemes)[0] as any;
+  expect(scheme.scheme_name).toEqual("Corridor 1");
+  expect(scheme.pipeline).toEqual(
+    expect.objectContaining({
```
dabreegster, line 51: https://playwright.dev/docs/api/class-genericassertions#generic-assertions-object-containing

tests/shared.ts
```diff
@@ -38,3 +38,11 @@ export async function clickMap(page: Page, x: number, y: number) {
     position: { x, y },
   });
 }
+
+export async function getLocalStorage(page: Page, key: string): Promise<any> {
+  let string = await page.evaluate(
```
dabreegster, line 43: https://playwright.dev/docs/evaluating
This seems like the most straightforward way to get local storage

src/lib/sidebar/ListMode.svelte
Robinlovelace: Great fix by the sounds of it!

src/pages/ChooseArea.svelte
```diff
@@ -38,10 +38,7 @@
   };
 
   let showAbout = false;
-  const params = new URLSearchParams(window.location.search);
-  if (params.get("schema")) {
-    schemaStore.set(params.get("schema") as Schema);
-  }
+  let params = new URLSearchParams(window.location.search);
```
Pete-Y-CS, line 41: any particular reason for this change? Don't we prefer const where we can use it?

Pete-Y-CS (approved):

src/pages/ChooseArea.svelte
dabreegster: Oops, just habit from jumping languages constantly. (Rust uses `let` and `let mut`). Will revert. Could be nice to set up eslint or similar in the future to enforce consistency

PR #411, opened 2023-11-24
Show/hide schemes

dabreegster:
New feature is toggling individual schemes in the map. A few other fixes along the way

### DIFF
--- a/src/lib/draw/InterventionLayer.svelte
+++ b/src/lib/draw/InterventionLayer.svelte
@@ -47,12 +47,12 @@
   $: showSchemes = getShowSchemes($hideSchemes);
   function getShowSchemes(schemesToHide: Set<string>): ExpressionSpecification {
     // TODO Can't get https://maplibre.org/maplibre-style-spec/expressions/#in to work
-    let expression = ["any"] as ExpressionSpecification;
-    for (let x of Object.keys($gjSchemeCollection.schemes)) {
-      if (!schemesToHide.has(x)) {
-        // @ts-ignore Can't figure out the problem
-        expression.push(["==", ["get", "scheme_reference"], x]);
-      }
+    // Note we filter _out_ schemes to hide, rather than specify the ones to
+    // show. Otherwise we have to know when new schemes are created.
+    let expression = ["all"] as ExpressionSpecification;
+    for (let x of schemesToHide.keys()) {
+      // @ts-ignore Can't figure out the problem
+      expression.push(["!=", ["get", "scheme_reference"], x]);
     }
     return expression;
   }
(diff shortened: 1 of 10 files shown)


src/lib/draw/InterventionLayer.svelte
```diff
@@ -30,15 +30,33 @@
   $: gj = addLineStringEndpoints($gjSchemeCollection);
 
   // TODO Maybe have a separate component for different modes.
-  const hideWhileEditing: FilterSpecification = [
+  const hideWhileEditing: ExpressionSpecification = [
     "!=",
-    "hide_while_editing",
+    ["get", "hide_while_editing"],
```
dabreegster, line 35: I wasn't aware about https://maplibre.org/maplibre-style-spec/deprecations/. We've been accidentally using the old style in some places. Half the bugs I've been hitting have been due to mixing the old and new style, which fails silently. I'm asking on slack about getting some loud console errors for this!

src/lib/draw/InterventionLayer.svelte
```diff
@@ -30,15 +30,33 @@
...
+  const notEndpoint: ExpressionSpecification = [
+    "!=",
+    ["get", "endpoint"],
     true,
   ];
-  const notEndpoint: FilterSpecification = ["!=", "endpoint", true];
 
   $: clickable = $mode.mode == "list";
 
+  // TODO Can't use this for the interventions-lines-endpoints layer, for unknown reasons
```
dabreegster, line 46: The same `"all"` trick seems to break there, and I'm at my wits end...

src/lib/maplibre/utils.ts
```diff
@@ -15,17 +15,29 @@ import type {
   StyleSpecification,
 } from "maplibre-gl";
 
-export const isPolygon: ExpressionSpecification = ["==", "$type", "Polygon"];
-export const isLine: ExpressionSpecification = ["==", "$type", "LineString"];
-export const isPoint: ExpressionSpecification = ["==", "$type", "Point"];
+export const isPolygon: ExpressionSpecification = [
+  "==",
+  ["geometry-type"],
```
dabreegster, line 20: https://maplibre.org/maplibre-style-spec/deprecations/
Using the new style now

Pete-Y-CS (approved):

dabreegster:
> I will see if I can find anything about the line endpoints, do you have any starting ideas why it might be

No :\  It's not a huge problem right now, just a bit weird to see the floating circles remain. I left some comments here about the deprecated maplibre styles, which we've been accidentally using. I _think_ there's some way to get maplibre to log warnings when we feed in bad styles, but I haven't figured out how yet. That's probably our best bet for figuring out things like this

PR #419, opened 2023-12-04
Add indicator for alternative intervention options

Pete-Y-CS:
Adding an indication of whether a listed intervention in 'an alternative' so that users can easily know if they've set that flag


Pete-Y-CS:
Small feature which I thought was unintrusive enough to add: basically sustrans wanted to know be able to see in the sidebar which segments they'd marked as 'alternative'

PR #420, opened 2023-12-04
Add new schemes to top

Pete-Y-CS:


src/lib/sidebar/scheme_data.ts
```diff
@@ -101,14 +101,20 @@ export function emptyCollection(): SchemeCollection {
...
 export function addEmptyScheme(gj: SchemeCollection) {
   let scheme_reference = uuidv4();
-  gj.schemes[scheme_reference] = {
+  const newSchemes: { [reference: string]: SchemeData } = {};
+  newSchemes[scheme_reference] = {
     scheme_reference,
     color: randomSchemeColor(),
   };
+
+  Object.keys(gj.schemes).forEach((schemeRef) => {
```
dabreegster, line 110: `Object.entries`?
and worth adding a comment here that this relies on insertion-order. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in   " other string keys in ascending chronological order of property creation."

dabreegster (approved):

PR #421, opened 2023-12-04
WIP: Add ability to merge routes

Pete-Y-CS:
So far I have created a tool which allows the user to select multiple routes but currently you can't do anything with that _and_ the click event is doubling up so routes are being double-added.

I was hoping to have click be add/remove so this is currently a bit of a problem but as a work around I can say ctrl+click or right click is remove.

Anyway more work to come but just getting the WIP in early since so review can be ongoing since this is a decent chunk of work


dabreegster:
Closing old PR (but keeping branch)

PR #426, opened 2023-12-31
Update route-snappers package and use new files

dabreegster:
I made a new route-snapper release recently (https://github.com/dabreegster/route_snapper/pull/47) that avoids unnecessary coordinate projection and produces more accurate lengths. This updates ATIP to use it. I've generated and uploaded the v2.5 files for everywhere with https://github.com/acteng/atip-data-prep/pull/43.

Manual tests in different areas all work fine

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "js-cookie": "^3.0.5",
         "maplibre-gl": "^3.5.1",
         "read-excel-file": "^5.6.1",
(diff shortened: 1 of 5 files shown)


dabreegster:
Sounds good! Since I don't have GCS access, I can give you read access to some place with all of these files and you can copy them over. And I'd still  vote for doing something like https://github.com/acteng/atip-data-prep/pull/38 (maybe in a private repo) so we can sanely track updates

PR #427, opened 2024-01-02
Split stores.ts

dabreegster:
The browse and draw pages (and some of the other smaller pages like critical entry) do different things, sharing some code but not all. The directory structure is already reasonably clear, but we today mostly have one shared `stores.ts`. This splits out stores specific to the drawing page, to make things a little more clear.

Diffbased on #425, review that first please

### DIFF
--- a/src/lib/draw/image/ImageMode.svelte
+++ b/src/lib/draw/image/ImageMode.svelte
@@ -1,11 +1,11 @@
 <script lang="ts">
+  import { mode } from "lib/draw/stores";
   import {
     ButtonGroup,
     DefaultButton,
     FormElement,
     WarningButton,
   } from "lib/govuk";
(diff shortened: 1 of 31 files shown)


src/lib/draw/image/ImageMode.svelte
```diff
@@ -1,11 +1,11 @@
 <script lang="ts">
+  import { mode } from "lib/draw/stores";
```
Pete-Y-CS, line 2: Should we use this or relative pathing everywhere? Is there a reason to use different ones in different places?

dabreegster: Good point, using absolute everywhere (to make later file reorgs easier). Except when it's a "private" file to one directory, like draw/route/.

PR #429, opened 2024-01-04
Include new funding programme and a total length in the browse page

dabreegster:
Coloring by funding was a bit broken and not reactive to new cases like "ATF2 or ATF3" and "LCWIP". Also add a total length count to the `Showing 639 schemes (5,811 routes, 184 areas, 0 crossings, 147 other, with total LineString length of 6606.7 miles)` checkbox, since it's maybe of interest. (But if it's just confusing, will remove)

### DIFF
--- a/src/lib/browse/colors.ts
+++ b/src/lib/browse/colors.ts
@@ -34,9 +34,15 @@ export const colors = {
     "#080C54",
   ],
 
-  atf2: "#00AFFF",
-  atf3: "#FF62DC",
-  atf4: "#FFD833",
+  // A qualitative set from colorbrewer2.org
+  funding_programmes: [
+    "#66c2a5",
+    "#fc8d62",
+    "#8da0cb",
+    "#e78ac3",
+    "#a6d854",
+    "#ffd92f",
+  ],
 };
 
 // For dense line layers, make individual lines easily distinguished when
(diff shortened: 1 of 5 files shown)


src/lib/browse/colors.ts
```diff
@@ -34,9 +34,8 @@ export const colors = {
     "#080C54",
   ],
 
-  atf2: "#00AFFF",
-  atf3: "#FF62DC",
-  atf4: "#FFD833",
+  // A qualitative set from colorbrewer2.org
+  funding_programmes: ["#66c2a5", "#fc8d62", "#8da0cb", "#e78ac3"],
```
Pete-Y-CS, line 38: Actually we may want 5 (When we've resolved data issues):
ATF2, ATF3, ATF4, ATF4e, LCWIP.

No?

Robinlovelace: To make the final category broader, we could also say "Planned pipeline" so it's not tied to any particular process or timescale.

dabreegster: Pete: Fixed, changed to 6 colors for more future-proofness.
Robin: Good point, https://github.com/acteng/atip-scheme-data/pull/11

dabreegster:
Rebased and fixed comments, thanks for reviews

PR #434, opened 2024-01-09
Colour by current milestone

Pete-Y-CS:


src/lib/browse/InterventionColorSelector.svelte
```diff
@@ -53,6 +30,72 @@
       color
     );
   }
+
+  function styleByFundingProgramme() {
```
dabreegster, line 34: Could we add `DataDrivenPropertyValueSpecification<string>` as the return here and others? `import type { DataDrivenPropertyValueSpecification } from "maplibre-gl";`

dabreegster (approved):
Nice!

src/lib/browse/InterventionColorSelector.svelte
```diff
@@ -53,6 +31,72 @@
...
+      colorMapping,
+      "grey"
+    );
+    legendRows = returnedLegendRows;
+    return color;
+  }
+
+  function styleByCurrentMilestone(): DataDrivenPropertyValueSpecification<string> {
+    let stageGates = [
+      "removed",
```
dabreegster, line 61: I was just looking at the possible values here from the real output file, and they all match except for this -- I think it's called "superseded"

PR #436, opened 2024-01-10
Make sure popups update

dabreegster:


src/lib/browse/InterventionPopup.svelte
```diff
@@ -4,7 +4,7 @@
 
   export let props: { [name: string]: any };
 
-  let scheme = $schemes.get(props.scheme_reference)!;
+  $: scheme = $schemes.get(props.scheme_reference)!;
```
dabreegster, line 7: NONE of the values below using `scheme` were changing; they were all based on the first popup you make. It's not always clear to me when a Svelte component gets mounted/created afresh and when the same one is reused, and just input props change. In this case, due to how svelte-maplibre works, it's the latter. So we need to recalc this when `props` change

Pete-Y-CS (approved):

PR #437, opened 2024-01-12
Add new types for pipeline mapping

Pete-Y-CS:


dabreegster (approved):
We should also update the data definition doc used in meetings, but that should be easy -- everyone seemed in support of these

PR #438, opened 2024-01-15
Toggle cycle layers individually

dabreegster:
As requested in the demo last week:

https://github.com/acteng/atip/assets/1664407/76f8d7db-0553-478b-a4e3-0a6fcb5736f8

Also fixed the checkbox size for the last two groups, because it was inconsistent:
Screenshot from 2024-01-15 11-42-06


Robinlovelace:
Just had a watch of the new functionality, very handy to be able to toggle individual infra types, not all cycleways are the same! So big +1 from me, see no downside to this and low risk so let's get it merged is my quickfire take.

Pete-Y-CS (approved):

PR #440, opened 2024-01-16
Sometimes browse page failes to load because map isn't loaded when it…

Pete-Y-CS:
… tries to 'switch off' the street view tool

### DIFF
--- a/src/lib/common/StreetViewTool.svelte
+++ b/src/lib/common/StreetViewTool.svelte
@@ -25,12 +25,14 @@
   }
 
   function off() {
-    $map.off("click", onClick);
-    $map.getCanvas().style.cursor = "inherit";
-    for (let helper of roadLayerHelpers) {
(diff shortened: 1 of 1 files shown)


dabreegster (approved):
I see this error from time to time. We shouldn't even be instantiating this component before maplibre is loaded, since we don't create the slot: https://github.com/acteng/atip/blob/fb92907373f01f8592e88c1c121d78b61dfdc4f4/src/lib/common/MapLibreMap.svelte#L59
But there must be some weird race condition happening.

This PR should help a bit. When this error happens, I also see warnings from our own `getBeforeId`, because none of the layers are set up yet. Ideally we could first repro the problem kind of reliably, then figure out everything going wrong and fix.

Pete-Y-CS:
I agree, those all seems to be originating from Street View Tool too. We should be able to investigate still when we see those warnings. In any case this will hopefully help us see if/what problems this condition will still cause

PR #441, opened 2024-01-16
Fix the race condition when loading a map

Pete-Y-CS:


dabreegster (approved):
Nice spot! Let's see if that sorts it out...

PR #442, opened 2024-01-17
Fix bug detecting schema

dabreegster:
The problem our user hit:

1) They sketched geometry using the v1 tool, somehow messing up the URL to not be in pipeline mode
2) They loaded the file with the correct pipeline mode URL, but then the pipeline forms didn't show up


tests/homepage.spec.ts
dabreegster, line 1: This is the case that was broken

tests/pipeline.spec.ts
dabreegster, line 1: These two cases worked fine already

Pete-Y-CS (approved):

PR #459, opened 2024-02-01
Remove old script in favor of atip-scheme-data

dabreegster:


Pete-Y-CS (approved):
Removing code always good

PR #461, opened 2024-02-05
add a script to auto deploy to the test env

Pete-Y-CS:


.github/workflows/deploy_appengine_test.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Authenticate to GCP
+        uses: google-github-actions/auth@v1
+        with:
+          credentials_json: '${{ secrets.GCP_TEST_CREDENTIALS }}'
+
+      - name: Deploy
+        uses: google-github-actions/deploy-appengine@v1
+        with:
+          project_id: ${{ vars.PROJECT_ID }}
+          deliverables: backend/app.yaml
```
dabreegster, line 47: I think we'll need another copy of this file, https://github.com/acteng/atip/blob/8dc01575ddd1f6482eae9a88eb9a317a6e648a82/backend/app.yaml#L5

And it'd be good to rename the existing .yaml stuff to 'dev' to be clear about dev vs test

backend/app_dev.yaml
```diff
@@ -0,0 +1,6 @@
+runtime: nodejs18
+env_variables:
+  #GCS_BUCKET: "atip-test-2"
```
dabreegster, line 3: You can remove the two commented things, that looks like old stuff from when I was originally getting this to work in my project

dabreegster (approved):
I think this looks fine -- just try and see if it works I guess? :)

PR #462, opened 2024-02-07
Fix the initial legend to show funding programme

Pete-Y-CS:


src/lib/browse/InterventionColorSelector.svelte
```diff
@@ -6,7 +6,8 @@
   import { styleByCurrentMilestone, styleByFundingProgramme } from "./colors";
 
   let colorInterventionsAccordingTo = "fundingProgramme";
-  let legendRows = schemaLegend("v1");
+  const arrayContainingLegendRows = styleByFundingProgramme();
```
Pete-Y-CS, line 9: I cannot do 
`let [unused, legendRows] = styleByFundingProgramme();` without a ts ignore comment. Maybe that would be preferable?

dabreegster: I need to try, but I think `_unused` or maybe even just `_` should work?

dabreegster (approved):

PR #466, opened 2024-02-09
Distinguish some crossing types

dabreegster:
Make it easier to see signalized, by request.

I was going to distinguish zebras as well, but we're not including enough raw OSM tags. We should probablydo the classification in the data-prep side.

### DIFF
--- a/src/lib/browse/colors.ts
+++ b/src/lib/browse/colors.ts
@@ -18,8 +18,9 @@ export const colors = {
   bus_route_with_lane: "#9362BA",
   bus_route_without_lane: "#C2A6D8",
   cycle_parking: "black",
-  crossings: "green",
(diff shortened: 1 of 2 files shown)


src/lib/browse/colors.ts
```diff
@@ -18,7 +18,8 @@ export const colors = {
   bus_route_with_lane: "#9362BA",
   bus_route_without_lane: "#C2A6D8",
   cycle_parking: "black",
-  crossings: "green",
+  signalized_crossing: "green",
```
dabreegster, line 21: I'm just making colors up :(

We can display arbitrary icons in maplibre pretty easily; if we found a good set with a compatible license (like the ones from an existing OSM web map), we could consider using those in the future

PR #467, opened 2024-02-12
add higher education layers

Pete-Y-CS:


Pete-Y-CS:
> This looks great! Two questions:
> 
> 1. Could we just fold the schools layer in here? It feels pretty weird for them to be split -- how about "Schools" as the group, and then "Primary & secondary schools" as the first thing in the individual list? I think the code would be simple, just another entry in your list, and renaming this Schools.svelte as the overall thing. (and we should regen the school layer to use the same date)
> 2. Does our audience know + care about the distinction between college and university? As an American, the terms are pretty fuzzily similar

1. Happy to do so, I think I'd call it education rather than schools because in the UK it's pretty rare to call colleges and Universities schools.
2. Colleges and Universities are very distinct here: colleges tend to be more vocational and universities tend to be more academic. What's more colleges are often immediately following school: ages 16-18, where universities tend to be 18+.

https://uk.indeed.com/career-advice/career-development/college-vs-university-uk#:~:text=If%20the%20institution%20doesn't,for%20children%2C%20it's%20a%20school.

https://edvoy.com/articles/what-is-the-difference-between-a-school-college-and-university-in-the-uk/

dabreegster:
+1 to "Education" and thanks for explaining :)

dabreegster (approved):

dabreegster (approved):
Nice! S3 is updated with education.pmtiles. When everything's merged, we can remove `schools.pmtiles` from all 3 environments and update manifest

PR #468, opened 2024-02-13
Upgrade many dependencies. No behavioral changes.

dabreegster:
Manual and playwright tests don't show any behavioral changes. I'll try other upgrades separately, because they're more complicated.


tests/shared.ts
```diff
@@ -11,7 +11,9 @@ export async function loadInitialPageFromBrowser(
 
 export async function checkPageLoaded(page: Page) {
   // Wait for the map to load by looking for controls
-  await expect(page.getByRole("button", { name: "Zoom in" })).toBeVisible();
+  await expect(
+    page.getByRole("button", { name: "Zoom to show entire boundary" }),
```
dabreegster, line 15: This button name changed ages ago, but Playwright tests have been fine... which doesn't strike confidence into my heart about Playwright.

Pete-Y-CS (approved):

PR #469, opened 2024-02-13
Upgrade govuk-frontend

dabreegster:
Other than the alpha button changing style a bit, I don't notice any behavioral changes.

You'll need to `npm i; npm run setup-govuk` locally after this is in

Full changelog in https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md. I skimmed through and didn't spot anything that should affect us

### DIFF
--- a/browse.html
+++ b/browse.html
@@ -14,9 +14,11 @@
       import BrowseSchemes from "./src/pages/BrowseSchemes.svelte";
 
       // For govuk setup. For some reason, the initAll step doesn't work here.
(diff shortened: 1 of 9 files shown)


dabreegster:
Two tests broke, so not ready for review yet

dabreegster:
Alright, should be good now. For checkboxes, the Playwright locator directives are a little different with new govuk. I've never been super clear on the difference between these selectors.

Pete-Y-CS (approved):

PR #472, opened 2024-02-15
Include optional budget and timing forms for each intervention

dabreegster:
As requested by our user. I referred to #398 but didn't preserve the behavior of aggregating per-feature timing/budget to inform the scheme-wide variant. We could consider adding this back if it's requested, but I'm not sure if there are situations when the values are known both overall and individually.

I added some tests and checked old files, but please also do test things out carefully

### DIFF
--- a/src/lib/sidebar/PipelineSchemeForm.svelte
+++ b/src/lib/sidebar/PipelineSchemeForm.svelte
@@ -20,11 +20,16 @@
   // synced to $gjSchemeCollection until finish() is called
   let pipeline = $gjSchemeCollection.schemes[scheme_reference].pipeline!;
 
-  // Check for all required values
-  $: errorMessage =
-    pipeline.scheme_type && pipeline.status && pipeline.timescale
+  // Svelte doesn't see nested updates in BudgetForm and TimingForm, so use a
+  // counter and manual callback to recalculate errorMessage
+  let updateError = 0;
+  $: errorMessage = checkRequiredValues(pipeline, updateError);
+
+  function checkRequiredValues(_: any, updateError: number): string {
+    return pipeline.scheme_type && pipeline.status && pipeline.timescale
       ? ""
       : "Missing some required data";
+  }
 
   function onKeyDown(e: KeyboardEvent) {
     if (showModal && e.key == "Escape") {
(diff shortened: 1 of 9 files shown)


src/lib/sidebar/PipelineSchemeForm.svelte
```diff
@@ -23,21 +20,16 @@
...
-    "atf4",
-    "atf4e",
-    "crsts",
-    "luf",
-  ] as Array<keyof FundingSources>;
-
-  // Check for all required values
-  $: errorMessage =
-    pipeline.scheme_type && pipeline.status && pipeline.timescale
+  // Svelte doesn't see nested updates in BudgetForm and TimingForm, so use a
```
dabreegster, line 23: An existing test failed, alerting me to this problem -- yay tests! The fine-grained reactivity for individual fields doesn't bubble up to the parent component, so this is the simplest way I could figure out to plumb updates through

src/lib/sidebar/pipeline/BudgetForm.svelte
dabreegster, line 1: This and the timing form pretty much just copied over

src/lib/sidebar/pipeline/TimingForm.svelte
```diff
@@ -0,0 +1,79 @@
+<script lang="ts">
+  import { NumberInput, Radio } from "lib/govuk";
+  import type { PipelineTiming } from "types";
+
+  export let data: PipelineTiming;
+  // Are status and timescale required?
```
dabreegster, line 6: This is a new difference; the fields aren't required at per-feature level, but are at scheme level

src/types.ts
```diff
@@ -28,7 +28,7 @@ export interface SchemeData {
 // Every field is optional in this type, to match the reality of starting with
 // a blank form. Mandatory fields are marked in the form UI. Optional string
 // types are encoded as "".
-export interface PipelineScheme {
+export interface PipelineScheme extends PipelineBudget, PipelineTiming {
```
dabreegster, line 31: This is a difference from #398 that simplifies our lives massively, by keeping the scheme-level data exactly the same. Moving the fields around would be extra backfill complexity, so just keep them at the top level

tests/data/pipeline_before_feb_fields.geojson
```diff
@@ -0,0 +1,50 @@
+{
+  "type": "FeatureCollection",
+  "features": [
+    {
+      "type": "Feature",
+      "properties": {
+        "scheme_reference": "bb649a68-5fc0-423a-b679-25dc43a8d020",
+        "intervention_type": "other",
+        "pipeline": {
```
dabreegster, line 9: Note lack of new fields. This is produced with the current live pipeline version

dabreegster:
The new test passes individually but fails as the group -- looking into it

dabreegster:
Alright, tests should be fixed now --ready for review

Pete-Y-CS (approved):

Issue #463, opened 2024-02-07
Browse page feedback


dabreegster:
From latest demo session

- [ ] Produce a training video going through the tools and layers (slowly!)
- [ ] Consider LCWIPs as a separate layer, or make it clear most aren't funded
- [ ] Clear all filters button
- [ ] Gradient layer (many notes and more usefulness for this in https://github.com/Urban-Analytics-Technology-Platform/od2net/issues/18)
- [ ] Bus stops
- [ ] Bus frequency (look into NAPTAN data)

dabreegster:
Improvements to gradient:

- [ ] Fix the alignment of labels in the legend
- [ ] Smoothing for short segments
- [ ] Generate an elevation profile from a route. Either select a route from a scheme or the route tool. (This is not easy to implement client-side for the England-wide browse tool)

PR #475, opened 2024-02-20
Add a gradient layer. #463

dabreegster:
Demo at https://acteng.github.io/atip/gradient/browse.html. Enable "Other > Gradient" and zoom in. See https://github.com/acteng/atip-data-prep/pull/55 for details how this layer was generated.

Behavior at low zoom is poor; #326 still pending to rethink most layers there.

The arrows point uphill right now. In a previous project using the same visualization, people have been pretty split over what direction they should point. Any ideas for less ambiguous symbology?


src/lib/browse/layers/lines/Gradients.svelte
```diff
@@ -0,0 +1,101 @@
...
+<VectorTileSource
+  url={`pmtiles://${publicResourceBaseUrl()}/v1/${name}.pmtiles`}
+>
+  <LineLayer
+    {...layerId(name)}
+    sourceLayer={name}
+    manageHoverState
+    paint={{
+      "line-color": makeColorRamp(
+        ["/", ["abs", ["get", "gradient"]], 100],
```
dabreegster, line 66: Note the values of the gradient are integer percents multiplied by 100, to make the pmtiles compression work nicely (loads and loads of floating points do not, and we don't care about extreme precision here). A value of 305 means a 3.05% grade

src/lib/browse/layers/lines/Gradients.svelte
```diff
@@ -0,0 +1,101 @@
...
+  >
+    <Popup let:props>
+      <p>
+        Gradient: <b>{Math.abs(props.gradient) / 100}</b>
+        &percnt;
+      </p>
+    </Popup>
+  </LineLayer>
+
+  <SymbolLayer
```
dabreegster, line 86: We're not using many of these in this codebase yet, but they're quite handy -- maplibre can draw an icon at regular intervals along a line segment, rotate it to match the line, change the density of the icon based on zoom, etc

robinlovelace-ate:
Spot check 1: Buslingthorpe Lane is v. steep and is represented as such here so :+1: on that

image

robinlovelace-ate:
A minor issue I've found with this is that small segments are sometimes out. In the slopes work I did with Rosa we introduced a normalisation factor for segments that were shorter than a given length, e.g. 10 m. The issue is that as you get to shorter lengths the calculation is more sensitive to quirks in the elevation estimation. See example below from the end of Harehills lane. There are 2 parallel lanes here that have pretty much the same gradient yet one is red while the other is green:

image

Could you introduce some kind of 'short segment normalisation'? We did that in slopes package using a lag so that for short links with unusual values we take the rolling mean of the gradient to avoid results being dominated by likely incorrect outliers like this. Example zoomed in on the errant link:

https://acteng.github.io/atip/gradient/browse.html?style=dataviz#20.69/53.8242227/-1.5315835

dabreegster:
> We did that in slopes package using a lag so that for short links with unusual values we take the rolling mean of the gradient to avoid results being dominated by likely incorrect outliers like this

Is this what you mean?
https://github.com/ropensci/slopes/blob/91292bdd860f34deb364943067c9755a7bd97050/R/slopes.R#L75
I'm having trouble understanding this code. If we have a short LineString, is the suggestion to look at the optional LineString attached at each end if it was part of the same OSM way originally?

src/lib/browse/layers/lines/Gradients.svelte
```diff
@@ -0,0 +1,101 @@
...
+    LineLayer,
+    VectorTileSource,
+    SymbolLayer,
+  } from "svelte-maplibre";
+  import { colors, denseLineWidth } from "../../colors";
+  import SequentialLegend from "../SequentialLegend.svelte";
+
+  let name = "gradient";
+  let colorScale = colors.gradient_flat_to_steep;
+  let limits = [0, 3, 5, 8, 10, 20, 100];
```
Pete-Y-CS, line 23: These limits are a bit strange compared to the legend:

For example if you have a gradient of 10.1 it shows you the colour associated with 20 on the legend.
I wonder if we want...


image
Looking at the legend again, 0 - 3 look right: the numbers represent the breakpoints and the legend reflects this. By the time we reach 8 and 10 it looks like the numbers are at the centre of that colour, as if the band of yellow should exist centred on 8, and the band of bright red should be centred on 10. Not sure the best solution here.

Pete-Y-CS: To explain what I mean: looking at that legend I'd expect 15 to be where we switch from bright to slightly dark red

dabreegster: Agreed, this is confusing. https://github.com/acteng/atip/blob/main/src/lib/browse/layers/SequentialLegend.svelte is drawing the legend and just trying to evenly distribute whitespace. But really we want to line up exactly. Any ideas how to do this better?

dabreegster:
Thanks for review! We can improve the legend and handle short segments later.

I'll wait to merge until https://github.com/acteng/atip-data-prep/pull/55 is done and the file is in both GCS buckets too

PR #477, opened 2024-02-21
Move govuk components to another repo

dabreegster:
I moved all the common govuk Svelte components to https://github.com/acteng/govuk-svelte, since we'll want to use these in more projects. No behavioral change here.


package.json
```diff
@@ -47,6 +47,7 @@
     "@turf/nearest-point-on-line": "^6.5.0",
     "@turf/random": "^6.5.0",
     "govuk-frontend": "^5.1.0",
+    "govuk-svelte": "github:acteng/govuk-svelte",
```
dabreegster, line 50: We could publish to NPM, but it might encourage other people to take a dependency on them, despite the warnings in the README. For the moment, I think the git dependency is simpler. When we need to update the components, we commit in the other repo, then `npm i govuk-svelte` here to update the pinning in package-lock.json.

I can publish to NPM if you prefer though

dabreegster:
Rebased. The test failure looks like a flake to me, but will see what the rerun looks like

Pete-Y-CS (approved):

PR #478, opened 2024-02-22
Allow users to deselect the timing and status radio buttons

dabreegster:
By request. Note if the value is required (scheme-level), "Unknown" still produces a warning:
image


dabreegster:
Per https://design-system.service.gov.uk/components/radios/, we're doing something reasonable: "Users cannot go back to having no option selected once they have selected one, without refreshing their browser window. Therefore, you should include ‘None of the above’ or ‘I do not know’ if they are valid options."

Pete-Y-CS (approved):

PR #481, opened 2024-02-28
Upgrade to maplibre 4

dabreegster:
Was waiting on a geocoder dependency to get support, but now we can do this. https://github.com/maplibre/maplibre-gl-js/blob/main/CHANGELOG.md lots of changes, most invisibly handled in svelte-maplibre already

### DIFF
--- a/src/lib/browse/layers/points/CriticalIssues.svelte
+++ b/src/lib/browse/layers/points/CriticalIssues.svelte
@@ -39,19 +39,15 @@
     }
   }
 
-  function clickCluster(e: CustomEvent<LayerClickInfo>) {
-    ($map.getSource(source) as GeoJSONSource).getClusterExpansionZoom(
-      e.detail.features[0].properties!.cluster_id,
-      (err, zoom) => {
-        if (!err) {
-          $map.easeTo({
-            // @ts-ignore We know this is a point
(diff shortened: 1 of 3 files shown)


src/lib/browse/layers/points/CriticalIssues.svelte
```diff
@@ -39,19 +39,15 @@
...
-        if (!err) {
-          $map.easeTo({
-            // @ts-ignore We know this is a point
-            center: e.detail.features[0].geometry.coordinates,
-            zoom: zoom as number,
-          });
-        }
-      },
-    );
+  async function clickCluster(e: CustomEvent<LayerClickInfo>) {
```
dabreegster, line 42: maplibre went through a major internal refactor to use promises instead of callbacks. Nicer APIs as a result :)

Pete-Y-CS: Is this the only place where it changes stuff for us? Wild that this refactor is happening now but looks good!

Pete-Y-CS (approved):

src/lib/browse/layers/points/CriticalIssues.svelte
Pete-Y-CS: I mean I guess it must be but it's surprising

Issue #482, opened 2024-03-06
Traffic Count Improvements


Pete-Y-CS:
A couple of things have been highlighted by inspections that could be improved about traffic count:

- Don't filter only to 2022 (we might want to discuss with folks a different cutoff). If we do this we will want to also display the year the count happened.
- Rescale our colouring, or at least have an inspector mode which colours the dots green, amber, red, CRITICAL COLOUR, depending on whether it's 0-1000, 1000-5000, 5000-10000, or >10000 (probably misremembering the exact bands)
- Use the 'raw counts' from the bottom of the manual count point page: https://roadtraffic.dft.gov.uk/manualcountpoints/47486 to determine the max motor vehicle count for any hour. More than 1000 is a critical issue

dabreegster:
Need to followup with more people about the bands. The route check 21 "motor vehicle volume" is based on per hour, which we don't have yet -- need to scrape or ask DfT for a bulk download. "Provision of crossings" is gated on vpd of < or > 8000, so maybe we have different controls depending on the check being done.

dabreegster:
Ah I'm blind, https://roadtraffic.dft.gov.uk/downloads has a raw count bulk download

PR #485, opened 2024-03-13
Update vehicle counts layer to show all years and emphasize data source

dabreegster:
Just the first task from #482. Before pushing, we need to update the pmtiles file


Pete-Y-CS (approved):
Code looks good, can't see it working because I think it's not gone up to s3 bucket? I'll try locally and if we can sort it out today I'm happy to merge and upload to bucket

dabreegster:
S3 updated now, you can generate yourself or copy https://atip.uk/layers/v1/vehicle_counts.pmtiles to GCS. Please update manifests when done, and I'll merge then, so there's minimal outage time

PR #486, opened 2024-03-14
Fix disapparing images when switching basemaps

dabreegster:
Two commits here. The first is just  refactor, trying to be more precise about the combinatorial explosion of basemap and overlay layers we have. And the second fixes the problem reported recently. Enable cycle parking (which uses an extra image we load), switch basemaps -- now the symbol still works. We just needed https://github.com/dimfeld/svelte-maplibre/blob/master/CHANGELOG.md#082.

Note I still think there might be some z-order oddness when changing basemaps. We decide the layer order when the layers are first created, which'll depend on the basemap layer names that exist at that point. If the user switches basemaps, I don't think we run this calculation again. Harder one to think about for the future.


src/lib/browse/layers/points/CriticalIssues.svelte
```diff
@@ -81,7 +81,7 @@
   }}
 >
   <CircleLayer
-    {...layerId(`${source}-clusters`)}
+    {...layerId(`criticals-clusters`)}
```
dabreegster, line 84: Indirections make it hard to grep code and find where stuff is used

src/lib/maplibre/zorder.ts
```diff
@@ -52,97 +52,110 @@ function getBeforeId(layerId: string): string | undefined {
...
+  browse("criticals-clusters"),
+  browse("criticals-counts"),
+  browse("criticals-points"),
 
-  "draw-split-route",
+  sketch("edit-polygon-fill"),
+  sketch("edit-polygon-lines"),
+  sketch("edit-polygon-vertices"),
 
-  "draw-street-view",
```
dabreegster, line 128: This was unused

Pete-Y-CS (approved):
lgtm

PR #489, opened 2024-03-28
Fix storage quota bug

Pete-Y-CS:


Pete-Y-CS:
Ahh just remembered you wanted us to display the amount of memory used, let me add that now

dabreegster:
And is a playwright test possible?

Pete-Y-CS:
image
image

I tried to add a breaking length string in playwright and it timed out. Maybe there's another way but I think it would be easier with unit rather than functional tests.

dabreegster:
Not sure why a 1MB file is necessary for a test when we can do "x".repeat(123). Also don't think we need to force a browser to deal with 5MB in a form. We can just set some unused local storage key directly. I started a test in https://github.com/acteng/atip/commit/7168babaac920114b1b6ee4c21c30327b0dd10c0. But the window.alerts appear to be quite flaky both in playwright and manually; they mostly haven't appeared for me, which is weird, because I thought window.alert is supposed to be blocking. What if we try a Modal instead and just have this handling in the sketch tool for now?

robinlovelace-ate:
Looks like a tricky one, happy to test locally if you guys want a second pair of eyes on it.

src/lib/sidebar/StorageModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+  function clearLocalStorageAndCloseModal() {
+    clearCurrentSketch();
+    clearLocalStorage();
+    show = false;
+    setStorageError = undefined;
+  }
+</script>
+
+<Modal title="Local Storage Quota Exceeded" bind:open={show}>
+  <p>{setStorageError?.message}</p>
```
dabreegster, line 37: I'd vote for simpler wording:

```
<p>Unable to save changes. Your browser's local storage quota has been exceeded. You can:</p>
<ul>
<li>Use your browser's private or incognito mode</li>
<li>Copy some GeoJSON files to your computer or Sharepoint, then delete them from your browser</li>
</ul>
```

src/lib/sidebar/StorageModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+  <p>
+    If you have saved your work locally to a geojson file recently, you can
+    delete all files stored in the browser for all authorities to free up space.
+    Be sure that you're not losing anything important before doing this!
+  </p>
+  <WarningButton on:click={clearLocalStorageAndCloseModal}>
+    Delete All Sketch Data From Browser
```
dabreegster, line 44: Don't capitalize every word

src/lib/sidebar/StorageModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+  {#if setStorageError}
+    <p>
+      Otherwise here is a breakdown of what is currently stored locally. You can
+      delete individual items (normally storage quota is 5mb or 10mb):
+    </p>
+    {#each setStorageError.storedStrings as storedStringDescriptor}
+      <p>Stored object: {storedStringDescriptor.key}</p>
+      <p>
+        Storage used (rounded to nearest 0.01MB): {storedStringDescriptor.storageUsedInMB
+          .toString()
```
dabreegster, line 55: slice is weird -- `toFixed(2)`?

src/lib/sidebar/StorageModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+  </p>
+  <WarningButton on:click={clearLocalStorageAndCloseModal}>
+    Delete All Sketch Data From Browser
+  </WarningButton>
+  {#if setStorageError}
+    <p>
+      Otherwise here is a breakdown of what is currently stored locally. You can
+      delete individual items (normally storage quota is 5mb or 10mb):
+    </p>
+    {#each setStorageError.storedStrings as storedStringDescriptor}
```
dabreegster, line 51: It could be useful to show these from largest to smallest

src/lib/common/storage.ts
```diff
@@ -0,0 +1,77 @@
...
+        key,
+        storageUsedInMB,
+      };
+    },
+  );
+
+  return results;
+}
+
+function getLengthInMB(text: string) {
```
dabreegster, line 66: Missing return type, and you don't need the intermediate variable. Every intermediate var is something that needs a name and takes mental overhead to understand when reading

src/lib/common/storage.ts
```diff
@@ -0,0 +1,77 @@
...
+        },
+      ];
+    }
+
+    return [false, getEmptySetStorageErrorObject()];
+  }
+}
+
+function getStoredStrings(): StoredStringDescriptor[] {
+  const itemsObject = { ...localStorage };
```
dabreegster, line 51: This does seem to work, but it doesn't seem documented at https://developer.mozilla.org/en-US/docs/Web/API/Storage. Alternatively there's `length` and `key`, like https://github.com/dabreegster/ltn/blob/47187e79fb5bcddc823a6224a1f80d39d624f9b2/web/src/title/TitleMode.svelte#L30

src/lib/sidebar/StorageModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+      <p>Stored object: {storedStringDescriptor.key}</p>
```
dabreegster, line 52: This is repetitive and visually confusing with a bunch of keys. Why not just have a bunch of buttons, keep the key in there, and add ` (123MB)` at the end or something?

src/lib/sidebar/StorageModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+      <p>
+        Storage used (rounded to nearest 0.01MB): {storedStringDescriptor.storageUsedInMB
+          .toString()
+          .slice(0, 4)}MB
+      </p>
+      {#if storedStringDescriptor.key === currentAuthority}
+        <p><WarningIcon text={"warning icon"} />This is the current sketch.</p>
+      {/if}
+      <SecondaryButton on:click={removeStorageItem(storedStringDescriptor.key)}>
```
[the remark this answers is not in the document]
Pete-Y-CS: an obvious trick I'm sure I've learned before that slipped my brain, thanks

src/lib/sidebar/StorageModal.svelte
```diff
@@ -0,0 +1,63 @@
...
+      {#if storedStringDescriptor.key === currentAuthority}
+        <p>
+          <WarningIcon text={"warning icon"} />The button below deletes the
+          current sketch.
+        </p>
+      {/if}
+      <SecondaryButton
+        on:click={() => removeStorageItem(storedStringDescriptor.key)}
+      >
+        Remove stored item for {storedStringDescriptor.key}({storedStringDescriptor.storageUsedInMB}MB)
```
dabreegster, line 59: `.toFixed(2)` on the number would look nicer and I think match the old slice previously

tests/quota.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  ).not.toBeVisible();
+});
+
+test("exceeding local storage quota allows user to clear specific local storage items, including other LAs", async () => {
+  // Get super close to the 5MB limit
+  await page.evaluate(() =>
+    window.localStorage.setItem("huge", "x".repeat(4.999 * 1024 * 1024)),
+  );
+
+  await page.goto("/index.html?schema=pipeline");
```
dabreegster, line 55: Could just go directly to the scheme page with the right query, to speed up the test. We test navigating from the homepage in other test cases, and local storage shouldn't really affect that

src/lib/common/storage.ts
```diff
@@ -0,0 +1,69 @@
+const storageQuotaErrorMessage =
+  "Unable to save because the local storage quota has been exceeded: you may need to clear out your web browser's local storage for this app, or  run in private mode before you can save again. Specific error here: ";
+
+export interface SetStorageErrorObject {
```
dabreegster, line 4: We maybe don't really need this, the empty object method, or `setLocalStorageItem`. The caller in `FileManagement` could call `localStorage.setItem` directly, catch the exception, and trigger the modal when appropriate. We could export the method to summarize current local storage items, and make the modal call it again after clearing/removing an item. But up to you

dabreegster (commented):
Sorted buttons feel much nicer, thanks! Still some open comments about simplifying the wording

src/lib/common/storage.ts
Pete-Y-CS: As mentioned in teams, I think it's plasuible we'd want to share this in the future and I prefer smaller files so I'd prefer to leave it this way.

src/lib/sidebar/StorageModal.svelte
Pete-Y-CS: Yep, did this at first then it got lost in refactor of how we're displaying this! Good spot!

dabreegster (approved):

Issue #493, opened 2024-04-02
Rethink LCWIPs in browse tool


dabreegster:
We should consider making LCWIPs their own layer, not part of the scheme data that's loaded automatically. Some reasons:

- it's semantically very different from the ATF data
- it's growing in size a fair bit
- it has special fields that we should be able to show / filter / color by, so maybe deserving its own controls

and a feature request to open the PDF associated with a scheme. We could upload these to GCS as well, but unsure how much work it'll be to match up schemes to the PDFs

PR #494, opened 2024-04-03
Use gzipped scheme data

dabreegster:
As fate would have it, we're forced to look at #493 immediately, because we're now over the 32MB app engine response limit. If we gzip that file, it's down to about 7MB and loads much faster on slow connections anyway!

Do `gzip --keep all_schemes_output.geojson` after running the combine script, and upload that to the same `private_layers/v1/` folder. Please test in the dev environment first, deploy this PR manually (backend/README has instructions), make sure everthing works end-to-end, then push to test env

### DIFF
--- a/backend/server.js
+++ b/backend/server.js
@@ -44,6 +44,9 @@ app.get("/data/*", async (req, resp) => {
       resp.type("application/geo+json");
     } else if (path.endsWith(".pmtiles")) {
       resp.type("binary/octet-stream");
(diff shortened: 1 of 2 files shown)


Pete-Y-CS (approved):

dabreegster:
I'll be offline until about 2, please do merge as soon as dev env is ready
for it

Pete-Y-CS (approved):

PR #497, opened 2024-04-04
Add sketch source to browse page

Pete-Y-CS:
A quick and dirty fix for filtering by sketch source: maybe useful, certainly useful to have a branch which modelling users can see the delineation!

### DIFF
--- a/src/lib/browse/Filters.svelte
+++ b/src/lib/browse/Filters.svelte
@@ -25,7 +25,7 @@
 
   export let show = true;
 
-  // Dropdown filters;
+  // Dropdown filters
   let authorities: [string, string][] = [];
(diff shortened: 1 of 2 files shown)


dabreegster (commented):
Looks fine, modulo maybe expressing the two values without hyphens and capitalization

src/lib/browse/Filters.svelte
```diff
@@ -25,7 +25,7 @@
 
   export let show = true;
 
-  // Dropdown filters
+  // Dropdown filters;
```
dabreegster, line 28:
```suggestion
  // Dropdown filters
```

dabreegster (approved):

PR #498, opened 2024-04-08
Show LCWIP fields in browse popups

dabreegster:
A start towards #493. The extra LCWIP fields are invisible now; this at least starts showing them in per-feature popups. Also added bolding to values, to try and be consistent with the style for other browse popups.

I don't know if this way of presenting information is useful, but it seems better than not showing anything


Pete-Y-CS (approved):
It looks good. I wonder  if it will  be confusing as is if there are cases where the feature and the scheme have budget and/or timeline data

PR #499, opened 2024-04-09
Infer the v1 intervention_type field in pipeline mode when loading an…

dabreegster:
… external file

### DIFF
--- a/src/lib/sidebar/scheme_data.ts
+++ b/src/lib/sidebar/scheme_data.ts
@@ -22,9 +22,19 @@ export function backfill(json: any): SchemeCollection {
       // We'll set this below when json.schemes is missing
       scheme_reference: "",
       description: "",
-      intervention_type: "other",
       is_coverage_polygon: false,
     };
+
+    // Set a default for this v1 property if it's missing. The field can't be
+    // edited by the user with the pipeline schema.
+    if (!f.properties.intervention_type) {
+      // Guess based on geometry
+      f.properties.intervention_type =
+        new Map([
+          ["LineString", "route"],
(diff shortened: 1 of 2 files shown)


src/lib/sidebar/scheme_data.ts
```diff
@@ -22,10 +22,20 @@ export function backfill(json: any): SchemeCollection {
...
       scheme_reference: "",
       description: "",
-      intervention_type: "other",
       is_coverage_polygon: false,
     };
 
+    // Set a default for this v1 property if it's missing. The field can't be
+    // edited by the user with the pipeline schema.
+    if (!f.properties.intervention_type) {
+      // Guess based on geometry
```
dabreegster, line 31: Same as what atip-scheme-data does anyway

Pete-Y-CS (approved):

Pete-Y-CS:
Changes look good but I've just noticed some tests are breaking. I wonder how that's been introduced or if it's some flakiness? Although we don't see much flakiness on GH these days

dabreegster:
Hmm, I also saw some timeout flakiness running locally. I'll look into this a bit.

dabreegster:
While looking into flakiness, there were many legitimate test failures in `savefiles`. Rethinking things, the previous fix is confusing in v1 mode, because it hides the problem of not selecting an intervention type, even though the user can control it. So I've changed the fix to only work in pipeline mode, where this is no radio button for that field.

Running individual tests locally is fine with low parallelism. Let's see what a second round of GH actions looks like. They only use 2 workers, according to logs

PR #502, opened 2024-05-13
Show both sides of the road for pavement width

dabreegster:
Needs https://github.com/acteng/atip-data-prep/pull/60 to be uploaded first


dabreegster:
Have the new files been uploaded?

dabreegster:
It looks like yes, so merging. Pavement layer is broken otherwise. Either we upload a new file and merge the frontend at about the same time to prevent this, or use versioned directories.

Pete-Y-CS:
Sorry didn't see the question, was waiting for my long upload home internet and forgot to check back in! Yes uploaded!

PR #503, opened 2024-05-14
Use the TS RouteTool from the upstreamed NPM package

dabreegster:
I've refactored more drawing code to upstream packages. Local testing works fine, but will let playwright run in GH CI


package.json
```diff
@@ -54,8 +54,14 @@
     "maplibre-gl": "^4.0.2",
     "read-excel-file": "^5.7.1",
     "route-snapper": "^0.3.0",
+    "route-snapper-ts": "^0.0.3",
     "svelte": "^4.2.10",
     "svelte-maplibre": "^0.9.2",
     "uuid": "^9.0.1"
+  },
+  "overrides": {
```
dabreegster, line 62: There's some slight trickery here -- the new -ts package depends on the latest 0.4.0 of the WASM package, but that has a breaking change to the binary file format. We don't have any need of that new version yet, and there's no reason to regenerate all the files right now.

PR #506, opened 2024-05-16
Indicate the side of road for pavement width

dabreegster:
Needs the new pmtiles, of course

### DIFF
--- a/src/lib/browse/layers/lines/PavementWidths.svelte
+++ b/src/lib/browse/layers/lines/PavementWidths.svelte
@@ -20,6 +20,32 @@
   let limits = [0, 2, 4, 6, 8, 13];
 
   let show = false;
+
+  function direction(angle: number): string {
+    // 0 means north and the orientation is weird; verify visually
+    if (angle < 45 || angle > 315) {
+      return "N";
+    } else if (angle < 90) {
+      return "NE";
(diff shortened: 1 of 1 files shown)


src/lib/browse/layers/lines/PavementWidths.svelte
```diff
@@ -82,27 +108,33 @@
       visibility: show ? "visible" : "none",
     }}
   >
-    <Popup let:props>
+    <Popup let:props openOn="hover">
       {#if props.left_average}
```
Pete-Y-CS, line 112: My overzealous DRY instincts are kicking in, is it overkill to turn this into its own component rather than having mostly the same code for both left and right?

Pete-Y-CS (approved):
Fine and certainly makes things better but a small comment about maybe turning the left/right parts of the popup into a component to be more DRY. Not a blocker but a maybe-nice-to-have

Pete-Y-CS:
new pmtiles uploaded

dabreegster:
Refactored the popups, but noticed the directions are wrong. There are 8 directions, 360/8 is 45, but north shouldn't be 0 to 45, it should be split to handle 360 - 45/2, and so on. Does that make sense? It's late, can pick up next week

PR #507, opened 2024-05-20
Move route split logic to upstream TS package

dabreegster:
I moved the core split logic upstream. This passes manual tests. (We don't have any playwright tests of splitting. Eventually there ought to be unit tests just for the split logic, but not in this repo.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,6 @@
     "@turf/boolean-point-in-polygon": "^6.5.0",
     "@turf/helpers": "^6.5.0",
     "@turf/length": "^6.5.0",
-    "@turf/line-slice": "^6.5.0",
-    "@turf/line-split": "^6.5.0",
     "@turf/mask": "^6.5.0",
     "@turf/nearest-point-on-line": "^6.5.0",
     "@turf/random": "^6.5.0",
@@ -54,7 +52,7 @@
     "maplibre-gl": "^4.0.2",
     "read-excel-file": "^5.7.1",
     "route-snapper": "^0.3.0",
-    "route-snapper-ts": "^0.0.3",
+    "route-snapper-ts": "^0.0.5",
     "svelte": "^4.2.10",
     "svelte-maplibre": "^0.9.2",
     "uuid": "^9.0.1"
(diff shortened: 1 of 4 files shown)


src/lib/draw/EditGeometryMode.svelte
```diff
@@ -40,14 +40,16 @@
...
+        feature as unknown as Feature<LineString, RouteProps>,
+      );
       $routeTool?.addEventListenerSuccess(onSuccess);
       $routeTool?.addEventListenerUpdated(onUpdate);
       $routeTool?.addEventListenerFailure(onFailure);
       controls = "route";
     } else if (feature.geometry.type == "Polygon") {
       if (feature.properties.waypoints) {
-        $routeTool?.editExistingArea(feature as Feature<Polygon, Props>);
+        $routeTool?.editExistingArea(feature as Feature<Polygon, AreaProps>);
```
dabreegster, line 52: We're not so precise about our properties in this repo's types, so the compiler gets confused because geometry type being polygon doesn't guarantee we have all the AreaProps

package.json
```diff
@@ -41,8 +41,6 @@
     "@turf/boolean-point-in-polygon": "^6.5.0",
     "@turf/helpers": "^6.5.0",
     "@turf/length": "^6.5.0",
-    "@turf/line-slice": "^6.5.0",
```
robinlovelace-ate, line 44: Would be interested in issues with turf implementation, impressive to see this now in `route-snapper-ts`, above my skill level but general :+1: to all this!

robinlovelace-ate (approved):

src/lib/draw/route/SplitRouteMode.svelte
```diff
@@ -86,56 +84,41 @@
...
+        gj.features.splice(
+          snappedIndex!,
+          1,
+          // TODO Maybe splitRoute's types can be more precise. Existing
+          // properties get copied, so this is valid.
+          piece1 as unknown as OurFeature<LineString>,
+          piece2 as unknown as OurFeature<LineString>,
+        );
+
+        return gj;
```
Pete-Y-CS, line 115: does returning the `gj` do anything? Looks to me like it may as well be `return;` since we're editing in place.

Pete-Y-CS (approved):

src/lib/draw/route/SplitRouteMode.svelte
dabreegster: https://svelte.dev/docs/svelte-store#writable
It's necessary. `update is a method that takes one argument which is a callback. The callback takes the existing store value as its argument and returns the new value to be set to the store.`

Pete-Y-CS: ahhh yes didn't spot we were in a callback cheers!
LGTM

PR #508, opened 2024-06-05
Capture browse layer state in the URL

dabreegster:
Demo: https://acteng.github.io/atip/browse_layer_urls/browse.html?style=dataviz&crossings=1&stats19=other%2F2019%2F2022&road_noise=1#5.69/53.021/-1.825

This syncs up the state of browse layers with URL query parameters. This makes URLs much more reproducible/shareable, and is a step towards linking directly to the browse page (or embedding in an iframe?) from route check.

Note:
- If someone changes the URL, there's a full page refresh. We could squeeze everything into the URL fragment if we wanted to change this, but I don't think there's a use case for it
- No change for the critical issues layer, because that requires a custom file input
- We can't use a single query param (like `layers=x,y,z`) because some layers have more details
- The URL as it is now isn't quite reproducible, because the camera viewport gets ignored. That was a choice made because of a bug in svelte-maplibre that's now fixed, so I'll followup separately and fix on all pages

### DIFF
--- a/src/lib/browse/layers/areas/CombinedAuthorities.svelte
+++ b/src/lib/browse/layers/areas/CombinedAuthorities.svelte
@@ -17,11 +17,12 @@
   } from "svelte-maplibre";
   import { colors } from "../../colors";
   import OsOglLicense from "../OsOglLicense.svelte";
+  import { showHideLayer } from "../url";
 
   let name = "combined_authorities";
   let color = colors.combined_authorities;
 
-  let show = false;
+  let show = showHideLayer(name);
 
   function onClick(e: CustomEvent<LayerClickInfo>) {
     window.open(
@@ -33,7 +34,7 @@
   }
 </script>
 
-<Checkbox bind:checked={show}>
+<Checkbox bind:checked={$show}>
   <ColorLegend {color} />
   Combined authorities
   <span slot="right">
@@ -58,7 +59,7 @@
       "fill-opacity": hoverStateFilter(0.0, 0.5),
     }}
     layout={{
-      visibility: show ? "visible" : "none",
+      visibility: $show ? "visible" : "none",
     }}
     manageHoverState
     eventsIfTopMost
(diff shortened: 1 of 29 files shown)


src/lib/browse/layers/areas/CombinedAuthorities.svelte
dabreegster, line 1: Example of a nice and simple change

src/lib/browse/layers/areas/Pollution.svelte
```diff
@@ -4,10 +4,30 @@
   import { layerId } from "lib/maplibre";
   import { RasterLayer, RasterTileSource } from "svelte-maplibre";
   import OsOglLicense from "../OsOglLicense.svelte";
+  import { customUrlState } from "../url";
 
-  let show = false;
-  let opacity = 50;
-  let pollutant = "PM25_viridis";
+  type State = {
```
dabreegster, line 9: Example of a more complex change. Any layer with more controls beyond show/hide has something like this. It's verbose, but I don't think it can really be much less code than this, without trying to encode as some very awkward JSON strings.

src/lib/browse/layers/url.ts
```diff
@@ -0,0 +1,45 @@
...
+export function customUrlState<T>(
+  name: string,
+  defaultValue: T,
+  stringify: (state: T) => string | null,
+  parse: (param: string) => T,
+): Writable<T> {
+  let param = new URLSearchParams(window.location.search).get(name);
+  let initialValue = param == null ? defaultValue : parse(param);
+  let store = writable(initialValue);
+  // TODO How do we avoid leaking this?
```
dabreegster, line 33: In prod, this won't be a problem -- each layer component is created only once, so there'll only be one store for the lifetime of the app. In dev mode with HMR, this will leak memory, because I couldn't figure out a clean way to call unsubscribe / teardown the stores. We could add onDestroy handlers to every single component I guess, but seems like overkill for a dev-mode mostly-invisible problem

Pete-Y-CS (approved):
Very cool feature! I didn't fully understand the memory leak but if it only exists for dev users if you don't refresh for ages I think that's absolutely fine

dabreegster:
Thanks for the review

>  I didn't fully understand the memory leak

https://learn.svelte.dev/tutorial/auto-subscriptions

PR #510, opened 2024-06-10
Replace the old critical issues layer with a new automtic layer

dabreegster:
Instead of manually loading an xlsx file, automatically download when logged in. The contents of the file has also changed.

This is a simple first version to get internal feedback. Many styling improvements to follow.

To test locally, you need to generate the layer, symlink it up to `public/private_layers/v1/problems.geojson.gz`, then do `VITE_RESOURCE_BASE="" VITE_MIMIC_GCP_LOCALLY="true" npm run dev`


src/lib/browse/layers/points/Problems.svelte
```diff
@@ -0,0 +1,114 @@
...
+    ["No Resolution Planned", "black"],
+  ];
+</script>
+
+<Checkbox bind:checked={$show}>
+  Issues & Conflicts
+  <span slot="right">
+    <HelpButton>
+      <p>
+        Critical issues and policy conflicts from the DA Log. This is internally
```
Pete-Y-CS, line 51: maybe call it Design Assistance Log just for absolute clarity, I can think of a few possible reasonable answers (Assurance, Aid, Assessment)

Pete-Y-CS (approved):
apart from a small wording gripe looks great

dabreegster:
Thanks, done. Could you look at https://github.com/acteng/atip-data-prep/pull/62 and upload the file please?

PR #512, opened 2024-06-27
Start refactoring scheme sketcher code, before moving to a library

dabreegster:
I want to move the bulk of the scheme sketcher code into a new git repo, exposed as a Sveltekit library. The goal is to more easily use pieces of this from other projects, such as the inspectorate tools. It's not an easy refactor; the bulk of the core drawing logic is already in `maplibre-draw-polygon` and `route-snapper`, but the code here manages the toolbox, switches modes, manages the store with the GeoJSON features, etc.

Roughly the split of responsibility I picture is like this. The sketch library will handle `lib/draw/` and parts of `lib/sidebar` -- giving people the mode switching logic, multi layer (scheme) management, sidebar controls, etc. What will remain in this repo (and have a new, much simpler variation in inspectorate tools) is per-schema forms, backfilling logic, etc. So the library shouldn't really care about feature properties beyond a few of the truly common ones.

This PR is a first step that makes `lib/draw/` much more self-contained and starts teasing apart the types. I think even as a standalone refactor without doing the rest, it has some value. It's not complete -- the code still pulls in repo-wide `map` and `userSetings` stores, and I need to think through how to make the `SchemeCollection` type more generic. Assuming tests pass and there are no comments, I'll merge next time I work on this. A next step will probably be starting the new repo and discovering more tricky complications.

### DIFF
--- a/src/lib/draw/EditGeometryMode.svelte
+++ b/src/lib/draw/EditGeometryMode.svelte
@@ -10,7 +10,6 @@
   import { ButtonGroup, DefaultButton, SecondaryButton } from "govuk-svelte";
   import type { FeatureWithProps } from "lib/maplibre";
   import { cfg } from "./config";
-  import { schema } from "stores";
   import { onDestroy, onMount } from "svelte";
   import type { FeatureUnion } from "types";
   import PointControls from "./point/PointControls.svelte";
@@ -129,17 +128,7 @@
     if (source.properties.waypoints) {
       destination.properties.waypoints = source.properties.waypoints;
     }
-    // Only copy route_name if the user hasn't set it. It's not simple to
-    // distinguish the user manually editing the name from it being auto-filled
-    // previously, so be safe and don't overwrite anything. The user can always
-    // use the auto-fill button explicitly.
-    if (
-      source.properties.route_name &&
-      !destination.properties.name &&
-      $schema != "pipeline"
-    ) {
-      destination.properties.name = source.properties.route_name;
-    }
+    cfg.updateFeature(destination, source);
   }
 
   function finish() {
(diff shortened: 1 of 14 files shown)


src/lib/draw/EditGeometryMode.svelte
```diff
@@ -129,17 +126,7 @@
     if (source.properties.waypoints) {
       destination.properties.waypoints = source.properties.waypoints;
     }
```
dabreegster, line 128: The stuff staying here isn't domain specific; it comes from the route tool

src/lib/draw/config.ts
```diff
@@ -0,0 +1,83 @@
...
+// override.
+
+// TODO As an alternate idea, users could implement a custom Svelte store with methods for doing these things
+
+export let cfg = {
+  interventionName: (f: FeatureWithAnyProps) => {
+    return interventionName(f as FeatureUnion);
+  },
+
+  newPointFeature: (f: Feature<Point>) => {
```
dabreegster, line 24: The theme here is that these functions get a "mostly just geometry" object and then make it obey the invariants imposed by `gjSchemeCollection`. Some of the logic is common,but stuff like `intervention_type` and fiddling with route names is not.

src/lib/draw/types.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+  | { mode: "new-point" }
+  | { mode: "new-freehand-polygon" }
+  | { mode: "new-snapped-polygon" }
+  | { mode: "new-route" }
+  | { mode: "split-route" }
+  | { mode: "set-image" }
+  | { mode: "streetview" };
+
+// Properties are guaranteed to exist, but can be anything, so TS is mostly disabled
```
dabreegster, line 23: I renamed this to be more clear. By setting the properties this way, TS assumes any key can exist, with any type. (It's what we've been doing before; there's no change, just the name makes it more clear)

robinlovelace-ate:
This sounds great to me and should make it easier for others to use the very useful functionality, in UK and beyond!

PR #515, opened 2024-07-14
Enormous refactor to use a generic scheme-sketcher-lib

dabreegster:
I've been doing some work to make the scheme sketcher code more generic. https://github.com/acteng/scheme-sketcher-lib is the result thus far. It handles the drawing tools, the mode switching between them, organizing features into layers (schemes), and the state management. Users of this ss-lib, like our scheme sketcher here, are mainly responsible for defining the per-feature and per-scheme forms / properties. The intent of this refactor is to make it easy to develop new SS variants that need to define their own forms, and to use some pieces directly in inspectorate tools for sketching scheme summaries. (The requirements there get a bit more interesting, because the auto-save and multiple layers aren't desirable.)

I've already copied/adapted code in https://github.com/acteng/scheme-sketcher-lib and made an ultra minimal example SS app in https://github.com/acteng/scheme-sketcher-lib/tree/main/src/routes. This PR cuts over the "ATIP scheme sketcher" (really trying to be non-confusing with naming here) to use the scheme-sketcher-lib. I'll make note of design choices in a code review below. It's an enormous diff; more than anything, it'd be useful if you can run locally and help test things (in both browse and sketch pages) to make sure there's no behavioral differences.)


dabreegster (commented):
Big change number something: for the remaining code, I've reorganized much of stuff in `src/lib/sketch`, to be less confusing with browse code

The only behavioral change I've found so far is the loss of special styling of coverage polygons. There are only 4 of them in data we've collected so far; I don't think it's too important to keep this around right now. It's tricky to fit into the generic `InterventionLayer` without making it have lots of special casing; if we need this behavior, I'll probably just copy the library file here and have the coverage stuff here too

robinlovelace-ate (commented):
Great work Dustin :fire:

dabreegster:
I have high confidence in the changes here after testing for a few days, so I'm going to merge

PR #518, opened 2024-08-22
Add a header to the app

Pete-Y-CS:


Pete-Y-CS:
hmm will investigate test failures

Pete-Y-CS:
tests fixed 🙌🏻

dabreegster:
Cool to see this! In Figma, the banner is just covering the left sidebar, not the full screen:
image
I'll work on making that change

src/style/main.sass
```diff
@@ -18,3 +18,11 @@
     @extend .govuk-link
   ul
     @extend .govuk-list, .govuk-list--bullet
+
+body
+  margin-top: 0
+
+.below-header
```
dabreegster, line 25: The SASS file is not the correct place for one-off classes that only make sense in one component. Svelte's scoped CSS is very nice, since it lets us keep styles in the same component where they're relevant

dabreegster:
I matched Figma and simplified the code. Trying to share the same CSS for all 3 pages doesn't quite work. There's more changes in Figma like the service banner:
image
and I have some questions about whether to display all of this in every mode of SS (as you activate a drawing tool, for instance), but we can wait to get designs for that.

dabreegster:
image

Also a bit odd to cover up part of the banner with the collapse tool, but I'm guessing that control will be redesigned soon anyway

PR #522, opened 2024-10-09
Add buttons to switch to browser from sketcher

Pete-Y-CS:
Moving the navigation logic outside of the components so that they can be reused. Seems overkill right now maybe but I think we might want a more complex version of open browse page at which point it being in its own file makes a bit more sense. Will have to have a bit of a general cleanup to move all nav stuff to this file

### DIFF
--- a/src/lib/common/navigation.ts
+++ b/src/lib/common/navigation.ts
@@ -1,10 +1,10 @@
 import { schema as schemaStore } from "../../stores";
-import { get } from "svelte/store"
+import { get } from "svelte/store";
 
 export function openBrowsePage() {
-    window.location.href = "browse.html";
+  window.location.href = "browse.html";
 }
 
 export function startSketching(authorityName: string) {
-    window.location.href = `scheme.html?authority=${authorityName}&schema=${get(schemaStore)}`;
+  window.location.href = `scheme.html?authority=${authorityName}&schema=${get(schemaStore)}`;
 }
--- a/src/lib/common/navigation.ts
+++ b/src/lib/common/navigation.ts
@@ -8,3 +8,7 @@ export function openBrowsePage() {
 export function startSketching(authorityName: string) {
   window.location.href = `scheme.html?authority=${authorityName}&schema=${get(schemaStore)}`;
 }
+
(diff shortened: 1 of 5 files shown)


src/lib/common/navigation.ts
```diff
@@ -0,0 +1,14 @@
+import { schema as schemaStore } from "../../stores";
+import { get } from "svelte/store";
+
+export function openBrowsePage() {
+  window.location.href = "browse.html";
+}
+
+export function startSketching(authorityName: string) {
```
dabreegster, line 8: I think this should take a parameter for the schema. The schema store doesn't make any sense for the browse page; it really should be refactored to something more clearly sketch specific. For this PR I'd just vote to make it a param

src/pages/ChooseArea.svelte
```diff
@@ -128,9 +124,14 @@
 
       <h1>Scheme Sketcher</h1>
 
-      <SecondaryButton on:click={() => (showAbout = !showAbout)}>
-        About
-      </SecondaryButton>
+      <div class="govuk-button-group">
```
dabreegster, line 127: We have a `ButtonGroup` in govuk-svelte

dabreegster (changes requested):
LGTM except for these two changes, and any feedback from Jadene about the button on browse

src/lib/common/navigation.ts
Pete-Y-CS: So do you think this should be an optional param? e.g. if a schema is provided: use that, otherwise, use what's in the store

Pete-Y-CS: I agree the browse page shouldn't know about schema. Currently the browse page doesn't use this function but I think it should eventually use it for 'edit a local version of the this scheme', or at least use a function from this module. I suppose in that instance you can find out from the scheme you're trying to edit what schema you want.

dabreegster: I think it should be a required parameter. So yes, in that second case, it would check if `pipeline` is in the scheme object and decide schema from that

PR #528, opened 2024-10-22
rename old deploy files and add public dev deploy workflow

Pete-Y-CS:


Pete-Y-CS:
Service request that I think should allow us to deploy has been fulfilled

Pete-Y-CS:
> There's other steps in `backend/README` to fill out the GCS buckets and point to them, but guessing that's coming later

ahh good shout. Bucket won't be ready yet but I'll make sure we're using the right project number to deploy in the yaml file used

Issue #526, opened 2024-10-21
Update browse layers


dabreegster:
Some browse layers could use an update by now.

- [x] Everything OSM based: trip generators, cycle paths, cycle parking, crossings, bus routes, trams
- [x] CAs and LADs. In particular, there are new LADs, because Cumbria has been split
- [x] Looks like 2023 stats19 data is available now

PR #529, opened 2024-10-22
Update LADs and CAs. #526

dabreegster:
Compare https://acteng.github.io/atip/ladsladslads_andcas/browse.html with the internal page. New CAs are clearly visible. I'm actually still hunting around to find the changes between January and May 2024 for LADs; Cumbria was already split by January.

Pete, you'll need to copy two new files to the GCS bucket: http://atip.uk/layers/v1/local_authority_districts.geojson and http://atip.uk/layers/v1/combined_authorities.geojson, and update manifest


Pete-Y-CS (approved):

Pete-Y-CS:
upload done, WSL misbehaving so reboot then I'll do manifest

PR #531, opened 2024-10-22
don't always require IAP when deploying to GCP

Pete-Y-CS:


backend/server.js
```diff
@@ -17,7 +17,9 @@ let expectedAudience = `/projects/${process.env.PROJECT_NUMBER}/apps/${process.e
 
 let app = express();
 
-app.use(checkIap);
+if(process.env.USE_IAP === "true") {
```
dabreegster, line 20: It'd be safer to flip this and check for != false. If something gets messed up with env variables, we always want to default to using IAP

dabreegster (changes requested):

backend/server.js
Pete-Y-CS: great shout. Will do.

dabreegster (approved):

PR #533, opened 2024-10-23
Serve route snapper files through the GAE proxy

dabreegster:
Untested!


Pete-Y-CS (approved):
Looks good, and I've tested in the dev env!

PR #535, opened 2024-10-29
Redesign scheme management

dabreegster:
Counterpart to https://github.com/acteng/scheme-sketcher-lib/pull/6

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "humanize-string": "^3.0.0",
     "js-cookie": "^3.0.5",
     "maplibre-gl": "^4.0.2",
-    "scheme-sketcher-lib": "github:acteng/scheme-sketcher-lib#redesign_schemes",
+    "scheme-sketcher-lib": "github:acteng/scheme-sketcher-lib",
     "svelte": "^4.2.10",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "humanize-string": "^3.0.0",
     "js-cookie": "^3.0.5",
     "maplibre-gl": "^4.0.2",
-    "scheme-sketcher-lib": "github:acteng/scheme-sketcher-lib",
+    "scheme-sketcher-lib": "github:acteng/scheme-sketcher-lib#redesign_schemes",
```
dabreegster, line 50: I'll merge the other PR and update this

dabreegster:
OK, this should all be ready to go. The only behavioral change is that per-scheme error messages are no longer shown upfront on the sidebar, only in the edit scheme modal. I've fixed the tests accordingly. We could preserve this behavior with some effort (and decisions where the error text should go), but I feel like it's not that important to worry about right now

package.json
dabreegster: Done

Pete-Y-CS (approved):

Issue #536, opened 2024-11-04
Scheme sketch changes pre-launch


dabreegster:
Each medium-largeish PR updating the design has a few leftover tasks, going to track them all here for simplicity.

- [ ] Check the geocoder position
- [ ] Decide where the Delete button should go for editing a feature
- [ ] Possibly have a second copy of the finish/cancel/delete buttons in the left sidebar, floating/sticky at the top or bottom
- [ ]  Use a marker for point mode, so editing an existing point is more normal
- [ ] Merge the freehand / snapped polygon tools
- [ ] Clean up the govuk-svelte components a bit; leftLabel and noBottomMargin may not be the right approach after all
- [ ] Rewrite the About popup for SS
- [ ] Write an About popup for SB
- [ ] Check names of all pages (including HTML titles), and scrub out "ATIP" generally

Issue #488, opened 2024-03-26
Handle local storage quota errors better


dabreegster:
When we `setItem`, quota could be exceeded. At the very least, have a real error show up. Maybe auto-download the current file and instruct the user how to clear their cache. Or we could have an option to do that for them -- display the list of items in local storage, let them confirm and clear all to free up space.

We could also consider gzipping before storing (with a runtime penalty)

dabreegster:
Ideas:

- first write a test to mimic running out of storage. Could use playwright to set some massive key slightly less than the limit to start, instead of trying to make up fake scheme data
- Catch the error with setItem. At absolute minimum, window.alert and explain the problem better
- If we can reasonably make a Modal appear when the problem happens, then having buttons to clear all local storage data or... even better, list out the keys, show the sizes, and let people delete each one to reclaim space
- We could also try compressing everything before we put it in local storage. I'd first check the runtime penalty to reading and writing. And then think through the backwards compatibility carefully; would definitely need some thorough tests before going this route

PR #539, opened 2024-11-06
Proper file management for scheme sketch

dabreegster:
Previously, SS only handled one local storage file per (authority, schema) pair. This introduces proper multi-file management. The user flow is now spread out across more pages:

1) choose an area
2) create, import, or load a file
3) sketch

To handle local storage issues that could potentially be triggered on any page (#488), this PR also introduces a simpler approach of detecting the quota problem, then redirecting to a new dedicated page for helping the user resolve the problem.

This PR tries to match Figma designs as much as possible. The list below is the MVP for this PR, and we can tackle smaller followups in #536:

- [x] Adapt existing tests to the new flow


tests/homepage.spec.ts
```diff
@@ -31,18 +33,12 @@ test("schema is plumbed along to the sketch page", async ({ page }) => {
     .fill("LAD_Adur");
   await page.getByRole("button", { name: "Start" }).click();
   await expect(page).toHaveURL(
-    /.*scheme.html\?authority=LAD_Adur&schema=pipeline/,
+    /.*files.html\?authority=LAD_Adur&schema=pipeline/,
   );
-});
 
-test("a v1 file with a pipeline hint redirects to pipeline mode", async ({
```
dabreegster, line 38: This behavior is removed; people can create a new file and add scheme from file instead

dabreegster:
OK, this is done! Tests have been overhauled -- they're now much more isolated from each other, with most starting a new file per test case. Behaviorally, the most unusual thing is how importing a file works. Unless it looks like a pipeline file, it'll get detected as the v1 schema. If a user needs to change these, they need to create an empty pipeline file and do "add schemes from file" instead. In practice, this won't be a problem. If we make a new schema for ATF5, we can revisit how the user specifies the type of file they create to make this more clear.

I've combed through the code changes carefully and made sure tests all make sense, and tried this for real against all the random sketch files in different areas I've got. I'm going to merge now, async review is welcome, and we have #536 to track followups.

PR #541, opened 2024-11-15
Fluidly link sketch and browse pages

dabreegster:
This PR adapts the ideas in https://github.com/acteng/atip/tree/show-local-sketches-browse-page. The browse page has new controls to import all local sketches as new layers. The user can then jump to the sketch tool and edit one of those sketches.

Demo: https://acteng.github.io/atip/ss_sb_integration

Design decisions I'm not entirely sure about:
- The phrasing generally. "Your ATF/LCWIP schemes"? I started to use "local", but that word is very overloaded.
- Should we have layers for both "your ATF" and "your LCWIP", or just one "your sketches" layer? The reasons for splitting them for the main data may not make as much sense for the (presumably small) amount of local sketches.
- Can the filters be cleaned up for local ATF data? We won't ever have funding programme or milestone info, because there's no join with external data for local sketches.
- When editing a file from browse, if it comes from the main data, you edit a _copy_. If it comes from locally, you edit the local file. That local file might have multiple schemes, even though you only selected one from browse.

### DIFF
--- a/src/lib/browse/schemes/SchemesLayer.svelte
+++ b/src/lib/browse/schemes/SchemesLayer.svelte
@@ -5,10 +5,11 @@
     FileInput,
     Select,
     CheckboxGroup,
+    SecondaryButton,
   } from "govuk-svelte";
   import { appVersion, Legend, WarningIcon } from "lib/common";
   import LoadRemoteSchemeData from "./LoadRemoteSchemeData.svelte";
-  import { setupSchemes } from "./data";
+  import { importAllLocalSketches, setupSchemes } from "./data";
   import Filters from "./Filters.svelte";
   import {
     atfSchemesGj,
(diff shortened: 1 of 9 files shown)


src/lib/browse/schemes/SchemesLayer.svelte
```diff
@@ -30,29 +35,47 @@
 
   let errorMessage = "";
 
-  let atfName = "atf_schemes";
-  let atfTitle = "ATF schemes";
-  let atfShow = showHideLayer(atfName);
-  let atfStyle = "fundingProgramme";
-  $: [atfColor, atfLegend] = pickStyle(atfStyle);
+  let mainAtfName = "main_atf_schemes";
```
dabreegster, line 38: There's some boilerplate copying here. We could consider abstracting it a bit, but it has to be done carefully to be useful (like bundling all these variables together, leaving room to customize names/labels, etc). I'd like to decide about having one local layer vs local ATF + local LCWIP before doing anything though.

PR #543, opened 2024-11-19
Simplify the SS sketches shown as a SB layer -- one layer, not split …

dabreegster:
…by ATF/LCWIP.

#536, and following up to #541

Demo: https://acteng.github.io/atip/sb_ss_one_layer/browse.html

The filters for local schemes are simple:
image

### DIFF
--- a/src/lib/browse/schemes/SchemesLayer.svelte
+++ b/src/lib/browse/schemes/SchemesLayer.svelte
@@ -12,18 +12,15 @@
   import { importAllLocalSketches, setupSchemes } from "./data";
   import Filters from "./Filters.svelte";
   import {
-    mainAtfSchemes,
-    mainLcwipSchemes,
-    filterMainAtfSchemeText,
-    filterMainAtfInterventionText,
-    filterMainLcwipSchemeText,
-    filterMainLcwipInterventionText,
-    localAtfSchemes,
-    localLcwipSchemes,
-    filterLocalAtfSchemeText,
-    filterLocalAtfInterventionText,
-    filterLocalLcwipSchemeText,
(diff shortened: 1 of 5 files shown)


src/lib/browse/schemes/SchemesLayer.svelte
dabreegster, line 1: Mostly just some careful renaming here

src/lib/browse/schemes/SchemesLayer.svelte
```diff
@@ -35,45 +32,37 @@
...
-
-  let localLcwipName = "local_lcwip_schemes";
-  let localLcwipTitle = "Your LCWIP schemes";
-  let localLcwipShow = showHideLayer(localLcwipName);
-  let localLcwipStyle = "interventionType";
-  $: [localLcwipColor, localLcwipLegend] = pickStyle(localLcwipStyle);
+  let localName = "local_schemes";
+  let localTitle = "Your schemes";
+  let localShow = showHideLayer(localName);
+  $: [localColor, localLegend] = pickStyle("interventionType");
```
Pete-Y-CS, line 50: Wondering about local legend: v2 will want to be rebased onto this and we may want to rethink the legend

Pete-Y-CS (commented):
Looks good, just one comment that we might wanna rework this after changing to v2?

src/lib/browse/schemes/SchemesLayer.svelte
dabreegster: Yep, let's figure that out in #544. There's more changes to browse needed for v2, like updating the intervention popup

PR #545, opened 2024-11-20
Update README and workflow

Pete-Y-CS:


backend/app.yaml
dabreegster, line 1: I'm a little confused what env this file is for. Maybe we should rename this and not have any default app.yaml?

dabreegster (approved):

PR #548, opened 2024-11-25
Add accessibility statement and begin to scrub ATIP from the site

Pete-Y-CS:
Will finish scrubbing ATIP in another PR

### DIFF
--- a/accessibility.html
+++ b/accessibility.html
@@ -6,7 +6,7 @@
       name="viewport"
       content="initial-scale=1,maximum-scale=1,user-scalable=no"
     />
-    <title>Plan Your Active Travel Scheme</title>
+    <title>Plan Your Active Travel Schemes</title>
   </head>
   <body>
     <div id="app"></div>
--- a/accessibility.html
+++ b/accessibility.html
@@ -6,7 +6,7 @@
       name="viewport"
       content="initial-scale=1,maximum-scale=1,user-scalable=no"
     />
-    <title>Plan Your Active Travel Schemes</title>
+    <title>Plan Your Active Travel Schemes - Accessibility Statement</title>
   </head>
   <body>
     <div id="app"></div>
--- a/accessibility.html
+++ b/accessibility.html
@@ -11,7 +11,7 @@
   <body>
     <div id="app"></div>
     <script type="module">
-      import ChooseArea from "./src/pages/ChooseArea.svelte";
+      import AccessibilityStatement from "./src/pages/AccessibilityStatement.svelte";
 
       // For govuk setup. For some reason, the initAll step doesn't work here.
       document.body.className +=
@@ -20,7 +20,7 @@
           ? " govuk-frontend-supported"
           : "");
 
-      new ChooseArea({
+      new AccessibilityStatement({
         target: document.getElementById("app"),
       });
     </script>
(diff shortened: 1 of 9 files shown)


accessibility.html
```diff
@@ -0,0 +1,28 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="initial-scale=1,maximum-scale=1,user-scalable=no"
+    />
+    <title>Plan Your Active Travel Scheme</title>
```
dabreegster: Should the title of each page differ? It may not matter much in practice -- even if it's "Plan Your Active Travel Scheme -- Browse Schemes", in most browsers with a few tabs open, something that long won't be very visible anyway

src/pages/AccessibilityStatement.svelte
```diff
@@ -0,0 +1,120 @@
...
+    <li>
+      <a href="https://mcmw.abilitynet.org.uk/">AbilityNet</a>
+       has advice on making your device easier to use if you have a disability
+    </li>
+  </ul>
+
+  <h2>How accessible this website is</h2>
+
+  <p>We know some parts of the website are not fully accessible:</p>
+  <p>
```
dabreegster, line 50: This paragraph seems out of place. Did someone write this wording for us and we're just copying it in?

src/pages/AccessibilityStatement.svelte
```diff
@@ -0,0 +1,120 @@
...
+    <li>
+      Colour will be used to convey information such as road width and speed
+      limits, until this is configurable it might present an accessibility
+      issue.
+    </li>
+  </ul>
+
+  <h2>Feedback and contact information</h2>
+  <p>
+    If you need information on this app or to provide further feedback on
```
dabreegster, line 72: Phrasing is a bit odd with the colon. "on issues, please email <a href....>"
Why the "Email - " bit?

src/pages/AccessibilityStatement.svelte
```diff
@@ -0,0 +1,120 @@
...
+    issues: Email - <a
+      href="mailto:plan-your-active-travel-schemes@activetravelengland.gov.uk"
+    >
+      plan-your-active-travel-schemes@activetravelengland.gov.uk
+    </a>
+    . We’ll consider your request and get back to you in 20 days.
+  </p>
+
+  <h2>We’ll consider your request and get back to you in 20 days.</h2>
```
dabreegster, line 81: Why repeat and in h2?

dabreegster (changes requested):
Please run fmt and check

accessibility.html
Pete-Y-CS: I don't mind, happy to make this change but like you say, I imagine it won't be visible a lot of the time

src/pages/AccessibilityStatement.svelte
Pete-Y-CS: I am copying the accessibility statement that was approved by DfT Accessibility guy

Pete-Y-CS:
> Please run fmt and check

check done, fmt not working suddenly for some reason:
```
npm run fmt

> atip@0.0.0 fmt
> npx prettier --write *.html tests/* src/*

sh: 1: prettier: Permission denied
```

dabreegster (approved):
Don't worry about fmt then

PR #552, opened 2024-11-26
Scrub a few remaining uses of 'ATIP' in docs. #536

dabreegster:


README.md
```diff
@@ -1,20 +1,13 @@
-# Active Travel Infrastructure Platform
+# Plan Your Active Travel Schemes
 
-ATIP lets you design, modify and share active travel infrastructure plans.
+This repository contains the Scheme Sketcher and Scheme Browser, two tools for mapping active travel schemes and viewing them alongside contextual layers. The tools are developed by the Alan Turing Institute in collaboration with Active Travel England.
```
Pete-Y-CS, line 3: Maybe we should add a sentence explaining why the repo is called ATIP?

Pete-Y-CS (approved):
Maybe we should add a little explainer of the repo name, but otherwise looks good

README.md
dabreegster: Good point, done

PR #555, opened 2024-11-27
Add a warning if you're in the public beta dev/test env

Pete-Y-CS:
may want some styling changes, and currently testing on public beta env but here's a WIP

### DIFF
--- a/src/lib/common/Header.svelte
+++ b/src/lib/common/Header.svelte
@@ -1,5 +1,7 @@
 <script lang="ts">
   import logoUrl from "../../../assets/ate_logo.png?url";
+  import { appVersion } from "lib/common";
+  import ExternalLink from "./ExternalLink.svelte";
 </script>
 
 <div class="black-bar">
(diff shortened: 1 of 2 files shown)


dabreegster (changes requested):

src/lib/common/Header.svelte
```diff
@@ -10,6 +13,15 @@
 
 <div class="green-bar" />
 
+{#if appVersion() != "Public Beta" && appVersion() != "Private (development)"}
+  <WarningText>
+    NB this is a testing environment. Should you be using
```
dabreegster, line 18: I'd still change the wording, see my other comment. At minimum, let's avoid "NB"

src/lib/common/Header.svelte
```diff
@@ -31,4 +43,9 @@
     margin-top: 15px;
     margin-bottom: 15px;
   }
+
+  .test-env-warning {
```
dabreegster, line 47: Unused, can cleanup

dabreegster (approved):

PR #559, opened 2024-12-04
Add new version of scheme sketcher lib

Pete-Y-CS:


dabreegster:
Actually this doesn't quite work yet. The two cancel buttons behave differently. Try editing something, filling out the form, pressing cancel on the left vs right. In `EditMode`, we first set `unsavedFeature = null` before switching to list mode and triggering `onDestroy`. Quick fix is to plumb through a cancel callback as well. Better fix is to rethink `unsavedFeature`. I can do the latter later today / tomorrow, but maybe the quick fix is worth it in the meantime

dabreegster:
Also as the tests show, now that there's two Finish and two Cancel buttons, we have to be more specific which ones Playwright uses

dabreegster (approved):

PR #560, opened 2024-12-05
Allow some currently unexpected values

Pete-Y-CS:
image
So currently if you make an area using scheme sketcher lib, it ends up with a bunch of disallowed values, as a stop gap, especially for the tutorial, I suggest we just allow them, and figure this out upstream later


dabreegster (commented):
Hmm, we need to update `newPolygonFeature`. Not sure how I missed this previously

PR #569, opened 2024-12-16
Use footer landing page

Pete-Y-CS:
dependent on https://github.com/acteng/govuk-svelte/pull/13


dabreegster:
Demo link https://acteng.github.io/atip/use-footer-landing-page/
Will review more later, but one more request please: let's move the About to a separate page like privacy/accessibility. It doesn't make sense as a modal anymore.

dabreegster (commented):
LGTM -- I left some comments in https://github.com/acteng/govuk-svelte/pull/13, after that's merged and package.json updated here, good to go! Thanks

Issue #480, opened 2024-02-26
Improve processing/display of crossings


Pete-Y-CS:
Improve the processing of crossings so that we can display e.g. zebra crossings. Then consider adding icons

dabreegster:
- Underpass/overpass crossings may not be tagged directly, like https://www.openstreetmap.org/way/50659424
- Could upgrade some unknown cases
  - Node attached to a cycleway, https://www.openstreetmap.org/node/6142822686
  - Node on top of a crossing way that's more detailed, https://www.openstreetmap.org/node/8001342847
- Could consider looking for more attributes, but each of these would require more work besides just directly looking for OSM tags on one object:
  - refuge islands
  - staggered crossings
  - tactile surfaces
  - raised tables
  - lowered kerbs
  - for uncontrolled crossings, the number of lanes to cross and speed of that road

PR #570, opened 2025-01-02
Revamp crossings layer

dabreegster:
#480 

Still iterating on the classification and how much we can extract from OSM; this is just for demoing.
https://acteng.github.io/atip/crossings_v2/browse.html?crossings


src/lib/browse/layers/points/Crossings.svelte
```diff
@@ -71,18 +62,41 @@
 </LayerControl>
 
 <VectorTileSource
-  url={`pmtiles://${publicResourceBaseUrl()}/v1/${name}.pmtiles`}
+  url={`pmtiles://${publicResourceBaseUrl()}/v2/${name}.pmtiles`}
```
dabreegster, line 65: TODO: after the file is updated in the GCS bucket, I'll change this

Pete-Y-CS: Bucket updated

Pete-Y-CS (approved):
pending switch to v1 approved

PR #579, opened 2025-01-13
Adding querystring to local storage on browse page to store "layer session" 

Pete-Y-CS:
…session can be resumed by default

Seems to  be working mostly okay except public transport layers don't work. will investigate but putting it up so you can see my approach and test it out

### DIFF
--- a/src/lib/browse/LayerControls.svelte
+++ b/src/lib/browse/LayerControls.svelte
@@ -74,6 +74,7 @@
     },
   };
 
+  // Before any individual layers check the URL state for their initial setup, restore the query parameters to whatever was saved last time. If there are already query parameters besides the basemap style, then keep them.
   const initialQueryParams = new URLSearchParams(window.location.search);
 
   if (
@@ -85,14 +86,15 @@
     const previousSessionQueryString =
       window.localStorage.getItem(localStorageQuerystringKey) || "";
     const newQueryParams = new URLSearchParams(previousSessionQueryString);
-    // @ts-ignore typescript is convinced foreach doesn't exist but it's working?
-    newQueryParams.entries().forEach(([key, value]) => {
+    for (const entry in newQueryParams.entries) {
+      const key = entry[0];
+      const value = entry[1];
       if (value == null) {
         url.searchParams.delete(key);
       } else {
         url.searchParams.set(key, value);
       }
-    });
+    }
     window.history.replaceState(null, "", url.toString());
   }
 </script>
--- a/src/lib/browse/LayerControls.svelte
+++ b/src/lib/browse/LayerControls.svelte
@@ -86,9 +86,7 @@
     const previousSessionQueryString =
       window.localStorage.getItem(localStorageQuerystringKey) || "";
     const newQueryParams = new URLSearchParams(previousSessionQueryString);
-    for (const entry in newQueryParams.entries) {
-      const key = entry[0];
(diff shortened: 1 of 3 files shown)


Pete-Y-CS:
Okay I was only previously doing the custom state layers but I think now it should work for everything. Will have a little play around

src/lib/browse/LayerControls.svelte
```diff
@@ -72,6 +73,28 @@
       return getRoadLayerNames(map, get(mapStyle));
     },
   };
+
+  const initialQueryParams = new URLSearchParams(window.location.search);
+
```
dabreegster, line 78:
```suggestion
// Before any individual layers check the URL state for their initial setup, restore the query parameters to whatever was saved last time. If there are already query parameters besides the basemap style, then keep them.
```

src/lib/browse/LayerControls.svelte
```diff
@@ -72,6 +73,28 @@
...
+  if (
+    !initialQueryParams ||
+    (initialQueryParams.size === 1 &&
+      initialQueryParams.keys().next().value == "style")
+  ) {
+    let url = new URL(window.location.href);
+    const previousSessionQueryString =
+      window.localStorage.getItem(localStorageQuerystringKey) || "";
+    const newQueryParams = new URLSearchParams(previousSessionQueryString);
+    // @ts-ignore typescript is convinced foreach doesn't exist but it's working?
```
dabreegster, line 88: Always prefer `ts-expect-error`, because if the code does _not_ have any TS error, then the annotation causes an error. `ts-ignore` could wind up being a stray annotation silently not doing anything.

src/lib/browse/LayerControls.svelte
```diff
@@ -72,6 +73,28 @@
...
+    newQueryParams.entries().forEach(([key, value]) => {
```
dabreegster, line 89: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/forEach works fine, no `entries`. Or do regular `for` loops to remain consistent with the rest of the codebase

dabreegster (changes requested):
Demo link https://acteng.github.io/atip/store-querystring-in-localstorage/browse.html

Seems to work great, nice!

src/lib/browse/LayerControls.svelte
```diff
@@ -85,14 +86,15 @@
     const previousSessionQueryString =
       window.localStorage.getItem(localStorageQuerystringKey) || "";
     const newQueryParams = new URLSearchParams(previousSessionQueryString);
-    // @ts-ignore typescript is convinced foreach doesn't exist but it's working?
-    newQueryParams.entries().forEach(([key, value]) => {
+    for (const entry in newQueryParams.entries) {
```
dabreegster, line 89: Destructuring works too, `for let ([key, value] of newQueryParams.entries())`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

And I don't think you want [for in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)

And [entries](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/entries) is a method, does `in x.entries` work? (and TS not complain?!)

dabreegster (approved):
LGTM

PR #582, opened 2025-01-23
When the user initially uses ?disable_remote_schemes and then later loads the schemes, fix up the URL

dabreegster:
More noticeable now that we save the query string!

To verify this works, go to `browse.html?disable_remote_schemes` initially, then "load latest scheme data", verify the URL changes. Go to `browse.html` with no params and verify schemes do load. (If you're testing locally, for the remote schemes to show up at all, follow the dev guide instructions for VITE_MIMIC_GCP_LOCALLY)


Pete-Y-CS (approved):
Looks good, better to centralise that storage functionality, too!

PR #583, opened 2025-01-30
Make prod deployment script

dabreegster:
After merging this in main, we can copy it to the `prod` branch.

Let's please test against dev env first


backend/deploy_prod.sh
```diff
@@ -0,0 +1,25 @@
...
+  exit
+fi
+
+read -p "Do you want to deploy to prod? (y/n)?" CONFIRM
+if [ "$CONFIRM" != "y" ]; then
+  exit
+fi
+
+# Build the app
+npm ci
```
dabreegster, line 17: ci, not i, to make sure we use package-lock. Following exactly what's in the working GH workflows

backend/deploy_prod.sh
```diff
@@ -0,0 +1,25 @@
...
+npm run setup-govuk
+VITE_RESOURCE_BASE="https://plan.activetravelengland.gov.uk/data" npm run build
+
+# Deploy to app engine
+cd backend
+rm -rf dist
+cp -R ../dist .
+gcloud app --project=dft-rlg-schemes-gis-prod deploy app_public_prod.yaml --quiet
```
dabreegster, line 25: I don't remember why `--quiet` is useful, I'm testing against my dev project...

deploy_prod.sh
dabreegster, line 1: Very old script, remove to not cause confusion

dabreegster:
OK, I have confidence this script works. I suggest we merge and then manually push a commit to `prod` with this script too.

Pete-Y-CS (approved):

PR #587, opened 2025-02-18
Pull in fix from https://github.com/acteng/scheme-sketcher-lib/pull/1…

dabreegster:
…6 for decompressing route snapper files when headers get intercepted

Letting tests run, but I don't expect anything to change


Pete-Y-CS (approved):

Pete-Y-CS:
approved pending tests

PR #588, opened 2025-02-19
Add a browse layer for existing OSM cycle routes

dabreegster:
Demo: https://acteng.github.io/atip/osm_routes/browse.html?cycle_routes=1#16.16/51.494749/-0.100203

I'm not sure why I never thought to include this. OSM has named cycle routes, modeled as relations. I think it's quite illustrative to look at these alongside the existing "cycle paths" layer, which should maybe be rephrased as "Cycle infrastructure". Compare the more semantic routes:
image
with the actual infrastructure making them:
image

Not sure it's visible easily in the screenshot, but Elliot's Row is an example where it's part of one logical route, but it's just a quiet residential street; there's no physical infrastructure there.

Note I'd love to let people click the routes and open in OSM, like https://www.openstreetmap.org/relation/11739129. But that'd require doing our own work to stitch together the linestring geometry from the relation, and I'm just getting a quick thing started over in data-prep.

### DIFF
--- a/src/lib/browse/LayerControls.svelte
+++ b/src/lib/browse/LayerControls.svelte
@@ -18,6 +18,7 @@
   import WardsLayerControl from "./layers/areas/Wards.svelte";
   import BusRoutesLayerControl from "./layers/lines/BusRoutes.svelte";
   import CyclePathsLayerControl from "./layers/lines/CyclePaths.svelte";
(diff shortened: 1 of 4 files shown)


Pete-Y-CS (approved):
It's all in the buckets and ready to go!

PR #590, opened 2025-02-26
Add a browse layer with OSMM road and sidewalk polygons. #532

### LINKED ISSUE #532: Enhanced road / pavement width browse layers
Can we show a line on each side of the pavement? Or better yet, can we just show the pavement polygons from OSMM and still have colouring/a popup show average/min width?

dabreegster:
Requires GCS updates for the private layer. Closes #532.


Pete-Y-CS (approved):
Files uploaded, have reviewed, haven't tested locally, can do if you like but happy to merge since it will only affect the internal version if it doesn't work.

PR #595, opened 2025-04-01
Add final inspections sketches as a separate category

Pete-Y-CS:


src/lib/browse/schemes/data.ts
```diff
@@ -23,11 +24,20 @@ export function setupSchemes(
...
+    schemes: {},
+    notes: gj.notes,
+  };
+
   for (let [scheme_reference, scheme] of Object.entries(gj.schemes)) {
-    if (scheme.pipeline) {
+    if (scheme.browse?.sketch_source === "LCWIP mapping") {
       lcwipGj.schemes[scheme_reference] = scheme;
-    } else {
+    } else if (scheme.browse?.sketch_source == "ATF assessment") {
```
Sparrow0hawk, line 37: Do we want this to be strict equality as with the first if branch?

src/lib/browse/schemes/data.ts
```diff
@@ -23,11 +24,20 @@ export function setupSchemes(
...
       atfGj.schemes[scheme_reference] = scheme;
+    } else if (scheme.browse?.sketch_source == "Final inspection sketches") {
```
Sparrow0hawk, line 39: Same question as [above](https://github.com/acteng/atip/pull/595/files#r2041781023)

Pete-Y-CS:
I tried to  do the refactor suggested in conversation - not working at first blush so gonna merge this and see if we can't resolve that down the line

PR #600, opened 2025-04-16
Make the internal deployments manually dispatchable

Pete-Y-CS:


Sparrow0hawk (approved):
Happy to merge this. 

Could we raise an issue just flagging the flaky test? As that is something it would be nice to try and address.

PR #604, opened 2025-05-28
A date_exported to sketch files - resolves #593

### LINKED ISSUE #593: When exporting a sketch, include export date
basically ensure that we have some measure of when a sketch was drawn and therefore how up to date it might be

Pete-Y-CS:


src/lib/common/files.ts
```diff
@@ -117,6 +117,7 @@ export function serializeSchemes(authority: string, gj: Schemes): Schemes {
   }
   copy.authority = authority;
   copy.origin = "atip-v2";
+  copy.sketch_exported = new Date().toDateString();
```
Sparrow0hawk, line 120: I think for dates/datetimes probably best to try and go with ISO 8601 format to help with machine readability as well as human. Which I think you can achieve with [`Date.prototype.toISOString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString). 

```suggestion
  copy.sketch_exported = new Date().toISOString();
```

Sparrow0hawk (changes requested):
Want your brief thoughts on tweak to proposed change but once covered off this will be good to merge.

src/lib/common/files.ts
Pete-Y-CS: makes complete sense - I tried a couple and went for increased human readability, but given it's going into the data platform better to go for the more easily processed format

PR #605, opened 2025-06-04
create public prod workflow

### LINKED ISSUE #606: Create a GitHub action for deploying to Plan your active travel schemes prod environment
To simplify deployments to production for Plan your active travel schemes we need a GitHub action.

Pete-Y-CS:
Resolves #606

### DIFF
--- a/.github/workflows/deploy_public_prod.yml
+++ b/.github/workflows/deploy_public_prod.yml
@@ -1,9 +1,6 @@
-name: Deploy to app engine (public test env)
+name: Deploy to app engine (public prod env)
 
 on:
-  push:
-    branches:
-      - main
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_public_test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - name: Authenticate to GCP
         uses: google-github-actions/auth@v1
         with:
-          credentials_json: '${{ secrets.GCP_PUBLIC_test_CREDENTIALS }}'
+          credentials_json: '${{ secrets.GCP_PUBLIC_TEST_CREDENTIALS }}'
```
Pete-Y-CS, line 39: It seems these are case insensitive because this was working before, but I figure better to have it be consistent

Sparrow0hawk (approved):

PR #623, opened 2025-08-21
Do not automatically show sketches

Pete-Y-CS:


Sparrow0hawk:
This is feels like a tricky one re the behaviour the end user wants/is also quite a big change away from the current default. Couple of notes:
- This change means no scheme sketches will immediately be shown but the other previously opened layers will still be shown on the map
- Considering the previous point, is the desireable starting point for a user on the browse page just an empty map and they build layers up? Or do we want to remember some layers from the previous session?

I think we need to know what the desired starting point would be from users before fully committing to a change. Could we discuss with Nicos and any other inspectors using the Browse page about this?

PR #636, opened 2026-02-19
Add support for CATF in browse apge

Pete-Y-CS:


Sparrow0hawk:
One thing we'll want to consider after looking at this is how best to visualise sketches that are immediately on top of each other.

Sparrow0hawk (approved):

PR #653, opened 2026-06-11
Serve route-snapper file locally for tests

Sparrow0hawk:
This PR attempts to allow the test suite to run without fetching route-snapper files from a remote host.

It includes:
- Clarifying our playwright dev dependency
- Upgrading playwright to 1.60.x
- Adding express as a dev dependency and using express to serve the single route-snapper file used for tests
- Configure playwright to use this test server during tests
- Explicitly disable playwright parallelism to reduce test flakiness


Pete-Y-CS:
Without the parallelism - these tests seem to be fine for me, but that said it might still be worth pulling [this](https://github.com/acteng/atip/pull/656) just to be sure in case something slows down - this is a fix that fixed flakiness on main for me

Sparrow0hawk:
I'd prefer to avoid applying specific timeouts to individual tests, that for me feels like a smell that something isn't working as we'd like if particular parts of the page are taking too long to load.

Did you find this branch was flaky? If so let me know so I can test it some more.

Pete-Y-CS:
Well, they fail with timeouts for me if I reintroduce parallelisation, so it may be 6 of one, half a dozen of the other for us to switch off paralellisation to make the timeouts not happen, vs having paralellisation and a longer timeout

Pete-Y-CS:
I guess the answer may be - do some meta-tests of what takes longer?

The route snapper loads plenty fast on PYATS itself, so I think these tests aren't testing how long the page takes to load, they're testing other features. With that in mind whatever gives us fastest running test suite is the best option, no?

Sparrow0hawk:
I am keen to avoid parallelism as it potentially hides slower running tests which is something we want to identify and address. 

In CI tests runs as quickly with this branch as they do on previous PRs on main (around 2mins) so I don't think we actually gain much from parallelism. If we think the test suite is too slow we probably need to get under the hood of which tests are really slowing us down and why.

tests/test-server/server.js
```diff
@@ -0,0 +1,24 @@
...
+  resp.setHeader("Access-Control-Allow-Origin", "*");
+  next();
+});
+
+app.use(
+  "/route-snappers",
+  express.static(path.join(import.meta.dirname, "route-snappers")),
+);
+
+app.get("/", async (req, resp, next) => {
```
Pete-Y-CS, line 16: This looks like it's just here for testing purposes - do we need/want it?

Pete-Y-CS (approved):
Added a comment - looks like there's a test hello world we might wanna remove but otherwise looks good!

tests/test-server/server.js
Sparrow0hawk: We need this so that playwright knows the server is up during tests.

Pete-Y-CS: Alright, then ship it!

PR #661, opened 2026-06-24
Begin the work of adding a new type, with more modular form

Pete-Y-CS:


src/lib/sketch/v3/FeatureSubform.svelte
Pete-Y-CS, line 1: Perhaps this wants a different name, it is awkward that we have 'features' of a geojson file and 'features' of a piece of infrastructure. This means our features will have features.....

src/lib/sketch/v3/FormV3.svelte
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      props.name = $routeTool!.inner.routeNameForWaypoints(props.waypoints);
+    } catch (e) {
+      window.alert(`Couldn't auto-name route: ${e}`);
+    }
+  }
+
+  let nameId = uuidv4();
+
+  function formatForDisplay(original: string): string {
```
Pete-Y-CS, line 44: This might want to be brought out to be shared with other bits of codebase, but I'm not sure

src/lib/sketch/v3/InfrastructureTypeSubform.svelte
```diff
@@ -0,0 +1,179 @@
...
+    );
+
+    if (filteredTypeOptions.length < 1) {
+      return undefined;
+    }
+
+    return filteredTypeOptions[0];
+  }
+
+  function formatForRadio(original: string): [string, string] {
```
Pete-Y-CS, line 36: Like here's a very similar but slightly different version of the same function, probably best to split it out elsewhere and settle on which way of it working is nicer

src/lib/sketch/v3/InfrastructureTypeSubform.svelte
```diff
@@ -0,0 +1,179 @@
...
+
+  $: contextObject = getContextObject(intervention.properties.context);
+
+  $: contexts = subtypeObject
+    ? subtypeObject.possible_contexts?.map((contextObject) => {
+        return formatForRadio(contextObject.context);
+      })
+    : [];
+
+  $: {
```
Pete-Y-CS, line 82: This is Svelte syntax for 'run this code if the inputs to it change' so it means that when you change the subtype or context change prune any unneeded features and measures from the properties, and add any newly needed features and measures

src/lib/sketch/config.ts
```diff
@@ -327,5 +329,14 @@ export function getUnexpectedProperties(props: { [name: string]: any }): {
     }
   }
 
+  if (schema == "v3" && copy.v3) {
```
Pete-Y-CS, line 332: This where we are checking if there are any properties on a given feature that aren't expected for the schema we are currently using

Pete-Y-CS:
@Sparrow0hawk  I've added some explanatory comments throughout the code, hopefully a decent amount of it is self explanatory but if you have any questions we can definitely go through em

PR #667, opened 2026-07-22
make the form work with other infra types

Pete-Y-CS:


src/lib/sketch/v3/InfrastructureTypeSubform.svelte
```diff
@@ -12,10 +12,24 @@
   import MeasureSubform from "./MeasureSubform.svelte";
 
   export let intervention: V3Intervention;
-  const intervention_type_options: InfrastructureSubtype[] =
-    intervention.intervention_type !== ""
-      ? infrastructureFormInformation[intervention.intervention_type].types
+  $: intervention_type_options = getInterventionTypeOptions(
```
Pete-Y-CS, line 15: This is important so that the options in the component change whenever a new intervention type is selected

PR #678, opened 2026-08-04
Update tooltips to match latest layer generation time

Pete-Y-CS:


Pete-Y-CS:
In order to merge and deploy to PYATS public version we'll have to add the new files to the bucket. I have updated the internal version

